PR #125, opened 2023-10-18
Make Default Message in English Always

CalebPena:
- Explicitly set default message to English


screener/views.py
```diff
@@ -269,6 +270,7 @@ def sort_first(program):
 
 
 def default_message(translation):
+    translation.set_current_language(settings.LANGUAGE_CODE)
```
ivonne-hernandez, line 273: This sets the current language to `en-us`.

ivonne-hernandez (approved):

PR #163, opened 2023-11-08
Define Marital Units for Policy Engine

CalebPena:
- Use our marital units in Policy Engine.

### DIFF
--- a/programs/programs/policyengine/policyengine.py
+++ b/programs/programs/policyengine/policyengine.py
@@ -341,7 +341,8 @@ def policy_engine_prepare_params(screen):
                     "co_tanf_countable_earned_income_grant_standard": {"2023": None},
                     "broadband_cost": {"2023": 500},
                 }
-            }
+            },
+            "marital_units": {}
         }
     }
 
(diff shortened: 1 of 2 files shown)


programs/programs/policyengine/policyengine.py
```diff
@@ -397,4 +398,8 @@ def policy_engine_prepare_params(screen):
         policy_engine_params['household']['households']['household']['members'].append(member_id)
         policy_engine_params['household']['spm_units']['spm_unit']['members'].append(member_id)
 
+    for member_1, member_2 in screen.relationship_map().items():
```
ivonne-hernandez, line 401: Does this assume that the second household member that was entered is the first household member's spouse?

screener/models.py
```diff
@@ -214,7 +213,7 @@ def relationship_map(self):
                         probabable_spouse = other_member['id']
                         break
             relationship_map[member['id']] = probabable_spouse
-            if probabable_spouse is None:
+            if probabable_spouse is not None:
```
ivonne-hernandez, line 216: Did you mean `probable`?

programs/programs/policyengine/policyengine.py
CalebPena: No, it assumes the first person with the relationship status of spouse is the spouse, and it also assumes that if there are 2 of `parent`, `fosterParent`, `stepParent`, or `grandParent` that those are spouses as well. You can see how the relationship map works [here](https://github.com/Gary-Community-Ventures/benefits-api/blob/main/screener/models.py#L185).

civicpixel (approved):

PR #214, opened 2024-01-09
Program Dependencies

CalebPena:
- Add ability to not calculate dependency if the screen is missing some fields.

- Add a base class for all programs.
- Move Policy Engine inputs onto a class for each program.
- Make urgent needs be able to have dependencies.


civicpixel (approved):
The new program architecture looks great. The size of this PR is intense, I'm going to recommend to QA team that we slow roll and do a lot of value / eligibility testing.

PR #233, opened 2024-01-22
Add Emergency Mortgage Assistance Urgent Need

CalebPena:
- Add Emergency Mortgage Assistance urgent need function.


programs/programs/urgent_needs/urgent_need_functions.py
```diff
@@ -158,6 +161,45 @@ def eligible(cls, screen: Screen):
...
+        limits = CoEmergencyMortgageAssistance.limits_cache.fetch()
+
+        if screen.county not in limits:
+            return False
+
+        income_limit = limits[screen.county][
+            screen.household_size - 1
+        ]
+
+        return income < income_limit
```
ivonne-hernandez, line 205: Do we want `<=`?

ivonne-hernandez (approved):
Do we want `<=` on line 205 of `programs/programs/urgent_needs/urgent_need_functions.py`?

programs/programs/urgent_needs/urgent_need_functions.py
CalebPena: @kbottagaro is going to go through the programs and check for this at some point. This is a big edge case, so I am going to merge as is.

PR #247, opened 2024-01-26
SSI SGA

CalebPena:
- Refactor the SSI disabled dependency so that it can use the SGA in it's calculation.

### DIFF
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -109,11 +109,25 @@ class SsiUnearnedIncomeDependency(Member):
         return int(self.member.calc_gross_income('yearly', ['unearned']))
 
 
-class SsiDisabledDependency(Member):
-    field = "is_ssi_disabled"
+class IsDisabledDependency(Member):
+    field = "is_disabled"
 
     def value(self):
-        return self.member.has_disability()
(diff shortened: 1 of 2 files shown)


programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -109,11 +109,25 @@ def value(self):
...
+    field = 'is_blind'
+
+    def value(self):
+        return self.member.visually_impaired
+
+
+class SsiReportedDependency(Member):
+    field = 'ssi_reported'
+
+    def value(self):
```
mexi-cano, line 129: This seems like a placeholder or stub implementation? It might be helpful to include a comment explaining the reason for always returning 0.

mexi-cano (commented):
Overall looks good to me aside from always returning 0 in SsiReportedDependency

programs/programs/policyengine/calculators/dependencies/member.py
CalebPena: Updated

CalebPena:
@mexi-cano FYI here is the Policy Engine [is_ssi_disabled](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/ssa/ssi/eligibility/status/is_ssi_disabled.py)

mexi-cano (approved):

PR #290, opened 2024-02-27
Add va to insurance types

myang1010:
- Updated cfhc value, if VA selected, return 0
- Removed "dont_know" from insurance_type as we will not use "dont_know" anymore

- "dont_know" is left in `Insurance` model. We may need to remove them in the future?

Please let me know if anything need to be changed. Thanks!


programs/programs/connect_for_health/calculator.py
```diff
@@ -35,5 +35,10 @@ def eligible(self) -> Eligibility:
         return e
 
     def value(self, eligible_members: int):
-        # https://stackoverflow.com/questions/6266727/python-cut-off-the-last-word-of-a-sentence
-        return tax_credit_by_county[self.screen.county.rsplit(' ', 1)[0]] * 12
+        value = 0
+        if not self.screen.has_insurance_types(('va', 'private')):
```
CalebPena, line 39: I think that this would be better to add as a condition on the eligibility method.

screener/models.py
```diff
@@ -526,6 +526,7 @@ class Insurance(models.Model):
     medicare = models.BooleanField(default=False)  # elderly health insurance
     emergency_medicaid = models.BooleanField(default=False)
     family_planning = models.BooleanField(default=False)
+    va = models.BooleanField(default=False)
```
CalebPena, line 529: Could you add this to the serializer as well.

CalebPena (changes requested):
Change the default of the `dont_know` field to `null` or `False` and change the default of `none` to `True`.

Could you also test that VA is saving. The easiest way to do this is to select VA for one of the household members and then go to the next member (saving the screen) and then go back, refresh the page, and make sure that it is still selected.

myang1010:
I think it makes sense to check "va" insurance status on household members. This "chfc" benefit should still be available for someone in the household doesn't have "va" insurance.

CalebPena (approved):
Looks good.

PR #299, opened 2024-03-07
Use Spreadsheet for Connect For Health value calculation

myang1010:
- Uses spreadsheet for value calculation and removes `tax_credit_value.py` file


CalebPena (approved):
LGTM

PR #312, opened 2024-04-02
Update RTD Live

CalebPena:
- Update income limit.
- Only include members of the household.
- Move `is_head`, `is_spouse`, and `is_dependent` to methods on members

### DIFF
--- a/screener/models.py
+++ b/screener/models.py
@@ -388,6 +388,33 @@ class HouseholdMember(models.Model):
     def has_disability(self):
         return self.disabled or self.visually_impaired or self.long_term_disability
 
+    def is_head(self):
+        return self.relationship == 'headOfHousehold'
+
+    def is_spouse(self):
+        return self.screen.relationship_map()[self.screen.get_head().id] == self.id
+
+    def is_dependent(self):
+        is_tax_unit_spouse = self.is_spouse()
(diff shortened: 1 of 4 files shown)


myang1010 (approved):
Looks good!

screener/models.py
```diff
@@ -388,6 +388,33 @@ def is_married(self):
...
+    def is_head(self):
+        return self.relationship == 'headOfHousehold'
+
+    def is_spouse(self):
+        return self.screen.relationship_map()[self.screen.get_head().id] == self.id
+
+    def is_dependent(self):
+        is_tax_unit_spouse = self.is_spouse()
+        is_tax_unit_head = self.is_head()
+        is_tax_unit_dependent = (
```
myang1010, line 400: More of a business logic question here. I saw on the [IRS website](https://apps.irs.gov/app/vita/content/globalmedia/4491_dependency_exemptions.pdf) that dependents include "a child, stepchild, brother, sister, or parent." Our definition of `is_depent` doesn't seem to include parents, siblings, step child, .etc, are we purposely leave them out here? Thanks.

CalebPena: People over 18 require special cases to be qualified as a dependent, such as being a student, or being disabled. We do account for these edge cases on lines 403 and 404.

PR #318, opened 2024-04-10
Move Calculators to State Folders

CalebPena:
- Create `nc` folder.

- Move state specific programs to the `co` folder.
- Move federal programs to the `federal` folder.
- Move Policy Engine calculators to the respective state or federal folder.
- Make Medicaid a base class where each of the states can inherit from it.


myang1010 (approved):
Thanks, Caleb, for initiating the change. It looks good and is quite self-explanatory. The examples provided are also helpful. Much appreciated!

msrezaie (approved):
Looks good. Also thank you for taking the time and adding some of the NC specific files/classes as well. Great work!

PR #329, opened 2024-04-18
Bug/no results

msrezaie:
- Added three new command scripts intended to be used for the NC instance of the screener app for resolving a not showing results bug caused due to program values mismatch.

- The programs list in the 'quick_start.py' has a repeated dictionary for 'emergency_medicaid', removed it since it was causing errors while running the command `quick_start`.

Were there any issues that arose?/Any other comments, questions, or concerns?
- Within the 'quick_start.py' there are two dictionaries with the same 'abbr' of 'coeitc' but one has a different 'external' value which is 'coexeitc'. The 'coeitc' refers to the Colorado EITC program which has calculation logic in the API. Though the 'coexeitc' which refers to the Colorado Expanded EITC program, does not have any calculation logic in the API. This causes the screener to throw the following error in the last step:
```
File "...\benefits-api\screener\views.py", line 232, in eligibility_results
    eligibility = program.eligibility(screen, data, missing_dependencies)
 eligibility
    Calculator = calculators[self.name_abbreviated.lower()]
KeyError: 'coexeitc'
```

This can be resolved by deactivating the 'coexeitc' through admin. Though is this:
a. 'abbr' naming typo in the programs list in 'quick_start.py'? or
b. the 'coexeitc' program is using the same calculation logic in the API as the 'coeitc' program?


programs/management/commands/quick_start.py
```diff
@@ -85,7 +84,7 @@ class Command(BaseCommand):
         {'abbr': 'chs', 'external': 'chs'},
         {'abbr': 'cccap', 'external': 'cccap'},
         {'abbr': 'chp', 'external': 'chp'},
-        {'abbr': 'coeitc', 'external': 'coexeitc'},
+        {'abbr': 'coexeitc', 'external': 'coexeitc'},
```
CalebPena, line 87: The reason that we have a `name_abbreviated` and an `external_name` is because of this program. The `name_abbreviated` refers to which calculator to use. The `external_name` helps to keep translations in sync when you import translations. COEITC or the Colorado Earned Income Tax Credit has the same calculator as COEXEITC or the Colorado Expanded Earned Income Tax Credit. They have different descriptions because one is for citizens, and the other is for non-citizens. Therefore, `coexeitc` does not have a calculator, and will throw an error. I think that it is better for each program to have its own calculator, so could you add `'coexeitc': tax.Coeitc` to `co_tax_unit_calculators` in `programs/programs/co/pe/__init__.py`? FYI I merged my state folder pr into dev, so you will need to pull it to get that file.

programs/management/commands/remove_programs.py
```diff
@@ -0,0 +1,61 @@
...
+
+        navigators = Navigator.objects.all()
+        programs = Program.objects.all()
+
+        for navigator in navigators:
+            translations_to_delete = []
+            for field in navigator_translated_fields:
+                translations_to_delete.extend(
+                    Translation.objects.filter(
+                        label__startswith=f'navigator.{navigator.name}_{navigator.id}-{field}')
```
CalebPena, line 43: Does this work? That way, if the label changes, we don't have to worry about changing this script.
```suggestion
                    getattr(navigator, field)
```

CalebPena (commented):
@bespinosa01 Mentioned that you are trying to come up with a solution. I would recommend taking a look at [bulk_import_translations.py](https://github.com/Gary-Community-Ventures/benefits-api/blob/main/translations/bulk_import_translations.py). This function is used to import translations. Lines 48 and 49 deal with adding a program that has translations for it, but does not exist in the database. Perhaps it can be updated so that it can better create new programs.

programs/management/commands/quick_start.py
msrezaie: Oh thanks for the clarification. After adding coexeitc, do not see any errors so thanks for the recommended fix.

msrezaie:
> The reason that we have a name_abbreviated and an external_name is because of this program. The name_abbreviated refers to which calculator to use. The external_name helps to keep translations in sync when you import translations. COEITC or the Colorado Earned Income Tax Credit has the same calculator as COEXEITC or the Colorado Expanded Earned Income Tax Credit. They have different descriptions because one is for citizens, and the other is for non-citizens. Therefore, coexeitc does not have a calculator, and will throw an error. I think that it is better for each program to have its own calculator, so could you add 'coexeitc': tax.Coeitc to co_tax_unit_calculators in programs/programs/co/pe/__init__.py? FYI I merged my state folder pr into dev, so you will need to pull it to get that file.

So after some testing, there doesn't seem to be any difference between the two solutions to this issue. These solutions are 1st: have the `abbr` name for the Colorado Expanded Earned Income Tax Credit program be the same as Colorado Earned Income Tax Credit program which is `coeitc`, or 2nd: have there be another key in `co_tax_unit_calculators` like so `'coexeitc': tax.Coeitc`.
I believe it would be better to leave this as is with the 1st (already implemented) solution, otherwise anyone currently working on this would have to change the abbreviated name for the Colorado Expanded Earned Income Tax Credit program to `coexeitc` either through the admin or through another script.

msrezaie:
@CalebPena it is.

CalebPena (approved):

PR #344, opened 2024-04-30
Use Column Names To Determine LEAP Values

CalebPena:
- Add method to get a list of values by column name.
- Add `GoogleSheetsCache` base class.

- Move `sheets_get_data` to a class.
- Move Google Sheets logic to integrations.
- Don't fetch from sheets on reload.


integrations/services/sheets/sheets.py
```diff
@@ -0,0 +1,98 @@
...
+        result = self.sheet.values().get(
+                spreadsheetId=self.spreadsheet_id, range=self.cell_range
+        ).execute()
+        values = result.get('values', [])
+
+        return values
+
+    def data_by_column(self, *column_names: str) -> list[dict[str, any]]:
+        '''
+        return an array of dictiontionaries containing the column names and their values
```
ivonne-hernandez, line 34:
```suggestion
        return an array of dictionaries containing the column names and their values
```

integrations/services/sheets/sheets.py
```diff
@@ -0,0 +1,98 @@
...
+        WARN: this should only be used during development
+        '''
+        raw_column_names = self.data()[0]
+
+        for name in raw_column_names:
+            print(repr(name))
+
+    def _raise_missing_columns(self, needed_columns: list[str], existing_columns: list[str]):
+        '''
+        raise an exception with the column names from needed_colomns that are not in existing_columns
```
ivonne-hernandez, line 76:
```suggestion
        raise an exception with the column names from needed_columns that are not in existing_columns
```

ivonne-hernandez (approved):

PR #351, opened 2024-05-02
1036: wip

Mdemenko1:

Implementing dark/light mode

image


translations/static/js/theme.js
```diff
@@ -0,0 +1,27 @@
+(() => {
+  "use strict";
+  const prefersDarkMode = window.matchMedia(
+    "(prefers-color-scheme: dark)"
+  ).matches;
+  const defaultTheme = prefersDarkMode ? "dark" : "light";
+  const preferredTheme = localStorage.getItem("theme");
+  const toggleDarkMode = document.querySelector("#bd-theme");
+  const isDarkTheme = localStorage.getItem("theme") === "dark";
```
msrezaie, line 9: Move this inside of the toggleDarkMode event listener, because in this implementation the toggle switch only works once.

translations/templates/base.html
```diff
@@ -9,13 +9,70 @@
...
+          <svg
+            xmlns="http://www.w3.org/2000/svg"
+            fill="currentColor"
+            class="bi bi-sun"
+            viewBox="0 0 16 16"
+          >
+            <path
+              d="M8 11a3 3 0 1 1 0-6 3 3 0 0 1 0 6m0 1a4 4 0 1 0 0-8 4 4 0 0 0 0 8M8 0a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 0m0 13a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 13m8-5a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2a.5.5 0 0 1 .5.5M3 8a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2A.5.5 0 0 1 3 8m10.657-5.657a.5.5 0 0 1 0 .707l-1.414 1.415a.5.5 0 1 1-.707-.708l1.414-1.414a.5.5 0 0 1 .707 0m-9.193 9.193a.5.5 0 0 1 0 .707L3.05 13.657a.5.5 0 0 1-.707-.707l1.414-1.414a.5.5 0 0 1 .707 0m9.193 2.121a.5.5 0 0 1-.707 0l-1.414-1.414a.5.5 0 0 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .707M4.464 4.465a.5.5 0 0 1-.707 0L2.343 3.05a.5.5 0 1 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .708"
+            />
+          </svg>
```
msrezaie, line 47: We can use something like this instead:
```
<span class="material-symbols-outlined"> light_mode </span>
```

and the same for the other `svg` tag. These icons within the `span` tag is coming from the google fonts import. You can find all of them from here:
https://fonts.google.com/icons

msrezaie (changes requested):
Not too sure on the positioning of the theme switch. One idea I was thinking of is to have the button be placed at the bottom of the sidebar and make the entire sidebar have the sticky property.
image

translations/static/js/theme.js
Mdemenko1: Then I will not be able to access those variables outside of the function

msrezaie: Use the `preferredTheme` for `toggleDarkMode.checked` instead of `isDarkTheme` which seems to be the only instance of it used outside of the event listener.
```
toggleDarkMode.checked = preferredTheme === "dark" ? true : false;
```

msrezaie (approved):
Looks great, awesome work!

PR #377, opened 2024-05-22
Small Admin Improvements

CalebPena:
- Import all programs with translations.
- Auto translate even if it has been manually edited.
- Try to add `external_name` automatically.
- Add a default translation.
- Add name back to translation admin.
- Fix urgent need warning link.

- The multi-selects are not scrollable.


translations/templates/urgent_needs/urgent_need.html
msrezaie, line 1: Good catch.

PR #456, opened 2024-07-19
Updated Configuration Seed File with More Help Values

msrezaie:
- Added a new config object which includes values for the More Help page.


CalebPena (approved):
Thank you

PR #482, opened 2024-08-02
Fix/nurturing futures inc limit

ivonne-hernandez:
- Fix nurturing futures inc limit


ivonne-hernandez:
Caleb has the fix in the release PR :).

PR #488, opened 2024-08-09
Feat/Added new configuration data field UI

msrezaie:
- With the use of [django-json-widget](https://pypi.org/project/django-json-widget/) package, the `data` field of the Configuration model now utilizes a JSON editor window in the main Django admin dashboard. This change helps in providing a more user friendly interface for admins to manage the configuration instances of the screener.
- The JSON editor for the `data` field provides two different views, namely Tree and Code. The JSON editor also provides search, sort, order, and other features.
image

- The JSON data, when saved in the data field of the Configuration model, previously included escape characters. However, after passing the data through `json.loads` in the `admin.py` of the Configuration model, the escape characters `\` are now excluded only while the `data` is being viewed in the JSON editor.

- The django-json-widget package is added in the requirements.txt, the package needs to be installed for the new `data` UI to work.

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -82,6 +82,7 @@ INSTALLED_APPS = [
     "rest_framework.authtoken",
     "phonenumber_field",
     "parler",
+    'django_json_widget',
     "django_filters",
     "django.contrib.admin",
     "django.contrib.auth",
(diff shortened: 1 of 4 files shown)


CalebPena (approved):
Looks good TY.

PR #503, opened 2024-08-20
Warning message admin

CalebPena:
- Add `WarningMessage` model.
- Add warning message` main admin.
- Add warning message translation admin.
- Translation import now imports warning messages.
- Add initial warning message calculators (_show, _dont_show, _tax_unit).

- Rename `NavigatorCounty` to `County`


programs/models.py
```diff
@@ -367,6 +368,54 @@ def __str__(self):
...
+            translation.label = f"navigator.{calculator}_{warning.id}-{field}"
+            translation.save()
+
+        return warning
+
+
+class WarningMessage(models.Model):
+    programs = models.ManyToManyField(Program, related_name="warning_messages", blank=True)
+    external_name = models.CharField(max_length=120, blank=True, null=True, unique=True)
+    calculator = models.CharField(max_length=120, blank=False, null=False)
```
msrezaie, line 401: while adding a new warning message, what can be the value for the `calculator` field?

translations/templates/warning_messages/filter.html
```diff
@@ -0,0 +1,20 @@
+{% load static %}
+<form
+  hx-get="warning_messages/filter"
+  hx-target="#document-container"
```
msrezaie, line 4:
```suggestion
  hx-target="#warning-message-container"
```

msrezaie (changes requested):

programs/models.py
CalebPena: Currently is can be: "_show", "_dont_show", or "_tax_unit".

This is defined in: programs/programs/warnings/__init__.py

This also lets us create custom calculators. For example, if we wanted to display a message on a program if they have children that are younger than a certain age, we could make a warning calculator for that.

msrezaie (approved):
LGTM

PR #511, opened 2024-08-21
Fix bug where Medicaid was not being calculated first

CalebPena:
- Change `medicaid` to the state versions of Medicaid when determining whether to calculate a program first.

CFH will still show if the household has CHP+. Will be fixed in a future issue.


bespinosa01 (approved):
I know I wasn't the requested reviewer. But was reported internally and came to open an issue/PR, but pleasantly surprised to find yours!

PR #514, opened 2024-08-27
Adds NC WIC

bespinosa01:
-Added `nc_wic` program, was necessary to update `wic_category` values to be more reflective to NC

- Added `NcWic` Class 
- Updated `nc_member_calculators` to include said class

N/A

- Please Review and Test :) 


programs/programs/nc/pe/member.py
CalebPena, line 1: Can you make a `CoWic`. I think that it is a little confusing to have the `Wic` be the Colorado WIC.

CalebPena (approved):

programs/programs/nc/pe/member.py
bespinosa01: Yeah I didn't want to force y'alls hands into any changes so left CO values as defaults. But makes sense to me, changes made.

CalebPena (approved):

PR #516, opened 2024-08-28
Feat/second tax unit

CalebPena:
- Add `other_tax_unit_structure` method to get the other tax unit.

- If household members are not in the main tax unit than put them in a second tax unit.
- Update PE dependencies to account for multiple tax units.
- Update PE calculators to account for multiple tax units.


programs/programs/policyengine/policy_engine.py
```diff
@@ -105,35 +115,44 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator]):
             if hasattr(program, "pe_output_period") and Data in program.pe_outputs:
                 period = program.pe_output_period
 
-            if not Data.member:
-                data = Data(screen, members, relationship_map)
-                value = data.value()
-                unit = raw_input["household"][data.unit][data.sub_unit]
+            if issubclass(Data, Member):
```
myang1010, line 118: I'm having trouble understanding the logic here. Just my guess about `Data`, `pe_inputs` and `pe_outputs`.
`pe_inputs` and `pe_outputs` are something look like this`dependency.member.AgeDependency`, which hold the fields that policy engine looks for, like "age", "is_child" and "is_tax_unit_dependent", etc.
Don't know how to define `Data`, but should be something like this `dependency.member.AgeDependency` as well?
And when we check `if issubclass(Data, Member):`, the "member" in `dependency.member.AgeDependency` gives `True`, so we update everything under `raw_input`, but the "tax_units"? 
If "[data.unit]" is Member, then what does the "raw_input["household"][data.unit]" in  `unit = raw_input["household"][data.unit][member_id]` looks like?
Thanks!

programs/programs/policyengine/policy_engine.py
```diff
@@ -105,35 +115,44 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator]):
...
-                if data.field in unit and period in unit[data.field]:
-                    if value != unit[data.field][period]:
-                        raise DependencyError(data.field, value, unit[data.field][period])
+    return raw_input
 
-                if data.field not in unit:
-                    unit[data.field] = {}
 
-                unit[data.field][period] = value
+def update_unit(unit, data: PolicyEngineCalulator, period: str):
```
myang1010, line 149: What does this function do? Is `period` a year or month?

myang1010 (commented):
Code looks great to me. Just have some questions. Thanks

programs/programs/co/rtdlive/calculator.py
```diff
@@ -23,27 +23,40 @@ class RtdLive(ProgramCalculator):
...
-        income_types = ["all"]
-        fpl = self.program.fpl.as_dict()
-        income_limit = RtdLive.percent_of_fpl * fpl[len(members)]
+        def income_eligible(member: HouseholdMember):
+            if member in main_tax_unit_members:
+                tax_unit = main_tax_unit_members
+            elif member in secondary_tax_unit_members:
+                tax_unit = secondary_tax_unit_members
+            else:
+                raise Exception("member is not in a tax unit")
```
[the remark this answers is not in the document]
CalebPena: No, Every member can only have one tax unit.

programs/programs/policyengine/policy_engine.py
CalebPena: `Data` is a subclass of `PolicyEngineScreenInput`. `pe_inputs` and `pe_outputs` are lists of `Data`. The `issubclass` logic is used to check if the `Data` is a `Member` dependency because they are handled differently. The way PE handles multiple members is like:
```py
{
  "people": {
    "member_id_one": {...}, # member 1 data
    "member_id_two": {...}, # member 2 data
    ...
  }
}
```

programs/programs/policyengine/policy_engine.py
CalebPena: The period is the year.

This function adds the value of data calculated with the `value` method, and adds it to the `unit` in the raw data. e.x. that takes in a member unit, and adds the earned income. In this example `data.value = 100` and `data.field = "earned_income"`

```py
update_unit({"age": {"2024": 44}}, data, "2024")

# will make the unit:
{"age": {"2024": 44}, "earned_income": {"2024": 100}}
```

programs/programs/policyengine/policy_engine.py
myang1010: Thanks for taking the time answering my question! Giving me some idea of how our screener works with PE! Thanks!

myang1010 (approved):

ivonne-hernandez (approved):

PR #518, opened 2024-08-28
LEAP income limits spread sheet

CalebPena:
- Pull LEAP income limits from a Google sheet.

### DIFF
--- a/programs/programs/co/energy_assistance/calculator.py
+++ b/programs/programs/co/energy_assistance/calculator.py
@@ -1,3 +1,4 @@
+from integrations.services.sheets.sheets import GoogleSheetsCache
 from integrations.util.cache import Cache
 from integrations.services.sheets import GoogleSheets
 from programs.programs.calc import ProgramCalculator, Eligibility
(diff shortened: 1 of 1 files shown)


programs/programs/co/energy_assistance/calculator.py
```diff
@@ -20,46 +21,39 @@ def update(self):
...
 
-    def value(self, eligible_members: int):
-        if self.screen.has_expense(["mortgage"]):
-            return DenverPropertyTaxRelief.mortgage_amount
-        elif self.screen.has_expense(["rent"]):
-            return DenverPropertyTaxRelief.rent_amount
-        return 0
-
-        return e
-
```
myang1010, line 62: If we remove the value here, does it mean no matter the members have mortgage or rent, they'll only get the same amount from `county_values`? Thanks.

myang1010 (commented):
Code looks good. Not sure about how the value listed in the spreadsheet, I assume it's something like "x,xxx". Just have one question on the removed "value" method. Thanks.

myang1010 (approved):

PR #571, opened 2024-09-27
State specific SNAP spm calculators

msrezaie:
- Abstracted snap program's PolicyEngine SPM calculator to be state specific, there are now `NcSnap` and `CoSnap` spm classes.

- The snap program's name abbreviated field needs to be changed to a state specific name through the dashboard. For NC it should be named `nc_snap` and for CO `co_snap`.
- Context for NC specific change: This PR also addresses the inaccurate results the screener was showing if a household was eligible for the snap program's [BUA or LUA](https://www.snapscreener.com/data#:~:text=If%20a%20household%20pays%20for%20two%20or%20more%20utility%20bills%2C%20this%20deduction%20amount%20can%20be%20claimed%20(but%20no%20other%20utility%20deductions)) (Basic or Limited utility allowance). According to the [snap screener](https://www.snapscreener.com/screener/north-carolina) and [PolicyEngine](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/main.yaml#L1), a household becomes eligible for LUA if they are paying at least two utility expenses (i.e. water, gas and fuel, electricity, sewage, phone) besides heating or cooling. For the state of North Carolina, there is no additional single utility allowance given besides a $41 allowance for phone expense. Currently in the screener, in order for a household to become eligible for a LUA deduction, they need to input 'Other Utilities' and 'Telephone' expense options in step-6 (other than rent). `PhoneExpenseDependency` was added to the `NcSnap` class's dependencies list in order to send the 'Telephone' expense amount to the PolicyEngine. The [SPM unit](https://github.com/Gary-Community-Ventures/benefits-api/blob/feat/seperate-snap/programs/programs/policyengine/calculators/dependencies/spm.py#L122) class `ElectricityExpenseDependency` responsible for processing the 'Other Utilites' expense option is mapped to a `electricity_expense` PolicyEngine variable. From my testing I wasn't able to make this dependency take effect. By replacing this dependency with another SPM unit class, `WaterExpenseDependency`, and mapping it to the `water_expense` PolicyEngine variable—which is calculated the same way as `electricity_expense`—with these changes, the estimated $ value returned from PolicyEngine whenever a household is eligible for LUA matches the results shown in the snap screener.

### DIFF
--- a/programs/programs/co/pe/spm.py
+++ b/programs/programs/co/pe/spm.py
@@ -5,7 +5,6 @@ from programs.programs.federal.pe.spm import Snap, Tanf
 class CoSnap(Snap):
     pe_inputs = [
         *Snap.pe_inputs,
-        dependency.spm.ElectricityExpenseDependency,
         dependency.household.CoStateCode,
     ]
 
--- a/programs/programs/co/pe/spm.py
+++ b/programs/programs/co/pe/spm.py
@@ -6,6 +6,7 @@ class CoSnap(Snap):
     pe_inputs = [
         *Snap.pe_inputs,
         dependency.spm.ElectricityExpenseDependency,
+        dependency.household.CoStateCode,
     ]
 
 
--- a/programs/programs/nc/pe/spm.py
+++ b/programs/programs/nc/pe/spm.py
@@ -2,8 +2,9 @@ import programs.programs.policyengine.calculators.dependencies as dependency
 from programs.programs.federal.pe.spm import Snap
 
 
-class CoSnap(Snap):
+class NcSnap(Snap):
     pe_inputs = [
         *Snap.pe_inputs,
-        dependency.spm.ElectricityExpenseDependency,
+        dependency.spm.WaterExpenseDependency,
+        dependency.spm.PhoneExpenseDependency,
     ]
--- a/programs/programs/nc/pe/spm.py
+++ b/programs/programs/nc/pe/spm.py
@@ -7,6 +7,7 @@ class NcSnap(Snap):
         *Snap.pe_inputs,
         dependency.spm.WaterExpenseDependency,
         dependency.spm.PhoneExpenseDependency,
+        dependency.household.NcStateCode,
     ]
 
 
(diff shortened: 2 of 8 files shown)


programs/programs/nc/pe/spm.py
```diff
@@ -1,5 +1,13 @@
 import programs.programs.policyengine.calculators.dependencies as dependency
-from programs.programs.federal.pe.spm import Tanf
+from programs.programs.federal.pe.spm import Snap, Tanf
+
+
+class NcSnap(Snap):
+    pe_inputs = [
+        *Snap.pe_inputs,
+        dependency.spm.WaterExpenseDependency,
+        dependency.spm.PhoneExpenseDependency,
```
CalebPena, line 9: Do these expenses not apply for CO?

programs/programs/co/pe/spm.py
```diff
@@ -1,5 +1,12 @@
 import programs.programs.policyengine.calculators.dependencies as dependency
-from programs.programs.federal.pe.spm import Tanf
+from programs.programs.federal.pe.spm import Snap, Tanf
+
+
+class CoSnap(Snap):
+    pe_inputs = [
+        *Snap.pe_inputs,
+        dependency.spm.ElectricityExpenseDependency,
```
CalebPena, line 8: Does this not apply for NC.

programs/programs/co/pe/spm.py
```diff
@@ -1,5 +1,12 @@
...
+    ]
```
CalebPena, line 9:
```suggestion
        dependency.household.CoStateCode,
    ]
```

programs/programs/nc/pe/spm.py
```diff
@@ -1,5 +1,13 @@
...
+    ]
```
CalebPena, line 10:
```suggestion
        dependency.household.NcStateCode,
    ]
```

CalebPena (commented):
If the water, phone, and electricity expenses apply in both CO and NC, then it should be moved to the federal calculator.

CalebPena (approved):

PR #581, opened 2024-10-04
SNAP medical expense deduction

msrezaie:
- Added PolicyEngine dependencies for calculating the SNAP program's medical expense deductions for a household that includes at least one elderly or disabled person.
- Mapped the 'Medical Insurance Premium &/or Bills' from the expense options for inputting out-of-pocket medical expenses to the screener.

### DIFF
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -108,8 +108,7 @@ class MedicalExpenseDependency(Member):
 
     def value(self):
         elderly_or_disabled_members = [
-            member for member in self.screen.household_members.all()
-            if member.age >= 60 or member.has_disability()
+            member for member in self.screen.household_members.all() if member.age >= 60 or member.has_disability()
         ]
         count_of_elderly_or_disabled_members = len(elderly_or_disabled_members)
 
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -100,20 +100,6 @@ class IsDisabledDependency(Member):
         return self.member.disabled or self.member.long_term_disability
 
 
-class IsUSDADisabledDependency(Member):
-    field = "is_usda_disabled"
-
-    def value(self):
-        return self.member.has_disability()
-
-
-class IsUSDAElderlyDependency(Member):
-    field = "is_usda_elderly"
-
-    def value(self):
-        return self.member.is_elderly()
-
-
 # In this class a dynamic attribute called '_medical_expense_assigned_member_id' is used
 # to ensure that medical expenses are only counted once per household, even when multiple
 # elderly or disabled members are in the same household. This is done because the Member
@@ -123,7 +109,7 @@ class MedicalExpenseDependency(Member):
     field = "medical_out_of_pocket_expenses"
 
     def value(self):
-        if self.member.is_elderly() or self.member.has_disability():
+        if self.member.age >= 60 or self.member.has_disability():
             assigned_member_id = getattr(self.screen, "_medical_expense_assigned_member_id", None)
             if assigned_member_id is None:
                 self.screen._medical_expense_assigned_member_id = self.member.id
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -100,24 +100,21 @@ class IsDisabledDependency(Member):
         return self.member.disabled or self.member.long_term_disability
 
 
-# In this class a dynamic attribute called '_medical_expense_assigned_member_id' is used
-# to ensure that medical expenses are only counted once per household, even when multiple
-# elderly or disabled members are in the same household. This is done because the Member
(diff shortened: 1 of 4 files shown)


programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -100,6 +100,42 @@ def value(self):
...
+            assigned_member_id = getattr(self.screen, "_medical_expense_assigned_member_id", None)
+            if assigned_member_id is None:
+                self.screen._medical_expense_assigned_member_id = self.member.id
+                return self.screen.calc_expenses("yearly", ["medical"])
+            elif assigned_member_id == self.member.id:
+                return self.screen.calc_expenses("yearly", ["medical"])
+            else:
+                return 0
+        else:
+            return 0
```
CalebPena, line 136: This will not work because the medical expense deduction could be attached to someone who is not elderly or disabled. What you need to do is `self.screen.calc_expenses("yearly", ["medical"]) / count_of_elderly_or_disabled_members`. I assume that using the count of elderly or disabled members is the most accurate assumption to make because those are the members of the household where this expense matters for SNAP purposes. You could also divide the medical expenses by all members, or by the number of adults in the household.

screener/models.py
```diff
@@ -446,6 +446,9 @@ def is_married(self):
     def has_disability(self):
         return self.disabled or self.visually_impaired or self.long_term_disability
 
+    def is_elderly(self):
+        return self.age >= 60
```
CalebPena, line 450: Different programs define elderly differently. Could you move this logic out of the member model. i.e. just call `member.age >= 60` in `HasUsdaElderlyDisabledDependency`.

programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -100,6 +100,42 @@ def value(self):
...
+
+    def value(self):
+        return self.member.has_disability()
+
+
+class IsUSDAElderlyDependency(Member):
+    field = "is_usda_elderly"
+
+    def value(self):
+        return self.member.is_elderly()
```
CalebPena, line 114: Did you have trouble with these? They should be calculated from the age and disabled parameters we send to PE. We might need to add those as dependencies to SNAP. We typically like to give PE the most root variable, so if changes are made to the rules on PE's side, we don't have to also make a change on our side.

CalebPena (changes requested):

programs/programs/policyengine/calculators/dependencies/member.py
msrezaie: In the SNAP [screener](https://www.snapscreener.com/screener/north-carolina), the 'out of pocket medical expenses' option only comes up when the option of to include a disabled/elderly person is enabled. So I think the medical expense bill only counts whenever a household has an elderly or disabled person?

> What you need to do is `self.screen.calc_expenses("yearly", ["medical"]) / count_of_elderly_or_disabled_members`. I assume that using the count of elderly or disabled members is the most accurate assumption to make because those are the members of the household where this expense matters for SNAP purposes.

The if condition essentially does that. If there is at least one disabled or elderly member, then the medical expense is sent to PE.

programs/programs/policyengine/calculators/dependencies/member.py
msrezaie: Yeah, you are right. Will refactor this out.

programs/programs/policyengine/calculators/dependencies/spm.py
```diff
@@ -105,6 +105,13 @@ def value(self):
         return self.screen.has_expense(["telephone"])
 
 
+class HasUsdaElderlyDisabledDependency(SpmUnit):
+    field = "has_usda_elderly_disabled"
+
+    def value(self):
+        return any(member.age >= 60 or member.has_disability() for member in self.members)
+
+
```
CalebPena, line 114: You can remove this and replace it with the member age dependency because PE will use the age variable, and the disabled variable to calculate this.

programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -100,6 +100,24 @@ def value(self):
...
+    def value(self):
+        elderly_or_disabled_members = [
+            member for member in self.screen.household_members.all() if member.age >= 60 or member.has_disability()
+        ]
+        count_of_elderly_or_disabled_members = len(elderly_or_disabled_members)
+
+        if self.member.age >= 60 or self.member.has_disability():
+            return self.screen.calc_expenses("yearly", ["medical"]) / count_of_elderly_or_disabled_members
+        else:
+            return 0
```
CalebPena, line 118:
```suggestion

        return 0
```

CalebPena (approved):
Looks good. Just a style change, and a suggestion to let PE calculate the `has_usda_elderly_disabled` variable based on age and disability of the members.

msrezaie:
Linking test document with different household scenarios:
https://docs.google.com/document/d/1Oc062F8xm-lhnb4DNTi-7Adbu-GKlp_LI8BeLVpMDeY/edit?tab=t.0#heading=h.j47490lqppdj

PR #582, opened 2024-10-08
Feature: Brevo Integration

mexi-cano:
- Added Brevo contact management 

- `.env` file needs to have contact management platform set and a corresponding api key 

`# CONTACT_SERVICE options: hubspot or brevo
CONTACT_SERVICE=brevo
HUBSPOT_API_KEY = 'your_hubspot_api_key'
BREVO_API_KEY='your_brevo_api_key'`


CalebPena (changes requested):
We want to make an interface where we can substitute different CMS integrations. It will look something like:
```
class CmsIntegration:
    def __init__(self, user, screen):
        self.user = user
        self.screen = screen

    def add(self):
        # adding a new contact
        raise NotImplementedError()

    def update(self):
        # updating the consent on a contact
        raise NotImplementedError()

    def shout_add(self):
        # additional conditions to determine if we should add the user to the CMS
        # for example, one of us might want to add tests, while the other does not
        return True


class HubSpotIntegration(CmsIntegration):
    def add(self):
        # I will implement this for Gary
        ...

    def update(self):
        # I will implement this for Gary
        ...


class BrevoIntegration(CmsIntegration):
    def add(self):
        # you implement this for Brevo
        ...

    def update(self):
        # you implement this for Brevo
        ...


def get_cms_integration():
    if settings.CONTACT_SERVICE == "brevo":
        return BrevoIntegration

    return HubSpotIntegration
```
And it would be used like:
```
def new_user(user, screen):
    Integration = get_cms_integration()
    integration = Integration(user, screen)

    if not integration.shout_add():
        return

    if user_already_exists_but_we_want_to_update_the_consent:
        integration.update()
    else:
        integration.add()
```

Also, we already have a class to send messages with Twilio and SendGrid, so you don't need to make those. Lines 49-53 in `authentication/views.py` send text and email with the results link. All you need to do is provide your API keys.

authentication/views.py
```diff
@@ -24,33 +26,45 @@ def update(self, request, pk=None):
...
+                message.email(screen.user.email)
+            if screen.user.cell is not None:
+                message.text(str(screen.user.cell))
+
+            if serializer.is_valid():
+                screen.user = serializer.save()
+                screen.save()
+
+                Integration = get_cms_integration()
+                integration = Integration(user, screen)
```
CalebPena, line 49: You did this on lines 32-36 already
```suggestion
```

integrations/services/brevo/brevo.py
```diff
@@ -0,0 +1,100 @@
...
+from django.utils import timezone
+from screener.models import Message, Screen
+from translations.models import Translation
+import uuid
+import json
+from pprint import pprint
+from sentry_sdk import capture_message
+
+
+class BrevoService:
```
CalebPena, line 13: Can we move this logic into the `BrevoIntegration` class?

integrations/services/cms_integration.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class HubSpotIntegration(CmsIntegration):
+    def add(self):
+        # Implement the logic for adding a user to HubSpot
+        pass
+
+    def update(self):
+        # Implement the logic for updating a user in HubSpot
+        pass
```
CalebPena, line 29: I am going to make a pr to implement this, and set the base branch to this branch.

authentication/views.py
```diff
@@ -24,33 +26,45 @@ def update(self, request, pk=None):
...
+
+                return Response(status=204)
         return Response(serializer.errors, status=400)
 
 
+def update_brevo(external_id, send_offers, send_updates):
+    brevo_service = BrevoService()
+    brevo_service.update_contact(external_id, {"send_offers": send_offers, "send_updates": send_updates})
+
+
```
CalebPena, line 67: Where do you call this?

CalebPena (changes requested):
The structure looks good for the most part, but I left some comments to tidy up some things. I will make a pr to be merged with this one to implement the `HubSpotIntegration`.

integrations/services/communications/message.py
```diff
@@ -61,6 +64,7 @@ def _email_body(self):
 
     def text(self, cell: str, send_tests=False):
         if not self.should_send() and not send_tests:
+            print("should not sent is true. returning")
```
CalebPena, line 67: remove this

integrations/services/brevo/brevo.py
CalebPena, line 1: We should be able to delete this file now.

integrations/services/communications/message.py
```diff
@@ -43,7 +43,10 @@ def email(self, email: str, send_tests=False):
         content = Content("text/html", self._email_body())
         mail = Mail(from_email, to_email, subject, content)
 
-        sg.client.mail.send.post(request_body=mail.get())
+        try:
+            sg.client.mail.send.post(request_body=mail.get())
+        except Exception as e:
+            print("e:", e)
```
CalebPena, line 49: I think that we should just let this error fall through.
```suggestion
        sg.client.mail.send.post(request_body=mail.get())
```

CalebPena (approved):

CalebPena:
@mexi-cano I am going to merge this pr because there are only a couple of things to clean up. Could you make another pr with the changes requested?

PR #604, opened 2024-10-24
Count pregnant people as 2 for Medicaid and CHP+

CalebPena:
- Count pregnant people as 2 when calculating FPL for Medicaid and CHP+


civicpixel (changes requested):
Typo in Children on ExpectedChilrenPregnancyDependency variable.

PR #609, opened 2024-10-29
Health check

CalebPena:
- Add `health_check` script.
- Check if the links are valid.
- Check that you can't write to HubSpot.
- Check that there is no PII in the DB.

- If `no_auto`, the other fields fill with English.
- Make links be set to `no_auto` by default.

- There are a bunch of broken links in other languages.


integrations/management/commands/health_check.py
```diff
@@ -0,0 +1,139 @@
...
+
+    HUB_SPOT_TEXT = "Can't read Hub Spot"
+    PII_IN_DB_TEXT = "PII not in the database"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "-s",
+            "--strict",
+            action=argparse.BooleanOptionalAction,
+            help="Should the he website hashes should be compared to the old hashes",
```
ivonne-hernandez, line 27: Is this supposed to say "The website hashes should be compared to the old hashes"?

ivonne-hernandez (approved):

PR #612, opened 2024-10-30
Document Links

kbrey:
- Add link fields to program documents, `link_url` and `link_text`, that are used to append a link to the end of the document `text` in the screener
- Populates existing documents with new translations in migration
- New and existing documents will have translations created, but with empty strings values so that nothing is shown in the screener unless the links field are populated.

- Updated translation admin views and menus with the new fields
- Updated Document `__str__` method to use `external_name`, with default to `text`, so that the `external_name` is shown when managing program documents in the admin app


programs/models.py
```diff
@@ -186,13 +186,19 @@ def create_instance(cls, external_name: str, Model: type["Document"]) -> "Docume
 class Document(models.Model):
     external_name = models.CharField(max_length=120, blank=True, null=True, unique=True)
     text = models.ForeignKey(Translation, related_name="documents", blank=False, null=False, on_delete=models.PROTECT)
+    link_url = models.ForeignKey(
+        Translation, related_name="document_link_url", blank=True, null=True, on_delete=models.PROTECT
+    )
+    link_text = models.ForeignKey(
+        Translation, related_name="document_link_text", blank=True, null=True, on_delete=models.PROTECT
+    )
```
CalebPena, line 194: We use empty translations for empty fields instead of `null` because it makes it so that we don't have to recreate the translation object if we want to create a link.

CalebPena (changes requested):
We also need to update the admin UI for this. I would be happy to schedule a time to go over how the translation admin works.

programs/models.py
```diff
@@ -162,15 +162,24 @@ def __str__(self):
 
 
 class DocumentManager(models.Manager):
-    translated_fields = ("text",)
+    translated_fields = ("text", "link_url", "link_text")
+    no_auto_fields = "link_url"
```
CalebPena, line 166: This needs to be a list.
```suggestion
    no_auto_fields = ("link_url",)
```

CalebPena (changes requested):

programs/models.py
CalebPena: I am just going to commit this because it is one line, so we can release it today.

CalebPena (approved):

PR #650, opened 2024-11-20
White label logic

CalebPena:
- Filter programs based on the screen white label.
- Filter urgent needs based on the screen white label.
- Filter web hooks based on the screen while label.
- Pick the CMS integration based on the screen white label.
- Add white label select when adding something in the translation admin.
- Add `stepDirectory` to referrerData in the config.

### Review and merge https://github.com/Gary-Community-Ventures/benefits-api/pull/646 first


CalebPena:
@kbrey Please take a look at the changes to the config, so y'all can implement them on your end.

ivonne-hernandez (approved):

configuration/admin.py
```diff
@@ -12,7 +12,14 @@ class ConfigurationAdmin(ModelAdmin):
             "widget": JSONEditorWidget(options={"modes": ["tree", "code"], "mode": "tree", "enableDrag": False})
         },
     }
-    search_fields = ("name",)
+    search_fields = ("name", "white_label__name")
```
kbrey, line 15: Does "white_label__name" have an extra underscore here?

configuration/admin.py
```diff
@@ -12,7 +12,14 @@ class ConfigurationAdmin(ModelAdmin):
...
+    list_display = ("name", "white_label_name")
+
+    def white_label_name(self, obj):
+        return obj.white_label.name
+
+    white_label_name.admin_order_field = "white_label__name"
```
kbrey, line 21: Same question here, is there an extra underscore?

configuration/white_labels/nc.py
```diff
@@ -0,0 +1,9 @@
+from .base import ConfigurationData
+from screener.models import WhiteLabel
+
+
+# TODO: Update NC configuration
```
kbrey, line 5: Is this something our team will be responsible for?

configuration/admin.py
CalebPena: Yes, the syntax in Django for filtering by a nested field is to use a double underscore. This is saying to search with the `name` field on the `white_label` model.

PR #654, opened 2024-12-02
Add NC Sun Bucks

msrezaie:
- Added a new program calculator for NC specific program Sun Bucks.
- Added a new field `has_sunbucks` in the Screener model to show that a screen has the Sun Bucks benefit.

-  In order to make use of this program, it must first be added through the Translations Admin with the `abbreviated_name` of `sunbucks`.
- Run `python manage.py migrate` to add the new model changes.


programs/programs/nc/sun_bucks/calculator.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def member_eligible(self, e: MemberEligibility):
+        member = e.member
+
+        # age eligibility
+        child_eligible = False
+        if member.age >= SunBucks.min_age and member.age <= SunBucks.max_age:
+            child_eligible = True
+
+        e.condition(child_eligible)
```
CalebPena, line 28: This syntax works in python. Feel free to discard if you find the other way of doing it more readable.
```suggestion
        e.condition(SunBucks.min_age <= member.age <= SubBucks.max_age)
```

CalebPena (approved):
Looks good. I left on optional suggestion.

screener/models.py
```diff
@@ -281,6 +282,7 @@ def has_benefit(self, name_abbreviated):
             "wic": self.has_wic,
             "nc_wic": self.has_wic,
             "snap": self.has_snap,
+            "sunbucks": self.has_sunbucks or self.has_tanf or self.has_snap or self.has_medicaid or self.has_medicaid_hi,
```
CalebPena: Can you move `self.has_tanf or self.has_snap or self.has_medicaid or self.has_medicaid_hi` to the `SunBucks` calculator. this method it meant to determine if the user already has the benefit. If they have TANF, SNAP, or Medicaid, they are not eligible for the program which is different than the use already having the benefit. An example of the calculator accounting for benefits the user already is eligible for is [EmergencyMedicaid](https://github.com/Gary-Community-Ventures/benefits-api/blob/main/programs/programs/co/medicaid/emergency/calculator.py).

CalebPena (changes requested):

screener/models.py
msrezaie: > this method it meant to determine if the user already has the benefit

This program (as per its [ticket's](https://github.com/Gary-Community-Ventures/benefits-api/issues/644) requirements) should show up if the user hasn't inputted in step 8 that they are already receiving TANF, SNAP, Medicaid, or Sun bucks itself.

CalebPena: The `has_benefit` method is used if the user already has a benefit. If the user has TANF, they do not have Sun Bucks. Instead, call `self.screen.has_benefit('tanf')` in the `SunBucks` calculator.

CalebPena (approved):

PR #660, opened 2024-12-04
UPK updates

CalebPena:
- Add UPK warning message for 3 year-olds

- Update UPK value.
- Update UPK qualifying conditions.
- Added eligibility param to the `WarningCalculator` class.

### DIFF
--- a/programs/programs/co/universal_preschool/calculator.py
+++ b/programs/programs/co/universal_preschool/calculator.py
@@ -5,8 +5,11 @@ from screener.models import HouseholdMember
 class UniversalPreschool(ProgramCalculator):
     qualifying_age = 3
     age = 4
-    percent_of_fpl = 2.7
-    amount_by_hours = {"10_hours": 4_837, "15_hours": 6_044, "30_hours": 10_655}
+    income_limit = 1
+    foster_income_limit = 2.7
+    amount_10_hr = 4_920
+    amount_15_hr = 6_204
+    amount_30_hr = 11_004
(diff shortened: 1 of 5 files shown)


programs/programs/co/universal_preschool/calculator.py
```diff
@@ -23,16 +26,20 @@ def member_value(self, member: HouseholdMember):
...
+        fpl = self.program.fpl.as_dict()[self.screen.household_size]
+        income = self.screen.calc_gross_income("yearly", ["all"])
 
-        return income_condition or member.relationship == "fosterChild"
+        income_limit = int(UniversalPreschool.income_limit * fpl)
+
+        if member.relationship == "fosterChild":
+            income_limit = int(UniversalPreschool.foster_income_limit * fpl)
+
+        return income < income_limit
```
ivonne-hernandez, line 45: Does the `income` have to be strictly less than the `income_limit` or can it also be equal to the `income_limit`?

ivonne-hernandez (changes requested):
Please see the question regarding the strictly less than the `income_limit` expression.

programs/programs/co/universal_preschool/calculator.py
CalebPena: I don't know. I just copied what we had previously. I am going to update it to allow it to be equal to because that is what we usually do.

CalebPena, line 45:
```suggestion
        return income <= income_limit
```

ivonne-hernandez (approved):

PR #680, opened 2024-12-20
SNAP child support

CalebPena:
- Fix SNAP child support.
- Add white label to the printed validation url.


msrezaie (approved):
LGTM

PR #692, opened 2025-01-03
Exclude cash assistance from certain programs

CalebPena:
- Add `exclude` param to `calc_gross_income` method.

- Refactor some programs to exclude cash assistance income.
- Rename the cash assistance income.

@msrezaie FYI, in case y'all want to do the same for some of the NC programs. The tax credits and Medicaid programs already had it excluded.


msrezaie (approved):
LGTM

PR #714, opened 2025-01-23
Add LIEAP Program to NC Screener

IrynaKolh:
ISSUE https://github.com/Gary-Community-Ventures/benefits-api/issues/706
- Add LIEAP Program to NC Screener
image


programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+        7: [2564, 5129],
+        8: [2856, 5711],
+        9: [3147, 6294],
+        10: [3439, 6877],
+        11: [3730, 7460],
+        12: [4021, 8043],
+        13: [4313, 8626],
+        14: [4604, 9208],
+        15: [4896, 9791],
+    }
```
CalebPena, line 27: The FPL is a common thing in programs, so we have a property on `self.program` that has the FPL which you can then multiply to get the income limit. You just have to select the correct year in the admin, and then call `self.program.fpl.as_dict()` and you will get the FPLs for the household sizes 1-8. Also, the max household size is 8, so you don't have to worry about any households with more than 8 members. Also, the FPL is a yearly value, so you will need to divide by 12 (or use the yearly income).

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def household_eligible(self, e: Eligibility):
+        household_size = self.screen.household_size
+        gross_income = self.screen.calc_gross_income("monthly", ["all"])
+
+        # has rent or mortgage expense
+        has_rent = self.screen.has_expense(["rent"])
+        has_mortgage = self.screen.has_expense(["mortgage"])
+        e.condition(has_rent or has_mortgage)
```
CalebPena, line 36: You can add multiple expenses
```suggestion
        has_rent_or_mortgage = self.screen.has_expense(["rent", "mortgage"])
        e.condition(has_rent_or_mortgage)
```

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+        11: [3730, 7460],
+        12: [4021, 8043],
+        13: [4313, 8626],
+        14: [4604, 9208],
+        15: [4896, 9791],
+    }
+
+    def household_eligible(self, e: Eligibility):
+        household_size = self.screen.household_size
+        gross_income = self.screen.calc_gross_income("monthly", ["all"])
```
CalebPena, line 31: Can you move these next to the income condition so that all of the logic is in one place?

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # has rent or mortgage expense
+        has_rent = self.screen.has_expense(["rent"])
+        has_mortgage = self.screen.has_expense(["mortgage"])
+        e.condition(has_rent or has_mortgage)
+
+        # income
+        if household_size:
```
CalebPena, line 39: You don't need this `if`. If `household_size` is `None`, then this program will be skipped because it has `household_size` as one of the dependencies.

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+            income_limit = int(self.fpl_percent * self.income_limits[household_size][1])
+
+        e.condition(gross_income < income_limit, messages.income(gross_income, income_limit))
+
+    def household_value(self):
+        household_size = self.screen.household_size
+        gross_income = self.screen.calc_gross_income("monthly", ["all"])
+
+        if household_size:
```
CalebPena, line 48: You can remove this `if` as well.

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+            if household_size <= 3:
+                if gross_income <= self.income_limits[household_size][0]:  # 0-50% FPL
+                    return 400
+                elif gross_income <= self.income_limits[household_size][1]:  # 51%-130% FPL
+                    return 300
+            else:  # Household size >= 4
+                if gross_income <= self.income_limits[household_size][0]:  # 0-50% FPL
+                    return 500
+                elif gross_income <= self.income_limits[household_size][1]:  # 51%-130% FPL
+                    return 400
```
CalebPena, line 58: Can you add the values and household size limit as class variables. Like this (alter the names as you wish):
```python
fpl = self.program.fpl.as_dict()[household_size]
if household_size < self.large_household_size:
    if gross_income <= fpl * self.max_value_fpl_percent:  # 0-50% FPL
        return self.small_household_low_income_value
...
```

CalebPena (changes requested):

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,49 @@
...
+    fpl_percent = 1.3
+    resource_limit = 2250
+    dependencies = [
+        "income_frequency",
+        "income_amount",
+        "zipcode",
+        "household_size",
+        "age",
+    ]
+    large_household_size = 3
```
CalebPena, line 16: Based on the code `if household_size <= self.large_household_size:`, I think that the large household size is 4. You can then change your code to be `if household_size < self.large_household_size:`. I think that this is more accurate to the name of the variable because a household size of 3 is not the large household size, it is the biggest small household size.
```suggestion
    large_household_size = 4
```

msrezaie (approved):
After a small change, everything else looks good. Great work on your first API ticket.

programs/programs/nc/nc_lieap/calculator.py
```diff
@@ -0,0 +1,48 @@
...
+    small_household_low_income_value = 400
+    small_household_large_income_value = 300
+    large_household_low_income_value = 500
+    large_household_large_income_value = 400
+
+    def household_eligible(self, e: Eligibility):
+        household_size = self.screen.household_size
+
+        # has rent or mortgage expense
+        has_rent_or_mortgage = self.screen.has_expense(["rent", "mortgage"])
```
msrezaie, line 26: Under the class you've added a `expenses` list, but in here you've used individual expenses for the `has_expense` method. You can either use the `expenses` variable like so:
```
has_rent_or_mortgage = self.screen.has_expense(NCLieap.expenses)
```
or remove the `expenses` variable and just add the 'heating' as an item to the list:
```
has_rent_or_mortgage = self.screen.has_expense(["rent", "mortgage", "heating"])
```

msrezaie (changes requested):

programs/programs/nc/nc_lieap/calculator.py
IrynaKolh: fixed

kbrey: Nice suggestion @msrezaie! I prefer the first approach using the class variable, because I think it makes the code easier to read and understand. Also, we'll want to update the comment that says "# has rent or mortgage expense" because we're also including heating.

kbrey (changes requested):
Great work @IrynaKolh! I have one additional request, please add nc_lieap to the nc.py file (configuration/white_labels/nc.py). This will ensure nc_lieap is included when running add_config for the nc white label.

configuration/white_labels/nc.py
```diff
@@ -1569,13 +1569,13 @@ def get_white_label(self) -> WhiteLabel:
...
+                "nc_lieap": {
                     "name": {
-                        "_label": "housingAndUtilities.leap",
-                        "_default_message": "Low-Income Energy Assistance Program (LEAP): ",
+                        "_label": "housingAndUtilities.nc_lieap",
+                        "_default_message": "Low Income Energy Assistance Program (LIEAP): ",
                     },
                     "description": {
-                        "_label": "housingAndUtilities.leap_desc",
+                        "_label": "housingAndUtilities.nc_leap_desc",
```
kbrey, line 1578: Tiny request, this should be "nc_lieap_desc"

kbrey (approved):
Well done! Looks great to me.

PR #737, opened 2025-02-03
Added white_label to the str() method in all of the program models

SonaliBedge:
-Added white_label to the str() method in all of the program models

image


CalebPena (approved):
Looks good. I added updated a couple of additional models so we could release this today.

PR #742, opened 2025-02-06
Refactor/renamed Program 'fpl' field to 'year'

msrezaie:
- The `fpl` field of the Programs model is renamed to `year`.

Renamed the `fpl` field to `year` in all the instances of every file that I could find and made sense in performing this refactor where `program` model's `fpl` field is used. Also tested `bulk_import` command to make sure that the refactor is not causing any problems when used for setting up programs data. Please feel free to point out any other files/places where I may have missed renaming `fpl` to `year`.


CalebPena (approved):
LGTM

PR #791, opened 2025-02-28
Feat/758 fix counties within config script for nc

SonaliBedge:
-Added all missing counties with their zip codes


CalebPena (approved):
LGTM

PR #799, opened 2025-03-04
Add state option to the configuration

CalebPena:
- Add state name option to config.


kbrey (approved):
Looks good to me!

PR #830, opened 2025-03-14
Feat/Script to add County Names

msrezaie:
- Added a `program` command script for adding a state's county names from `counties_by_zipcode` config object. This script can be used with the following terminal command:
```
python manage.py add_counties "white_label code"
```
For example:

```
python manage.py add_counties nc
```


programs/management/commands/add_nc_counties.py
```diff
@@ -0,0 +1,124 @@
...
+            "Wayne County",
+            "Wilkes County",
+            "Wilson County",
+            "Yadkin County",
+            "Yancey County",
+        ]
+
+        white_label = self.get_white_label()
+        if not white_label:
+            self.stdout.write(self.style.WARNING(f"NC White label does not exist"))
```
CalebPena, line 118: I don't think that this will work because get will throw an error if it does not exist.

CalebPena (changes requested):
Could you make this so that it pull the JSON from the config, and gets the counties like that, so we can use this for other states?

programs/management/commands/add_counties.py
```diff
@@ -0,0 +1,31 @@
...
+    help = "Adds counties"
+
+    def add_arguments(self, parser):
+        parser.add_argument("white_label", type=str, help="The state code for the white label")
+
+    def handle(self, *args, **options):
+        try:
+            white_label = WhiteLabel.objects.get(code=options["white_label"])
+        except ObjectDoesNotExist:
+            self.stdout.write(self.style.WARNING(f'White label for "{options["white_label"]}" is not in the database'))
```
CalebPena, line 18: We want to stop the script if there is an error.
```suggestion
            self.stdout.write(self.style.WARNING(f'White label for "{options["white_label"]}" is not in the database'))
            return
```

programs/management/commands/add_counties.py
```diff
@@ -0,0 +1,31 @@
...
+        try:
+            counties_from_config = Configuration.objects.get(name="counties_by_zipcode", white_label=white_label)
+
+            counties = list(set([key for value in eval(counties_from_config.data).values() for key in value.keys()]))
```
CalebPena, line 22: Lets not use `eval`. The config is JSON, so you should be able to use `json.loads`. Also, can you use a `for` loop; it is a little hard to read this.

CalebPena (changes requested):
Can you remove the eval?

CalebPena (approved):

PR #838, opened 2025-03-18
Explicit TCPA consent

CalebPena:
- Add a new field that will represent the new TCPA consent field.
- Change checkboxes on the sign up page.

**This applies to NC as well**


integrations/services/cms_integration.py
```diff
@@ -246,8 +247,10 @@ def _hubspot_contact_data(self):
             "states": "NC",
             "send_offers": self.user.send_offers,
             "tcpa_consent": self.user.tcpa_consent,
```
kbrey, line 249: Do we want to keep tcpa_consent now that there's explicit_tcpa_consent?

integrations/services/cms_integration.py
```diff
@@ -246,8 +247,10 @@ def _hubspot_contact_data(self):
...
+            "explicit_tcpa_consent": self.user.explicit_tcpa_consent,
             "send_updates": self.user.send_updates,
             "hs_language": self.user.language_code,
+            "hubspot_owner_id": "47185138",
```
kbrey, line 253: Wondering why this is here and if it needs to be moved into a class variable?

kbrey (commented):
Thanks @CalebPena , left a few questions.

CalebPena:
@kbrey I made the requested changes. Let me know if there is anything else.

Issue #841, opened 2025-03-19
Annual tax credit value at top of results screen not populating


Linda-Vu:
in this [test profile](https://screener.myfriendben.org/nc/77b35cd9-c842-4f7e-b3d4-2578cc25c27d/results/benefits), the top of the results page says I'm getting -0- in tax credits even though tax credits results show both the EITC and CTC; 

![Image](https://github.com/user-attachments/assets/e2ef2d6c-55be-4059-a49e-9cef90de6ad3)

IrynaKolh:
![Image](https://github.com/user-attachments/assets/e3a6c6bf-7aa9-4556-a3ca-b836ce54b3c4)

@Linda-Vu and @Lkp2104 I didn't get this error.... I checked calculation from main and dev branches several times. I also asked to check Mo and Sonali.

Linda-Vu:
appears to be resolved!

PR #848, opened 2025-03-20
#841 Annual tax credit value at top of results screen not populating

IrynaKolh:


CalebPena (approved):

programs/models.py
```diff
@@ -233,6 +236,7 @@ class ProgramCategory(models.Model):
     external_name = models.CharField(max_length=120, blank=True, null=True, unique=True)
     calculator = models.CharField(max_length=120, blank=True, null=True)
     icon = models.CharField(max_length=120, blank=False, null=False)
+    tax_category = models.BooleanField(default=False)
```
kbrey, line 239: Tiny nitpick but I suppose "tax_category" is redundant and we could have done `is_tax`?

kbrey (approved):
Have a suggestion but I don't think it needs to prevent the merge. Nice work!

programs/models.py
CalebPena: I agree. Can you make a ticket, so we don't lose this.

msrezaie (approved):

PR #875, opened 2025-04-08
Remove TCPA requirement from should_add method

CalebPena:
- Remove TCPA requirement from should_add method.


myang1010 (approved):
Looks good!

PR #876, opened 2025-04-09
Energy: Add more help options to energy calculator

CalebPena:
- Add more help options to energy calculator
- Add feature flag to not return the user to the confirmation page from the zip code step

**Maybe we should make this page a little prettier**
image


CalebPena:
@sydney-devine I added an energy calculator referrer, so we can know how many people use this link. Let me know if you want to remove. Also the link will always go to the prod instance, so be careful with that.

sydney-devine:
> @sydney-devine I added an energy calculator referrer, so we can know how many people use this link. Let me know if you want to remove. Also the link will always go to the prod instance, so be careful with that.

Thanks!

PR #892, opened 2025-04-17
Created a script to transfer programs from one white label to another…

SonaliBedge:
…, including their translations.

- The transfer.py script is added to transfer programs from one white label (state) to another, including their translations, 
- they share the same abbreviated name. 
- The command syntax is as follows:
- python manage.py transfer <target_state> <name_abbreviated> [additional_name_abbreviated...]
For example:
python manage.py transfer ma ssi ctc

In this command: ma denotes the target white label (Massachusetts).
ssi refers to the Supplemental Security Income program.
ctc stands for the Child Tax Credit program.

This script will transfer both the ssi and ctc programs to the ma (Massachusetts) white label, ensuring that all associated translations are included.

### DIFF
--- a/programs/management/commands/transfer.py
+++ b/programs/management/commands/transfer.py
@@ -102,8 +102,6 @@ class Command(BaseCommand):
                 new_program.year = source_program.year
                 new_program.external_name = None
                 new_program.category = None
-                # new_program.documents = None
-                # new_program.required_programs = None
 
                 # Set the new translations
                 for field, translation in translation_mapping.items():
--- a/programs/management/commands/transfer.py
+++ b/programs/management/commands/transfer.py
@@ -4,6 +4,8 @@ from django.db import transaction
 from programs.models import Program
 from screener.models import WhiteLabel
 from configuration.white_labels import white_label_config
+from typing import Dict
+from translations.models import Translation
 
 
 class Command(BaseCommand):
@@ -16,7 +18,7 @@ class Command(BaseCommand):
             help="The white label code to transfer programs to target whilte label",
         )
         parser.add_argument(
-            "external_names",
+            "names_abbreviated",
             nargs="+",
             type=str,
             help="External names of programs to transfer",
@@ -48,7 +50,7 @@ class Command(BaseCommand):
     @transaction.atomic
     def handle(self, *args, **options):
         target_code = options["target_white_label"]
-        external_names = options["external_names"]
+        names_abbreviated = options["names_abbreviated"]
 
         # Validate white label exists
         if not self.validate_white_label(target_code):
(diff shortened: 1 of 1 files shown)


programs/management/commands/transfer.py
```diff
@@ -0,0 +1,90 @@
...
+
+        target_white_label = WhiteLabel.objects.get(code=target_code)
+
+        for external_name in external_names:
+            try:
+                # Get source program
+                source_program = Program.objects.get(external_name=external_name)
+
+                # Check if program already exists
+                new_external_name = f"{target_code}_{external_name.split('_')[-1]}"
```
CalebPena, line 65: We don't have a consistent format for the external names, so lets just leave it empty replace it manually in the admin latter. Also, it will need an index because if there are multiple they need to be unique. Maybe add a message at the end reminding the user to add the external name.
```suggestion
                new_external_name = None
```

programs/management/commands/transfer.py
```diff
@@ -0,0 +1,90 @@
...
+                    continue
+
+                legal_statuses = source_program.legal_status_required.all()
+
+                # Create new program
+                new_program = source_program
+                new_program.pk = None
+                new_program.white_label = target_white_label
+                new_program.external_name = new_external_name
+                new_program.year = source_program.year
```
CalebPena, line 77: Remove the documents, category. and required programs as well. i.e. make those empty. Additionally, please loop through all of the translations, and make copies of those. That way, if you edit the translation of the original program, it will not affect the new program.

programs/management/commands/transfer.py
```diff
@@ -0,0 +1,90 @@
...
+
+                # Save new program
+                new_program.save()
+
+                # Then set the many-to-many relationship
+                new_program.legal_status_required.set(legal_statuses)
+
+            except Program.DoesNotExist:
+                print(f"Error: Program '{external_name}' not found")
```
CalebPena, line 86: See comment above about what to use instead of print.

CalebPena (changes requested):

programs/management/commands/transfer.py
```diff
@@ -0,0 +1,125 @@
...
+
+                # Then set the many-to-many relationship
+
+                new_program.required_programs.set([])
+                new_program.documents.set([])
+                new_program.legal_status_required.set(legal_statuses)
+                self.stdout.write("Reminder: Please add external names to the transferred programs.")
+
+            except Program.DoesNotExist:
+                self.stdout.write(self.style.ERROR(f"Error: Program '{name_abbreviated}' not found"))
```
CalebPena, line 121:
```suggestion
                self.stdout.write(self.style.ERROR(f"Error: Program '{external_name}' not found"))
```

CalebPena (changes requested):

CalebPena (approved):
LGTM

PR #898, opened 2025-04-18
MA: Add MA senior SNAP override and warning

CalebPena:
- Add MA SNAP translation override to link to a new application.
- Add MA SNAP warning message to tell people about the senior application.


myang1010 (approved):
Looks good!

PR #903, opened 2025-04-22
MA: State Supplement Program

CalebPena:
- Add MA State Supplement Program.


programs/programs/ma/pe/member.py
```diff
@@ -111,3 +111,15 @@ def member_value(self, member: HouseholdMember):
...
         return 0
+
+
+class MaStateSupplementProgram(PolicyEngineMembersCalculator):
+    pe_name = "ma_state_supplement"
+    pe_inputs = [
+        dependency.member.AgeDependency,
+        dependency.member.IsDisabledDependency,
+        dependency.member.IsBlindDependency,
+        dependency.member.SsiCountableResourcesDependency,
```
myang1010, line 122: Looks like the `*Ssi.pe_inputs` already includes all these pe inputs. Do we still want them here?

myang1010 (commented):
Looks good to me. Just have one question for the `pe_inputs`.

programs/programs/ma/pe/member.py
CalebPena: Yes, this is okay, because I want to make sure that if the `SSI.pe_inputs` change, we are still making sure to include the inputs we need for `ssp`

kbrey (approved):

PR #904, opened 2025-04-22
MA: CHIP

CalebPena:
- Add CHIP to MassHealth

### DIFF
--- a/programs/programs/federal/pe/member.py
+++ b/programs/programs/federal/pe/member.py
@@ -82,7 +82,7 @@ class Medicaid(PolicyEngineMembersCalculator):
 
 
 class Chip(PolicyEngineMembersCalculator):
-    pe_name = "chip"
+    pe_name = "chip_category"
     pe_inputs = [
         dependency.member.AgeDependency,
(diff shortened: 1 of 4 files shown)


programs/programs/federal/pe/member.py
```diff
@@ -76,11 +73,37 @@ def member_value(self, member: HouseholdMember):
...
+    pe_name = "chip"
+    pe_inputs = [
+        dependency.member.AgeDependency,
+        dependency.member.PregnancyDependency,
+        *Medicaid.pe_inputs,
+    ]
+    pe_outputs = [dependency.member.ChipCategory]
+
+    # NOTE: Monthly
+    chip_values = {
```
kbrey, line 94: Is `chip_values` used? Should this be `chip_categories` and should the line below be ` return self.chip_categories[chip_category] * 12`?

CalebPena: Fixed. This code is not running anywhere yet. I just wanted a place to store the dependencies for CHIP, and also provide a possible implementation for the future.

kbrey (approved):
Cool, looks good.

PR #918, opened 2025-04-28
MA: Urgent needs

CalebPena:
- Add MA specific urgent need calculators. Still waiting on AMI and SMI for some of them.

https://docs.google.com/spreadsheets/d/1j0jM8UkF1L4IsGHy-tGjn7FdW0gAdPm7/edit?gid=85747426#gid=85747426


kbrey (approved):
Don't have access to the spreadsheet but the code looks good.

programs/programs/co/denver_sidewalk_rebate/calculator.py
```diff
@@ -32,7 +17,7 @@ def household_eligible(self, e: Eligibility):
         e.condition(DenverSidewalkRebate.county in counties, messages.location())
 
         # income condition
-        income_limit = DenverSidewalkRebate.income_limits.fetch()[self.screen.household_size - 1]
+        income_limit = ami.get_screen_ami(self.screen, self.ami_percent, self.program.year.period)
```
myang1010, line 20: Just curious, why do we use `.period` here, not `.year`?

myang1010 (approved):
Looks good! Just a question and a couple of minor stuff. Thanks!

programs/programs/co/denver_sidewalk_rebate/calculator.py
CalebPena: We are going to remove the year field, and just make it a text box. It used to be that the year field was the name of the year because we would have "THIS YEAR" and "LAST YEAR". However, now we have them all named the number, it does not matter. That is why we are going to remove that model and make it just a number field.

PR #929, opened 2025-05-01
Added 211co and 211nc to the featureFlages in the co.py and nc.py config files.

SonaliBedge:
-Added 211co and 211nc to the featureFlages in the co.py and nc.py config files.


configuration/white_labels/co.py
CalebPena, line 1: looks like you formatted a little too much. Make sure that you have black running with 120 line length.

CalebPena (changes requested):
Can you update the formatter?

kbrey (approved):

CalebPena (approved):
LGTM

PR #930, opened 2025-05-01
Central: AMI and SMI sheets

CalebPena:
- Switch to centralized AMI and SMI spreadsheets.


kbrey (approved):
Nice, looks good.

Issue #928, opened 2025-04-30
NC: Update NC WIC "Estimated savings" value


Lkp2104:
Context: It is difficult to estimate the cash value/estimated savings of WIC food benefits because it depends on the age of the children in the family and wether a person is pregnant or breastfeeding and whether they breast feed partially or fully. Therefore, Louise reached out to WIC program staff at NCDHHS and they advised we use the the Average Monthly Food Cost Per Person for NC from USDA [data](https://fns-prod.azureedge.us/sites/default/files/resource-files/25wifyavgfd-costs-4.pdf)  $58.63/per person per month. and round up to $60/per month per person across the board rather than the different values for different categories we have now. 
Current estimated values are: 
Children 0–1: $130 
Children 2–5: $26 
Pregnant women: $47 

Need to change estimated savings values in code for NC WIC to be $60/per person per month or all eligible children.

MaxGhenis:
Why did they advise averaging rather than reflecting the age-specific variation in benefit levels?

Linda-Vu:
validated in staging

Linda-Vu:
@MaxGhenis i think Louise is better suited to answer this, but she's OOO for a few days. could you ask her this in the benefits slack and she'll respond to you when she gets back? thank you!

Lkp2104:
> Why did they advise averaging rather than reflecting the age-specific variation in benefit levels?

@MaxGhenis They recommended using the average monthly per person spending according to USDA data to ensure our estimated values are grounded in the WIC program data. They said it is simply impossible to estimate the actual cash value by age-specific variation b/c it depends on so many factors - including whether or not the family redeems all their vouchers and for families with infants whether or not the mother is fully or partially breastfeeding (info the screener does not capture). Happy to discuss more and open to changing if we can find better data. Thanks!

PR #936, opened 2025-05-02
Updated NC WIC estimated saving value

SonaliBedge:
Fixes issue #928 
-changed estimated savings values in code for NC WIC to be $60/per person per month or all eligible children.


kbrey (approved):
Nice work @SonaliBedge, looks good to me.

MaxGhenis:
Out of curiosity could you share the source for this value?

CalebPena (approved):

PR #940, opened 2025-05-05
Central: Add AMI option to use "IL" data

CalebPena:
- Add option to use "IL" data for AMI.

- Updated the programs that use "IL"


myang1010 (approved):
Looks great! At first, I thought `IL` stood for Illinois and that Ben was expanding again... 😅

kbrey (approved):

PR #944, opened 2025-05-06
Central: Add pull screen command

myang1010:
- Add `pull_screen` command to pull a screen from another environment to the current environment, given the uuid.

- I tried removing the API key by setting `permission_classes = []` in `ScreenViewSet` and also setting `"DEFAULT_AUTHENTICATION_CLASSES" = []` in `settings.py`, after removing the API key prompt from the command. But I kept getting a 403 error: "Authentication credentials were not provided." Please let me know if there's anything I missed. Thanks.

### DIFF
--- a/screener/management/commands/pull_screen.py
+++ b/screener/management/commands/pull_screen.py
@@ -1,4 +1,5 @@
 from django.core.management.base import BaseCommand
+from decouple import config
 from getpass import getpass
 
 from django.db import transaction
@@ -26,7 +27,10 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         domain = options["domain"]
         uuid = options["uuid"]
-        api_key = getpass("API key: ")
+
+        api_key = config("DEV_API_KEY", "")
+        if not api_key:
+            api_key = getpass("API key: ")
 
         remote_screen = self._get_screen_from_remote(uuid, domain, api_key)
 
--- a/screener/management/commands/pull_screen.py
+++ b/screener/management/commands/pull_screen.py
@@ -63,7 +63,7 @@ class Command(BaseCommand):
         screen.save()
 
         white_label = remote_screen.get("white_label", screen.white_label)
-        results_url = f"http://localhost:3000/{white_label}/{uuid}/step-3"
+        results_url = f"http://localhost:3000/{white_label}/{uuid}/results/benefits"
 
         self.stdout.write(f"Screen pulled successfully. Results: {results_url}")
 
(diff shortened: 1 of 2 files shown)


screener/management/commands/pull_screen.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        domain = options["domain"]
+        uuid = options["uuid"]
+
+        api_key = config("DEV_API_KEY", "")
+        if not api_key:
+            api_key = getpass("API key: ")
```
CalebPena, line 33: Can you do this for the `pull_validations` script as well. Also, can you add a flag to not use the environment variable, in case we need to pull from staging so that we don't need to change the environment variable?

screener/management/commands/pull_screen.py
```diff
@@ -0,0 +1,69 @@
...
+            serializer = ScreenSerializer(data=remote_screen)
+
+        serializer.is_valid(raise_exception=True)
+
+        screen = serializer.save()
+
+        screen.uuid = uuid
+        screen.save()
+
+        self.stdout.write(f"Successfully pulled screen with UUID {uuid} into the local database.")
```
CalebPena, line 57: Can you add the results url to the message so that we can easily access the results?

CalebPena (approved):
A few minor improvements. After that, it looks good.

screener/management/commands/pull_screen.py
myang1010: Do we also want to add the same flag to `pull_validations` as well?

screener/management/commands/pull_screen.py
myang1010: Should the default results URL be set to "step-3" (http://localhost:3000/{white_label}/{uuid}/step-3)?

CalebPena: Do to `/results/benefits`

screener/management/commands/pull_screen.py
CalebPena: Yes

kbrey (approved):
Thanks @myang1010 ! And would love to document this so devs know to use it 🙌

myang1010:
> Thanks @myang1010 ! And would love to document this so devs know to use it 🙌

Thanks Katie! @kbrey and @CalebPena , what do you think about adding a new page called "Commands" under the Wiki?

CalebPena:
> Thanks Katie! @kbrey and @CalebPena , what do you think about adding a new page called "Commands" under the Wiki?

Lets do that. Can you make an issue to add that.

myang1010:
> > Thanks Katie! @kbrey and @CalebPena , what do you think about adding a new page called "Commands" under the Wiki?
> 
> Lets do that. Can you make an issue to add that.

Done. New issue link, https://github.com/Gary-Community-Ventures/benefits-api/issues/947

PR #949, opened 2025-05-12
CO, Energy: Fix UBP

CalebPena:
- Fix UBP presumptive eligibility.

### DIFF
--- a/programs/programs/co/low_wage_covid_relief/calculator.py
+++ b/programs/programs/co/low_wage_covid_relief/calculator.py
@@ -1,4 +1,4 @@
-from programs.programs.calc import ProgramCalculator, Eligibility, MemberEligibility
+from programs.programs.calc import ProgramCalculator, Eligibility
 from programs.programs.helpers import STATE_MEDICAID_OPTIONS
 import programs.programs.messages as messages
 from programs.co_county_zips import counties_from_screen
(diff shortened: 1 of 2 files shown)


myang1010 (approved):
The changes look good. Should we also apply similar updates to calculators that use the "categorical_eligible" variable, like `co/denver_sidewalk_rebate/calculator.py` and `co/denver_trash_rebate/calculator.py`?

CalebPena:
> Should we also apply similar updates to calculators that use the "categorical_eligible" variable, like `co/denver_sidewalk_rebate/calculator.py` and `co/denver_trash_rebate/calculator.py`?

Those do not have the same problem. In the two that I fixed, we were overriding the `has_benefit` variable to `False`, if the user was not eligible for Medicaid, but was eligible for another program like SNAP.

myang1010:
> > Should we also apply similar updates to calculators that use the "categorical_eligible" variable, like `co/denver_sidewalk_rebate/calculator.py` and `co/denver_trash_rebate/calculator.py`?
> 
> Those do not have the same problem. In the two that I fixed, we were overriding the `has_benefit` variable to `False`, if the user was not eligible for Medicaid, but was eligible for another program like SNAP.

Thanks for the explanation. Makes sense!

PR #951, opened 2025-05-13
Energy: Use urgent need category

CalebPena:
- Add feature flag `dont_show_category_values` to default to not showing the category values.
- Also, I finally got annoyed enough at the `--all` flag that I created a message if you don't select any white labels for the `add_config` command.


myang1010 (approved):
Looks good!

PR #953, opened 2025-05-13
MA: HubSpot

CalebPena:
- Add MA HubSpot integration.

### DIFF
--- a/integrations/services/cms_integration.py
+++ b/integrations/services/cms_integration.py
@@ -60,7 +60,6 @@ class HubSpotIntegration(CmsIntegration):
         self._update_contact(self.user.external_id, data)
 
     def should_add(self):
-        return True  # FIXME: remove
         if settings.DEBUG:
             return False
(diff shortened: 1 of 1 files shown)


integrations/services/cms_integration.py
```diff
@@ -257,7 +257,20 @@ def _hubspot_send_offers_data(self):
         }
 
 
-CMS_INTEGRATIONS = {"brevo": BrevoIntegration, "co_hubspot": CoHubSpotIntegration, "nc_hubspot": NcHubSpotIntegration}
+class NcHubSpotIntegration(CentralHubSpotIntegration):
+    OWNER_ID = "47185138"
```
kbrey, line 261: Wondering if CoHubSpotIntegration should have an OWNER_ID? I don't see one.

CalebPena: That one is still going to the Gary HubSpot because we have not migrated it over yet.

kbrey (approved):

PR #954, opened 2025-05-14
Energy: Old car and category priority

CalebPena:
- Store the old/diesel car question.
- Create a way to prioritize program categories.


myang1010 (approved):
Looks great!

PR #955, opened 2025-05-14
Central: Fix SNAP asset test

CalebPena:
- Remove SNAP asset override.

**We don't know why this was added originally. All of the validations passed though, so it can't be that bad**


SonaliBedge (approved):
I'm reviewing Caleb's PR for the first time. 
I checked that `MeetsSnapAssetTestDependency` is not actually used in the SNAP calculations, so it should be safe to remove the MeetsSnapAssetTestDependency class and any references to it in other files.

PR #962, opened 2025-05-15
Energy: Update renter no expense path

CalebPena:
- Don't show a bunch of programs if a renter does not have expenses.
- Add translation override for renters


programs/programs/translation_overrides/energy.py
```diff
@@ -0,0 +1,10 @@
+from programs.programs.co.energy_calculator.util import has_renter_expenses
+from programs.programs.translation_overrides.base import TranslationOverrideCalculator
+
+
+class CoEnergyCalculatorRenter(TranslationOverrideCalculator):
+    def eligible(self) -> bool:
+        """
+        Show for renters with no expenses
```
kbrey, line 8: This is showing for renters WITH expenses? Potentially this comment is backwards?

kbrey (approved):
minor question but looks good

programs/programs/translation_overrides/energy.py
CalebPena: Good catch. The code is wrong.

PR #966, opened 2025-05-16
Deactivate program when using the transfer script

SonaliBedge:
-Made change in transfer.py file to deactivate program when using the transfer script

### DIFF
--- a/programs/management/commands/transfer.py
+++ b/programs/management/commands/transfer.py
@@ -117,14 +117,10 @@ class Command(BaseCommand):
                 new_program.required_programs.set([])
                 new_program.documents.set([])
                 new_program.legal_status_required.set(legal_statuses)
-                self.stdout.write(
-                    "Reminder: Please add external names to the transferred programs."
-                )
+                self.stdout.write("Reminder: Please add external names to the transferred programs.")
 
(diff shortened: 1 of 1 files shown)


programs/management/commands/transfer.py
```diff
@@ -79,7 +79,7 @@ def handle(self, *args, **options):
                         new_translation = Translation.objects.add_translation(
                             label=f"program.{source_program.external_name}_{source_program.id}-{field}-{str(uuid.uuid4())}",
                             default_message=source_translation.default_message,
-                            active=source_translation.active,
+                            active=False,
```
kbrey, line 82: I'm not sure we want the translation itself to be inactive. Will defer to @CalebPena .

SonaliBedge: @kbrey When I first worked on adding the transfer scripting feature, Caleb mentioned that the program should be set to active = false when transferring it from one white label to another. However, I missed that part while working on other changes.

CalebPena: We only need to change the `new_program.active = False` line. This line would make it so those translations don't get sent to the frontend.

CalebPena (approved):
LGTM

PR #975, opened 2025-05-19
CO, Energy: Use Weatherization Google Sheet

CalebPena:
- Switch back to using the weatherization sheet provided by the Weatherization Assistance Program.

### DIFF
--- a/programs/programs/co/weatherization_assistance/calculator.py
+++ b/programs/programs/co/weatherization_assistance/calculator.py
@@ -1,23 +1,41 @@
-from integrations.services.income_limits import ami, smi
+from integrations.services.sheets import GoogleSheetsCache
+from programs.co_county_zips import counties_from_screen
 from programs.programs.calc import Eligibility, ProgramCalculator
 import programs.programs.messages as messages
 
 
+class IncomeLimitsCache(GoogleSheetsCache):
+    sheet_id = "1ZzQYhULtiP61crj0pbPjhX62L1TnyAisLcr_dQXbbFg"
+    range_name = "A2:K"  # WARN: This selects the first tab because the tab name is "(Updated mm/dd/yyyy)"
+    default = {}
+
+    def update(self):
+        data = super().update()
+
+        return {self._format_county(r[0]): self._format_amounts(r[1:9]) for r in data}
+
+    @staticmethod
+    def _format_county(county: str):
+        return county.strip() + " County"
+
+    @staticmethod
+    def _format_amounts(amounts: list[str]):
+        return [float(a.strip().replace("$", "").replace(",", "")) for a in amounts]
+
+
 class WeatherizationAssistance(ProgramCalculator):
+    income_limits = IncomeLimitsCache()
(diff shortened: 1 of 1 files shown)


programs/programs/co/weatherization_assistance/calculator.py
```diff
@@ -1,23 +1,41 @@
-from integrations.services.income_limits import ami, smi
+from integrations.services.sheets import GoogleSheetsCache
+from programs.co_county_zips import counties_from_screen
 from programs.programs.calc import Eligibility, ProgramCalculator
 import programs.programs.messages as messages
 
 
+class IncomeLimitsCache(GoogleSheetsCache):
+    sheet_id = "1ZzQYhULtiP61crj0pbPjhX62L1TnyAisLcr_dQXbbFg"
+    range_name = "A2:K"  # WARN: This selects the first tab because the tab name is "(Updated mm/dd/yyyy)"
```
jmejia, line 9: No need to change if there are higher priorities but do the sheet or range ever change? Is it possible the sheet could be missing? Should we handle that scenario?

programs/programs/co/weatherization_assistance/calculator.py
```diff
@@ -1,23 +1,41 @@
...
         # income condition
-        fpl_limit = self.program.year.as_dict()[self.screen.household_size] * self.fpl_percent
-        ami_limit = ami.get_screen_ami(self.screen, self.ami_percent, self.program.year.period)
-        smi_limit = smi.get_screen_smi(self.screen, self.program.year.period) * self.smi_percent
-
-        income_limit = max(fpl_limit, ami_limit, smi_limit)
+        counties = counties_from_screen(self.screen)
+        income_limits = []
+        for county in counties:
+            income_limits.append(self.income_limits.fetch()[county][self.screen.household_size - 1])
```
jmejia, line 37: Less of a suggestion and more of a question for my own understanding: How are we guaranteeing that the counties gathered in the Screen don't deviate from the IncomeLimitCache?

jmejia (approved):
Few questions for my own learning but nothing major jumps out to me.

programs/programs/co/weatherization_assistance/calculator.py
CalebPena: The sheet range is unlikely to change because it is a table of income limits by household size and county. If the sheet goes missing we can switch back to the other way to calculate the income limit that this pr is removing. Here is the sheet for reference: https://docs.google.com/spreadsheets/d/1ZzQYhULtiP61crj0pbPjhX62L1TnyAisLcr_dQXbbFg/edit?gid=0#gid=0

programs/programs/co/weatherization_assistance/calculator.py
CalebPena: We are not. We get a Sentry error if one is missing when a user tries to go through under a missing county. We also know that there are 64 counties in CO, and there are 64 counties in the sheet. I have thought of making a script that tries to use every county on a screen and sees if there are any errors, but that has not been an issue so far.

PR #976, opened 2025-05-20
Energy: Colorado Medical Exemption Program

CalebPena:
- Add new member condition option.
- Add Colorado Medical Exemption Program


programs/admin.py
```diff
@@ -49,9 +49,16 @@ def render_change_form(self, request, context, add=False, change=False, form_url
...
-            restricted_query_set = form_field.queryset.filter(
-                Q(white_label=obj.white_label) | Q(id__in=getattr(obj, field).all().values_list("id", flat=True))
-            )
+            if hasattr(obj, field) and hasattr(getattr(obj, field), "all"):
+                restricted_query_set = form_field.queryset.filter(
+                    Q(white_label=obj.white_label) | Q(id__in=getattr(obj, field).all().values_list("id", flat=True))
+                )
+            elif obj.white_label is not None:
+                restricted_query_set = form_field.queryset.filter(
+                    Q(white_label=obj.white_label) | Q(id=getattr(obj, field).id)
```
myang1010, line 58: Could you give me an example of this condition to help me understand the change better?

myang1010 (approved):
The code for the newly added program looks good. I was able to select the dialysis condition from the frontend, but couldn't check the results page due to Google Sheets access limitations. Thanks!

programs/admin.py
CalebPena: This is for the normal selects. Right now it looks like the image bellow, but we want it only show categories for that specific white label.

image

myang1010: Thanks! Like this improvement!

PR #981, opened 2025-05-21
MA: More help

CalebPena:
- Add more help options for MA.


kbrey (approved):
looks good to me

Issue #972, opened 2025-05-19
NC: update values for NC ACA based on new Policy Engine formula


Linda-Vu:
- Context: Policy Engine is working on a new formula to calculate ACA. once that's live, and once Caleb implements the CO version of this, we can replicate that logic for North Carolina.

  - link to Policy Engine ticket: https://github.com/PolicyEngine/policyengine-us/issues/5869
  - link to Caleb's CO ticket: https://github.com/Gary-Community-Ventures/benefits-api/issues/829

- Dev details from Caleb: We are waiting for PE to make the change. I just pinged them to get an update on it. You will just need to inherit from the federal ACA and add the state code dependency.
```
class MaAca(Aca):
    pe_inputs = [
        *Aca.pe_inputs,
        dependency.household.MaStateCodeDependency,
    ]
```

  - You can use these results to test the logic for the new formula:
    - result 1: the estimated monthly savings amount for the ACA program in these pilot [results](https://screener.myfriendben.org/nc/57497b28-ff8d-4173-8e35-c80b629786c0/results/benefits) (UUID: 57497b28-ff8d-4173-8e35-c80b629786c0) was showing as $545/month from our old spreadsheet. By Louise's calculations, the actual estimated monthly savings amount should be $341 once we use the new Policy Engine formula (assuming that the formula is correct).

    - result 2: Screener results were previously not showing ACA results at all for this [result](https://screener.bennc.org/d2ccf3d9-7dd7-4691-91dc-c33cd4a337b0/results). they should be showing $1178/month in estimated savings for ACA with the new Policy Engine formula.

Lkp2104:
@SonaliBedge I think the issue is we need to use the PolicyEngine calculator for ACA (referenced in this [ticket for CO](https://github.com/Gary-Community-Ventures/benefits-api/issues/829)) in the NC ACA program instead of the spreadsheet we are using now. Thank you!

Linda-Vu:
(currently blocked, see ticket description)

Linda-Vu:
tested in prod successfully with these households:

https://screener.myfriendben.org/nc/ed289682-f689-4dd1-b5e6-ab5a6658be19/results/benefits?admin=true

https://screener.myfriendben.org/nc/57497b28-ff8d-4173-8e35-c80b629786c0/results/benefits

@Lkp2104 maybe you can double-check my tests once you get back? thank you!!

Lkp2104:
[...]

PR #982, opened 2025-05-22
Used new Policy Engine formula to update values for ACA program.

SonaliBedge:
-Fix issue #972 
-The ACA program's calculated value is coming from the new Policy Engine formula in "aca_ptc".


programs/programs/nc/__init__.py
CalebPena, line 1: Can you delete the old file as well?

CalebPena (approved):
LGTM. Can you just clean up one thing?

programs/programs/nc/__init__.py
SonaliBedge: @CalebPena, I deleted 'programs/programs/nc/nc_aca/calculator.py' and 'programs/programs/nc/nc_aca/__init__.py' files from nc_aca directory.

Issue #971, opened 2025-05-19
NC: Investigate Incorrect Result for Emergency Medicaid


Linda-Vu:
Context from Louise: when I select the "individuals without lawful presence or U.S. citizenship," these [results](https://screener.myfriendben.org/nc/9dc07e31-c014-4b54-9125-1747b752a2fc/results/benefits) should only show WIC and Emergency Medicaid, but when you select the filter the results page still shows Medicaid but the estimated value changes to the estimated value for Emergency Medicaid. 

Dev details: Figure out why this is happening in prod, talk to Louise for any further clarity, and try to resolve.

Lkp2104:
@Linda-Vu and @kbrey -

After meeting with @SonaliBedge to discuss, we determined the issue is that Emergency Medicaid, which should appear for ALL results where 1) the household is eligible for Medicaid, **and** 2) the household selects the immigration filter "individuals without lawful presence or U.S. citizenship." 

Currently, Emergency medicaid only appears 1) if a member of the household is pregnant, 2) the household is eligible for Medicaid, **and** 3) the household selects the immigration filter "individuals without lawful presence or U.S. citizenship."

Additionally, the "More Info" page for Emergency Medicaid needs the following updates: 
1. Add the NC Navigator Consortium and Medicaid Ambassadors program in the "Get Help Applying" section
2. Add the following copy in the block text section:
"Emergency Medicaid helps pay for care during an emergency or childbirth. It is for people who cannot get regular Medicaid because of their immigration status. You apply the same way as regular Medicaid. You do not need to give your immigration status or Social Security Number. You must apply within 3 months after the emergency or birth."

![Image](https://github.com/user-attachments/assets/aa1a89b5-a7a5-4680-96a1-be79328b9888)
@SonaliBedge Hope this clarifies everything and let me know if you have any more questions.

Lkp2104:
Testing in staging = passed and ready to move into prod (context note: estimated savings for Emergency medicaid is the same for all people getting emergency medicaid - $522/month - and is based of the per person labor and delivery spending reported by NCDHHS) 

**Test household scenarios**:
- Estimated savings for this result is correctly showing as $1022/month - coverage for two people when you select immigration filter "individuals without lawful presence or U.S. Citizenship." 
https://benefits-calculator-staging.herokuapp.com/nc/3876c81c-dbf3-4f06-887e-a9e435a4088d/results/benefits/327/

[...]

PR #984, opened 2025-05-22
Fixed incorrect results for Emergency Medicaid.

SonaliBedge:
fixes issue #971 
-Removed the member.pregnant condition to fix the incorrect results for Emergency Medicaid.


CalebPena (approved):
LGTM

PR #994, opened 2025-05-28
Add Support for Senior Housing Income & Connect for Health Tax Credits

jmejia:
-

Implements BE API support for https://github.com/Gary-Community-Ventures/benefits-calculator/issues/1592

-
-
-


configuration/white_labels/co.py
```diff
@@ -2166,6 +2176,16 @@ def get_white_label(self) -> WhiteLabel:
                     },
                     "description": {"_label": "taxCreditBenefits.fatc_desc", "_default_message": "State tax credit"},
                 },
+                "co_shi_tc": {
```
mexi-cano, line 2179: Small observation. Other tax credit names do not separate the 'tc' part. (coeitc, ctc, coctc, fatc)

screener/models.py
```diff
@@ -375,6 +377,8 @@ def has_benefit(self, name_abbreviated: str):
             "ma_mass_health": self.has_medicaid or self.has_medicaid_hi,
             "co_andso": self.has_co_andso,
             "co_care": self.has_co_care,
+            "co_cfh": self.has_co_cfh,
+            "co_shi_tc": self.has_co_shi_tc,
```
CalebPena, line 381: This key maps to the `name_abbreviated` field on the program. We don't show programs that the user already has by putting a flag on those programs if the `name_abbreviated` key in the map is `True`. It is called it `co_shitc` right now, so we can either change it here, or change the `name_abbreviated` in https://github.com/Gary-Community-Ventures/benefits-api/blob/main/programs/programs/co/__init__.py
```suggestion
            "co_shitc": self.has_co_shi_tc,
```

CalebPena (changes requested):
Looks good. You just need to rename one thing to not show the benefit if the user already has it.

PR #1000, opened 2025-06-03
Central: Secure the admin

CalebPena:
- Make it imposible to edit anything that is not attached to a white label the admin has access to.

- Add a `SecureAdmin` base class for all of the admins.
- Create a secure `TranslationAdminViews` base class for the translations admin.
- Remove some side bars that everyone does not have access to.


CalebPena:
@jmejia Do you think any of these things are blocking this pr from going out before I am back.

> I'm a fan of automated testing in general but this PR in particular I think would benefit from spelling out the expectations of each scenario

I think that this is a good idea. Where do you think that this would be best documented (comments, wiki, something else)? Can you make an issue for this? We could also do the thing where we set up time to talk through it, and you document as we go so that you have all of the context.

> How will we handle migrating existing users?

We already partially restricted people's white label access. This pr completes that, and makes sure that the API is secure. Each user already has the white labels they are allowed to access already.

> Are we tracking when users access admin pages?

Good idea. Can you make an issue.

PR #1004, opened 2025-06-09
Refactor/Urgent need categories

msrezaie:
- Added a new model called `CategoryIconName`. Currently the `icon` field of the `ProgramCategory` model is a `CharField`, requiring a user to manually enter and match the name of the icon (hardcoded in the frontend code) in order for a category to use that icon. This new model is used to store the names of the icons, and the `ProgramCategory` and `UrgentNeedType` (new model for urgent need categories) models' `icon` fields will have a `ForeignKey` relationship to it. This makes it so that the user won't have to manually enter and match the names of the icons.
- Added a new model called `UrgentNeedType`. This model is used to make urgent needs have their own categories (similar to program categories).
- Added a 'Fields Overview' section in `UrgentNeed`'s admin page. This section is used to provide brief descriptions for what each field of the model does.

- `UrgentNeed` model's `type` field is refactored with a new field `category_type`. This new field is used to provide the name, icon, and white label for urgent need categories (since there already exists a `UrgentNeedCategory` model, this model's class and other variables is named `UrgentNeedType`).
- `ProgramCategory` model's `icon` field now uses `CategoryIconName` as a `ForeignKey`.

Please use the following command which will trigger the `data-sync` script.
```
python manage.py data_sync
```

This script will synchronize `ProgramCategory`, `UrgentNeed`, `UrgentNeedType`, and `CategoryNameIcon` models data associations by doing the following:
- will create new urgent need type records and link them to its associated urgent need.
- will link program categories to its associated icons through the new `CategoryIconName` model.

The mapping of theses associations is based off of the 2025_05_30 translations JSON file.


programs/serializers.py
```diff
@@ -41,12 +55,31 @@ class Meta:
...
     class Meta:
         model = UrgentNeed
-        fields = ("id", "name", "website_description", "type")
+        fields = ("id", "name", "website_description")
+
+
+class UrgentNeedTypeSerializer(serializers.ModelSerializer):
+    name = ModelTranslationSerializer()
+    urgent_needs = serializers.SerializerMethodField()
+    icon = serializers.SerializerMethodField()
```
CalebPena, line 74: I think that you can do this because `icon_name` is a property.
```suggestion
    icon = serializers.CharField(source="icon_name")
```

programs/models.py
```diff
@@ -691,12 +715,75 @@ def __str__(self):
...
+        )
+
+        for [field, translation] in translations.items():
+            translation.label = f"urgent_need_type.{icon}_{urgent_need_type.id}-{field}"
+            translation.save()
+
+        return urgent_need_type
+
+
+class UrgentNeedType(models.Model):
```
CalebPena, line 746: I don't have a super strong preference, but I think that `UrgentNeedCategory` might be a little more clear. It would also match `ProgramCategory`.

translations/views.py
```diff
@@ -391,6 +394,81 @@ def navigator_filter_view(request):
...
+
+        paginator = Paginator(urgent_need_types, 50)
+        page_number = request.GET.get("page")
+        page_obj = paginator.get_page(page_number)
+
+        context = {"page_obj": page_obj}
+
+        return render(request, "urgent_need_types/list.html", context)
+
+
```
CalebPena, line 471: I changed how we do this in https://github.com/Gary-Community-Ventures/benefits-api/pull/1000. Can you pull that branch in do this the new way?

screener/serializers.py
```diff
@@ -408,7 +408,6 @@ class UrgentNeedSerializer(serializers.Serializer):
     name = TranslationSerializer()
     description = TranslationSerializer()
     link = TranslationSerializer()
-    type = TranslationSerializer()
```
CalebPena, line 411: I think that you need to change `type` to `category_type` instead of removing this.

programs/management/commands/data_sync.py
```diff
@@ -0,0 +1,213 @@
...
+                icon = CategoryIconName.objects.filter(name__iexact=type_name).first()
+
+                need_type = UrgentNeedType.objects.filter(
+                    white_label=white_label, name__translations__text__iexact=type_name.strip()
+                ).first()
+
+                if not need_type:
+                    need_type = UrgentNeedType.objects.new_urgent_need_type(white_label=white_label.code, icon=icon)
+                    # translate the 'name' field
+                    base_lang = "en-us"
```
CalebPena, line 179:
```suggestion
                    base_lang = settings.LANGUAGE_CODE
```

programs/management/commands/data_sync.py
```diff
@@ -0,0 +1,213 @@
...
+from collections import defaultdict
+from django.apps import apps
+from django.conf import settings
+from tqdm import tqdm
+
+
+class Command(BaseCommand):
+    help = "Urgent need types and program categories data sync"
+
+    URGENT_NEED_TYPE_MAP = {
```
CalebPena, line 12: You should not use the `id`s because the differ between environments. Instead use the `code` for the `WhiteLabel`, and the `external_name` for the `UrgentNeed` and `ProgramCategory`

programs/management/commands/data_sync.py
CalebPena, line 1: You can run custom python code when running a migration (Example [here](https://github.com/Gary-Community-Ventures/benefits-api/blob/main/screener/migrations/0092_auto_20250109_1721.py)). Lets do that instead of having to run a script. Also, it supports creating a function for roll backs if we want to.

@jmejia FYI for your question about rollbacks.

CalebPena (changes requested):

programs/models.py
```diff
@@ -157,6 +157,13 @@ def __str__(self):
         return self.status
 
 
+class CategoryIconName(models.Model):
+    name = models.CharField(max_length=120)
```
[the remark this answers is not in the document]
msrezaie: Yeah I think that would work.

screener/serializers.py
CalebPena: The `UrgentNeedSerializer` maps the structure of the `need_data`. Because you have `category_type` in the `need_data`, you need to add `category_type` to the `UrgentNeedSerializer`.

PR #1009, opened 2025-06-11
Changed health insurance title to "Private (student or non-employer) health insurance"

### LINKED ISSUE #1007: Central: Change copy on health insurance tile that references private, non-employer insurance
Copy edit: Based on Massachusetts user feedback, we should add a reference to student health insurance in the health insurance question.

Modify private insurance tile to read as follows:

"Private (student or non-employer) health insurance"

Image

Solution: determine where we need to make this change to have it persist across the board. if it's a script update, make a PR for that, if it's a simpler django admin update, let Linda know.

SonaliBedge:
Fix #1007 
-Made changes in ma.py config file

### DIFF
--- a/configuration/white_labels/ma.py
+++ b/configuration/white_labels/ma.py
@@ -20,21 +20,12 @@ class MaConfigurationData(ConfigurationData):
     more_help_options = {
         "moreHelpOptions": [
             {
-                "name": {
-                    "_label": "moreHelp.ma.211.name",
-                    "_default_message": "Mass 211",
-                },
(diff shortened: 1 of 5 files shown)


configuration/white_labels/ma.py
kbrey, line 1: @SonaliBedge I think this change should be made in all the white label config files, because the ticket mentions it should be done "all across the board".

kbrey (changes requested):

CalebPena (approved):
LGTM

PR #1015, opened 2025-06-12
Central: Make boxes in django admin expandable

IrynaKolh:
- Central: Make boxes in django admin expandable

https://github.com/user-attachments/assets/e6f35b23-26e6-4e40-91b3-4ed9eb407c03


programs/static/css/style.css
```diff
@@ -0,0 +1,4 @@
+.selector select {
+    resize: vertical;
+    overflow: auto;
```
msrezaie, line 3: This allows for both vertical and horizontal resize:
```suggestion
    resize: both;
    overflow: auto; 
    white-space: nowrap; 
    width: 100%;
    max-width: 100%; 
```

msrezaie (approved):
Just a small suggestion, other than that looks good.

programs/static/css/style.css
```diff
@@ -0,0 +1,4 @@
+.selector select {
+    resize: vertical;
```
CalebPena, line 2: Can we make it resizable in both directions? Some of the long names get cut off and are hard to tell what they are.

programs/admin.py
```diff
@@ -27,6 +27,9 @@
 class WhiteLabelModelAdminMixin(ModelAdmin):
```
CalebPena, line 27: I removed the `WhiteLabelModelAdminMixin` class and replaced it with a `SecureAdmin` class in https://github.com/Gary-Community-Ventures/benefits-api/pull/1000. Can you pull that branch, and make the change there instead?

CalebPena (approved):
Looks good. I appreciate the simple solution.

programs/static/css/style.css
IrynaKolh: yes we can) I will update

programs/admin.py
IrynaKolh: ok, I will do

CalebPena (approved):
LGTM

PR #1020, opened 2025-06-17
Refactor/co new hubspot

CalebPena:
- Use the new HubSpot for CO
- Remove old version
- Remove Brevo integration


mexi-cano (approved):
LGTM

PR #1034, opened 2025-06-25
Central: Update legal services text

CalebPena:
- Change `ID` to `identity documents`


jmejia (approved):
🆔

PR #1054, opened 2025-07-21
Feature/History tracking for Translations

msrezaie:
- Using [django-simple-history](https://django-simple-history.readthedocs.io/en/stable/) package, translation records now have a revision history.
- In the translation admin, every label has its own history table that shows any text changes on any/all of its languages. A history record can be used to revert back a label's current text to its original text.
image

- In order to make use of the django-simple-history package, the unfold package (used for admin dashboard's design layer) had to be upgraded to `0.62.0`. In this version of [unfold](https://pypi.org/project/django-unfold/), the Chinese language (language code: `zh`) is no longer supported as it is divided into `zh-hans` (simplified Chinese) and `zh-hant` (traditional Chinese). Translation label texts saved in `zh` will be moved to `zh-hans` after running `migrate`.
- The `zh` language code of `language_options` in all config files are updated to `zh-hans`

**Important**
- Please update the package version of unfold to `0.62.0` either manually or by running:
```
pip install -r requirements.txt
```
- Please run the following to update the config:
```
python manage.py add_config --all
```


msrezaie:
@jmejia thanks for your feedback,.

A clean up strategy would be a good idea as the translations history can pile up. I also think once the clean up logic is implemented, not saving auto updates wouldn't make much of a difference.

FYI, I've created a [ticket](https://github.com/Gary-Community-Ventures/benefits-api/issues/1070) for this.

kbrey (approved):
Awesome work @msrezaie, looks good to me.

PR #1077, opened 2025-08-05
Bug/Move zh to zh-hans logic to separate migration

msrezaie:
- Moving 'zh' translations to 'zh-hans' has been removed from 0005_historicaltranslation.py and placed into a new migration file, `0006_move_zh_data.py`.

- Run any new migrations with:
```
python manage.py migrate
```


jmejia (approved):
Thanks @msrezaie!

PR #1128, opened 2025-08-26
IL: Fix Moms and Babies Eligibilty (BE)

patwey:
## Context & Motivation

If a household eligible for **FamilyCare** and **Moms and Babies** selects an immigration status filter that excludes **FamilyCare**, we want to show **Moms and Babies**.

Since the immigration filters are applied on the FE, we'll send an API response including eligibility for both programs and a new `excludes_programs` field that the FE will use to ensure the program eligibility is mutually exclusive after the immigration filters (and others) are applied.

- Fixes: https://github.com/MyFriendBen/benefits-api/issues/1085
- Related PR: [link if applicable]

## Changes Made

- Add `excludes_programs` field to `Program`
- Update program model and admin views
- Update `MomsAndBabies` calculator to remove `FamilyCare` eligibility check

## Testing

- Migrations to run: Yes
- Configuration updates needed: See **Deployment** admin updates needed

## Deployment

- Run script: None
- Update production config: None
- Admin updates needed: Add `[Illinois] Moms and Babies` to `FamilyCare` program's `excludes_programs`

## Notes for Reviewers

- Known limitations: We'd be adding a field that's currently only used for one scenario at one white label. This also comes with additional complexity in the front end filtering code.
- Future considerations: The current implementation has likely grown organically based on requirements. Is a higher level pass needed to refactor our filtering?


patwey:
@coderabbitai review

programs/models.py
```diff
@@ -588,6 +600,7 @@ class Program(models.Model):
         on_delete=models.SET_NULL,
     )
     required_programs = models.ManyToManyField("self", related_name="dependent_programs", symmetrical=False, blank=True)
+    excludes_programs = models.ManyToManyField("self", related_name="excluded_by", symmetrical=False, blank=True)
```
jmejia, line 603: My Django-ness is still fresh so I might be wrong on this but would we want this to be `symmetrical=True`?
Thinking through the line above and this excludes example this is how it works in my brain. Let me know if I'm thinking about this incorrectly.

  **required_programs (asymmetric is correct):**
  Advanced_Medicaid → requires → Basic_Medicaid  ✓
  Basic_Medicaid → requires → Advanced_Medicaid  ✗ (wrong direction)

  **excluded_programs (symmetric should be true):**
  FamilyCare → excludes → MomsAndBabies  ✓
  MomsAndBabies → excludes → FamilyCare  ✓ (both directions)

jmejia: What do you think of adding docstrings for these sorts of lines? Maybe it's because I'm new but it took me a bit to think through the scenarios.

kbrey (approved):
Nice! Looks good to me. If there's a way to provide a description for how to use the new field within the django admin UI, that'd be cool.

programs/models.py
patwey: Thank your for catching that, Your Djangoness! `symmetrical=True` is definitely what we want. Will fix!

patwey:
> Nice! Looks good to me. If there's a way to provide a description for how to use the new field within the django admin UI, that'd be cool.

Totally! I learned about `help_text` today. That does what we want!

Screenshot 2025-09-04 at 4 13 00 PM

programs/models.py
kbrey: Just trying to follow this part, would we want to set both in django admin? 

> Admin updates needed: Add [Illinois] Moms and Babies to FamilyCare program's excludes_programs

So would we also want to add FamilyCare to Moms and Babies excludes_programs?

The symmetric behavior is hard for me to wrap my head around, because if they both exclude the other, which is prioritized? Let me know if my Friday brain isn't making sense 😁

programs/models.py
```diff
@@ -588,6 +600,11 @@ class Program(models.Model):
         on_delete=models.SET_NULL,
     )
     required_programs = models.ManyToManyField("self", related_name="dependent_programs", symmetrical=False, blank=True)
+    excludes_programs = models.ManyToManyField(
+        "self",
+        blank=True,
+        help_text="Programs that are mutually exclusive with this program. If Program A excludes Program B, Program B automatically excludes Program A. Example: FamilyCare and MomsAndBabies cannot both be received by the same household.",
```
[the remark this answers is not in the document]
jmejia: @patwey this is great and even better than the docstring!

jmejia (approved):
🍼

Issue #164, opened 2023-11-14
Scope creating new django app for managing regional configuration


sydney-devine:
Move zip codes to the backend
Zip code model (upload json file. . .have frontend pull from json)
Key to a county model?

CalebPena:
Created instructions in https://github.com/Gary-Community-Ventures/benefits-api/issues/178

PR #1161, opened 2025-09-17
#164 Customize Navigator Order

mexi-cano:
- Add django-sortedm2m dependency
  - requirements.txt: add django-sortedm2m==3.1.1
- New migration to introduce ordered navigators per program
  - programs/migrations/0117_program_navigators.py
  - Adds Program.navigators_sorted = SortedManyToManyField(Navigator)
  - Backfills from legacy Navigator.programs membership using a stable WL/id ordering
  - Depends on 0116_update_category_icon_names
- Resolve migration branch conflict on 0117
  - Created 0118_merge_YYYYMMDD_HHMM to merge:
    - 0117_add_excludes_programs
    - 0117_program_navigators

Why
- Maintain deterministic order of navigators per program and avoid ad-hoc sorting in code.
- Unify migration graph after concurrent 0117s landed on main.
- Surface excludes_programs in admin and optimize read paths.
- Cleaned up code

Migrations / deploy notes
- pip install -r requirements.txt
- python manage.py migrate programs

## Context & Motivation

- Fixes Issue #164

## Changes Made

- Ability to now reorder navigators for each program in the django program admin.

Screenshot 2025-09-17 at 10 50 31 AM

## Testing

- pip install -r requirements.txt
- python manage.py migrate programs

## Deployment

- Run script: python manage.py runserver
- Admin updates needed: 1) In the Django admin, go to Programs, select a program (i.e. Medicaid) and scroll down to the 'Navigators sorted' section, the selected navigators are the corresponding ones and can be dragged-and-dropped to reorder. Reorder as you see fit and then save. 2) Complete the frontend screener to receive results, within results page select 'More Info' for any program, the next page lists the new order of the navigators.  

## Notes for Reviewers

- Known limitations and Future considerations: All navigators currently display in the program admin dashboard, in the future it'd be better to have only the corresponding navigators for each program show.


patwey:
@coderabbitai review

@mexi-cano feel free to tag me once you've resolved anything that comes up from CodeRabbit! As a reminder, here's our new PR review flow: https://myfriendbenworkspace.slack.com/archives/C09EFM3JY67/p1758213495505259

screener/views.py
```diff
@@ -433,7 +438,6 @@ def sort_first(program):
                     "documents": [serialized_document(document) for document in program.documents.all()],
                     "warning_messages": warnings,
                     "required_programs": [p.id for p in program.required_programs.all()],
-                    "excludes_programs": [p.id for p in program.excludes_programs.all()],
```
patwey, line 436: Why is this being removed?

templates/admin/programs/change_list.html
```diff
@@ -67,4 +69,4 @@
   document.addEventListener("DOMContentLoaded", initializeDropdowns);
 </script>
 
-{% endblock %}
+{% endblock %}
```
patwey, line 72: [nit] Github likes a newline at the end of the file here

jmejia (commented):
I think I follow what's happening here. Can you confirm whether this is the case?
1. This PR attempts to leverage `django-sortedm2m` which requires a new column to sort by
2. It looks like the migration you have will create a new table rather than add a column to the existing many-to-many table (`programs_navigator_programs`)
3. This means that we will have two tables with the identical data, but one will have the new sort column
4. If these assumptions are true, is it possible to add the column to the existing `programs_navigator_programs` table and leverage `django-sortedm2m` on that table? The [README](https://github.com/jazzband/django-sortedm2m) implies we can use `AlterSortedManyToManyField` to possibly achieve this. Again, new to this library so it's possible you have tried this and it didn't work.
5. I also think an expand/contract approach might be wise on the interface here. Can we keep `programs` and `sorted_programs` simultaneously to make sure we don't break existing functionality? I left a couple of comments that indicate we might not want to remove the existing interface quite yet. Could be good to create a followup PR that only focuses on cleanup once we validate this approach is working as expected.

PR #1442, opened 2026-04-01
MFB-832: Fix translations bulk import failure

nihalerdal:
## Context & Motivation

- Fixes [832](https://linear.app/myfriendben/issue/MFB-832/fix-translations-bulk-import-failure)

Running `bulk_import` from production into staging was failing on the `WarningMessage` step with
`MultipleObjectsReturned` because multiple `trump_account` white-label program variants
all had `external_name=None`. 
## Changes Made

`ProgramManager.new_program()`: removed support for `external_name=None` and introduced a priority-based resolution strategy for `external_name`:

- Uses explicit `external_name` if provided (raises ValueError if already taken)
- Falls back to `name_abbreviated` if it is unique
- Falls back to `<white_label>_<name_abbreviated>` as a final deterministic fallback
- Raises `ValueError` if all candidates are already taken

This ensures every program is assigned a unique `external_name` while supporting multi-white-label program creation.

## Testing

- Migrations to run: None
- Manual testing steps:

1. Create a program with an existing name to verify the `ValueError` is raised.
```
python manage.py shell
from programs.models import Program
try:
    Program.objects.new_program('_default', 'co_snap')
    print("FAIL: ValueError was not raised!")
except ValueError as e:
    print(f"PASS: ValueError raised → {e}")

PASS: ValueError raised → A Program with external_name='co_snap' already exists. 
```

2. Create same program for a different white label to verify the auto-generated external_name
```
try:
    Program.objects.new_program('co', 'co_snap')
    print("PASS: Program created with auto-generated external_name")
except ValueError as e:
    print(f"FAIL: Unexpected ValueError → {e}")

<Program: [Colorado] >
PASS: Program created with auto-generated external_name
```

3. Repeat step 2:
```
FAIL: Unexpected ValueError → A Program with external_name='co_snap_co' already exists.
```
## Deployment

- Post-deployment scripts needed: 
After running the data migration, you must re-export translations from production. The new export will contain the updated `external_name` keys.
```
python manage.py bulk_import
```
- Production config updates:
- Admin updates needed:
Fix null `external_name` values via Django Admin(required before enabling null=False):
 trump_account (IL)
 trump_account (MA)
 trump_account (NC)
 trump_account (TX)
⚠️ Must be applied in production and staging before migration is run.
- Notify team/users of:
Schema change: `Program.external_name` will be enforced as non-null .
Ensure all environments have been updated via admin before deployment to avoid migration failure.
## Notes for Reviewers

- Known limitations:
- Future considerations:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent duplicate external identifiers when creating programs; creation now fails if an identifier is already taken.
  * Validate imported program data to ensure external identifiers are not null and fail early on invalid entries.

* **Chores**
  * Added an administrative command to locate and correct missing external identifiers with dry-run and commit modes; aborts on detected conflicts.


programs/models.py
```diff
@@ -1602,6 +1604,11 @@ def from_model_data(self, data: DataType):
         # get programs
         programs = []
         for external_name in data["programs"]:
+            if external_name is None:
```
kbrey: I know this approach was specified in the ticket, to thrown an error if the external name is None. But I'm curious if we know why the data model allows external_name to be None in the first place, and why we didn't go with a solution to just change it to be required?

kbrey (dismissed):
@nihalerdal This looks good to me! I have one question about approach, and I think either @catonph or @patwey should lay eyes on this since they will be using the command.

programs/management/commands/fix_null_external_names.py
```diff
@@ -0,0 +1,85 @@
...
+                    transaction.set_rollback(True)
+                    self.stdout.write(
+                        self.style.WARNING("Dry run complete. No changes were saved. Use --commit to apply changes.")
+                    )
+                    return
+
+                self.stdout.write(self.style.SUCCESS(f"\nSuccessfully updated {count} programs."))
+
+        except ValueError as e:
+            self.stderr.write(self.style.ERROR(f"Aborting due to error: {e}"))
```
catonph: I'd prefer this not to be a management command! We should be fixing these null values through the admin portal

catonph (changes requested):
Let's avoid a management command here - please just add to the deployment section of the PR the programs that need to be fixed in the admin portal

programs/management/commands/fix_null_external_names.py
nihalerdal: Thanks Caton, I've added program names in the deployment section

nihalerdal: Hi @catonph , I’ve removed the management command as requested. Could you please re-review when you have a chance?

programs/models.py
kbrey: @nihalerdal curious if you have thoughts about this? also @catonph @patwey since you created the ticket, maybe you have context of why?

programs/models.py
```diff
@@ -479,14 +454,16 @@ def new_program(self, white_label: str, name_abbreviated: str):
                 no_auto=(field in self.no_auto_fields),
             )
 
-        # try to set the external_name to the name_abbreviated
-        external_name_exists = self.filter(external_name=name_abbreviated).count() > 0
+        candidate_external_name = external_name or f"{name_abbreviated}_{white_label}"
```
catonph: Let's update this logic to prioritize in this order

1. external_name (if exists and is unique)
2. name_abbreviated (If is unique) 
3. `{white_label}_{name_abbreviated}`

And then throw an error if `{white_label}_{name_abbreviated}` is not unique (the existence check below doesn't make sense since white_label and name_abbreviated are required)

programs/models.py
catonph: Yeah let's go ahead and remove this check here and update the field to be required on the Program model

catonph (changes requested):
A couple other changes needed - thanks!

programs/models.py
```diff
@@ -642,7 +656,7 @@ class Program(models.Model):
         on_delete=models.CASCADE,
     )
     name_abbreviated = models.CharField(max_length=120)
-    external_name = models.CharField(max_length=120, blank=True, null=True, unique=True)
+    external_name = models.CharField(max_length=120, unique=True)
```
catonph, line 659: I think we need a migration for this change?

catonph (changes requested):
Need a migration

programs/models.py
nihalerdal: Thank you Caton, I’ve added the migration!

catonph (approved):
LGTM!

PR #1448, opened 2026-04-07
Migrate referral_options from config files to Referrer model

catonph:
## Context & Motivation

- Closes MFB-864
- Related PRs: MyFriendBen/benefits-calculator#2096, MyFriendBen/data-queries#68

Referral source options were previously stored in white-label config files as a `referral_options` key and read by the frontend via the config endpoint. This made them hard to manage (required a code deploy to add/change options), invisible to the admin portal, and impossible to scope per-WL without duplicating config.

This PR moves referral options into the existing `Referrer` model (which already existed for webhook/navigator config), adds new fields to support the screener dropdown UX, and exposes them via a dedicated API endpoint.

## Changes Made

**Model (`programs/Referrer`)**
- Add `name` (CharField, non-blank enforced via DB `CHECK` constraint) — the display label shown in the dropdown
- Add `show_in_dropdown` (BooleanField, default True) — controls visibility in the screener dropdown
- Add `is_partner` (BooleanField, default False) — distinguishes named partner orgs from generic options (Friend, Google, etc.)
- Add `unique_together` constraint on `(white_label, referrer_code)` — referrer codes are now per-WL

**Migrations**
- `0141`: Adds the new fields and per-WL uniqueness constraint
- `0142`: Seeds `Referrer` rows from existing `referral_options` Configuration data using raw SQL (avoids django-parler incompatibility with historical ORM models); sets `is_partner` based on a known set of generic codes; back-fills blank names; adds DB `CHECK` constraint enforcing non-blank `name`

**New endpoint**: `GET /api/screener-options/{wl}/referral-options/`
- Returns `{ "generic": {code: name, ...}, "partners": {code: name, ...} }`, both groups sorted alphabetically
- Requires authentication (`DjangoModelPermissions`)
- Unknown WL returns empty groups (not 404)

**Admin (`programs/ReferrerAdmin`)**
- Add `name`, `show_in_dropdown`, `is_partner` to `list_display`, `list_filter`, `list_editable`
- Add help text for `referrer_code` (URL param example), `name` (dropdown label), and `is_partner` (grouping explanation)

**README** (`configuration/white_labels/README.md`)
- Document the 6 standard generic referrer codes to add when setting up a new WL

## Testing

**Automated** (`screener/tests/test_referral_sources.py` — 11 tests):
- Response always has `generic` and `partners` keys
- Only `show_in_dropdown=True` referrers are returned
- Hidden referrers (`show_in_dropdown=False`) are excluded
- `is_partner=True` rows appear in `partners`, `is_partner=False` in `generic`
- Results are scoped to the requested WL (cross-WL isolation)
- Correct display names returned for both groups
- Both groups sorted alphabetically by name
- Blank `name` is rejected by DB constraint
- Unknown WL returns `{"generic": {}, "partners": {}}` with 200
- Unauthenticated request returns 403

**Manual testing steps:**
1. Run migrations: `python manage.py migrate`
2. Start the server: `python manage.py runserver`
3. Hit `GET /api/screener-options/co/referral-options/` — verify response has `generic` and `partners` groups, both alphabetically sorted
4. In the admin portal, open a Referrer row — verify help text appears on `referrer_code`, `name`, and `is partner` fields
5. Verify `show_in_dropdown` and `is_partner` are editable inline from the list view

## Deployment

- **Migrations to run:** `0141`, `0142` (seeds data from existing `referral_options` config — idempotent, safe to run on prod)
- **Admin updates needed:** After deploy, go through each WL's Referrer rows in the admin and verify `is_partner` is set correctly. The migration seeds this from the known generic codes list but any WL-specific codes will default to `is_partner=True`.
- **Config files:** `referral_options` Configuration rows are left in the DB as a rollback path but are no longer read by the API or frontend.

## Notes for Reviewers

- Raw SQL in migration 0142 is intentional — django-parler throws `AttributeError: 'NoneType' object has no attribute 'get_all_fields'` when using the ORM on historical `Referrer` models in migrations
- The `referral_options` config entries are NOT deleted — they remain as a safe rollback path
- Follow-up ticket MFB-875 tracks moving `Referrer` out of the `programs` app (it's a screener/WL concept, not a programs concept) and splitting partner config fields into a separate model
- Known limitation: webhook/navigator/program config fields (`webhook_url`, `webhook_functions`, `primary_navigators`, `remove_programs`) still live on the same `Referrer` model — this is pre-existing and tracked in MFB-875


elliottwaissbluth:
Successfully completed manual testing steps 1-5, approved!

elliottwaissbluth (approved):
see comment above

programs/models.py
```diff
@@ -1712,15 +1712,24 @@ class Referrer(models.Model):
...
+    referrer_code = models.CharField(max_length=64)
+    name = models.CharField(max_length=255)
+    show_in_dropdown = models.BooleanField(default=True)
+    is_partner = models.BooleanField(default=False)
     webhook_url = models.CharField(max_length=320, blank=True, null=True)
     webhook_functions = models.ManyToManyField(WebHookFunction, related_name="web_hook", blank=True)
     primary_navigators = models.ManyToManyField(Navigator, related_name="primary_navigators", blank=True)
     remove_programs = models.ManyToManyField(Program, related_name="removed_programs", blank=True)
 
+    class Meta:
```
kbrey, line 1724: I think in the original config solution we could change the dropdown order just by changing the json, in this solution there's no way to adjust the order in the dropdown, right? Maybe we can do something similar to how we did [ordering for program navigators](https://github.com/MyFriendBen/benefits-api/blob/3a7c1926b991020c844bd12315632eb75cc46480/programs/models.py#L1519)?

configuration/white_labels/co.py
```diff
@@ -173,6 +173,7 @@ def get_white_label(self) -> WhiteLabel:
         },
         "ccig": "Colorado Design Insight Group",
         "coAccess": "Colorado Access",
+        "salud": "Salud",
```
kbrey, line 177: Are we planning to remove the referral options from the config in a follow up PR? Maybe worth a deprecated comment?

kbrey (commented):
@catonph some small questions, otherwise looks awesome to me.

programs/models.py
catonph: For now they are alphabetical within their categories. I might let that addition be a future enhancement if the need arises.

configuration/white_labels/co.py
catonph: added!

PR #1453, opened 2026-04-10
MFB-862: Part 1 - Data integrity, gap tracking programs, calculator correctness

catonph:
## Context & Motivation

- Related to MFB-862 (screener current benefits redesign)
- Part 2 (endpoint) depends on this PR

The `show_in_has_benefits_step` field on Program was inconsistently set and not authoritative. Several calculators suppressed their own results when a user declared already having a benefit — this is wrong because the `already_has` flag on the serialized result handles filtering on the frontend, not the calculator. Additionally, CO/MA lacked WL-scoped Program rows for section_8 needed by the `screener_current_benefits` join table, and `cesn_chp`/`cesn_medicaid` were incorrectly modeled as screen-level benefits.

## Changes Made

- **Migration 0140**: Rename `co_andso`/`co_care` under CESN to `cesn_andso`/`cesn_care`
- **Migration 0141**: Create missing WL-scoped Program rows: `co_section_8`, `ma_section_8`, `co_andso`, `co_care`
- **Migration 0142**: Authoritative reset of `show_in_has_benefits_step` — sets `True` on the explicit allowlist, `False` on everything else
    - These are the programs that inform presumed eligibility for other programs
- **Migration 0143**: Delete `cesn_chp` and `cesn_medicaid` Program rows (insurance handled at member level, not screen level)
- **`screener/models.py`**: Add `co_section_8`/`ma_section_8` to `_build_benefit_map()`; remove `cesn_medicaid`
- **`configuration/white_labels/cesn.py`**: Update category_benefits keys to match renamed programs (`co_andso` → `cesn_andso`, `co_care` → `cesn_care`)
- **CO calculators**: Update `section_8` → `co_section_8` in presumptive eligibility (affordable_residential_energy, vehicle_exchange); move `chp` to `member_presumptive_eligibility` in utility_bill_pay
- **IL/MA/TX calculators**: Remove `e.condition(not self.screen.has_benefit(...))` guards — calculators should not suppress their own results based on declared benefits - we rely on `already_has()` for those.

## Testing

- Migrations to run: 0140, 0141, 0142, 0143, 0144 (in order)

### Migration verification
- After 0140: confirm CESN programs previously named `co_andso` and `co_care` are now named `cesn_andso` and `cesn_care` in the admin
- After 0141: confirm new Program rows exist in admin for `co_section_8` (CO), `ma_section_8` (MA), `co_andso` (CO), `co_care` (CO) — all with `active=True`, `has_calculator=False`, `show_in_has_benefits_step=True`
- After 0142: spot-check that programs NOT in `PROGRAMS_TO_FLAG` (e.g. `cesn_chp`) have `show_in_has_benefits_step=False`
- After 0143: confirm `cesn_chp` and `cesn_medicaid` Program rows are deleted

### Calculator testing
Each of these calculators had their `already_has_benefit` guard removed or presumptive eligibility key updated. Verify they still return eligible for a qualifying household and do NOT show if the program is selected in step 8:

**CO — presumptive eligibility key changes:**
- `co/energy_calculator/vehicle_exchange` — no longer active so can't test this update
- `co/utility_bill_pay` — `chp` moved from `presumptive_eligibility` to `member_presumptive_eligibility`; test that a household where a member has CHP insurance is presumed eligible

**IL — `already_has_benefit` guard removed:**
- `il/bap` (Benefit Access Program)
- `il/ccap` (Child Care Assistance)
- `il/commodity_supplemental_food_program`
- `il/pe/tax` (IL EITC and IL CTC)
- `il/transit_reduced_fare`

**MA — `already_has_benefit` guard removed:**
- `ma/cpp` (Cambridge Pre-K Program)
- `ma/dhsp_afterschool`
- `ma/door2door`
- `ma/homebridge`
- `ma/middle_income_rental`
- `ma/taxi_discount`

**TX — `already_has_benefit` guard removed:**
- `tx/pe/member` (Harris County Rides) — `member_value` no longer returns 0 for households that declared having it; verify PE result is used instead

## Deployment

- Admin updates needed: none (migrations handle all data changes)

## Notes for Reviewers

- Migration 0142's `exclude().update(show_in_has_benefits_step=False)` is intentional — makes `PROGRAMS_TO_FLAG` the single source of truth
- `cesn_care` in prod currently has `has_calculator=True` (it's the CO calculator row scoped to CESN); 0140 renames only the tracking-only copy, not the calculator row
- The calculator guard removal is safe: the FE already filters results using the `already_has` flag returned by the serializer
- `co_section_8` and `ma_section_8` are new Program rows only — they will become selectable on the has-benefits step when the DB-driven step ships in a follow-up FE PR


programs/migrations/0141_create_gap_tracking_programs.py
```diff
@@ -0,0 +1,147 @@
...
+        "name_text": "Aid to the Needy Disabled - State Only (AND-SO)",
+        "description_text": "State cash assistance for individuals who are disabled and not yet receiving SSI",
+        "base_program": "andso",
+    },
+    {
+        "white_label_code": "co",
+        "name_abbreviated": "co_care",
+        "name_text": "Colorado's Affordable Residential Energy (CARE) via Energy Outreach Colorado",
+        "description_text": "Home energy upgrades",
+        "base_program": "care",
```
patwey: Do `care` and `andso` (above) need to be added to the `BaseProgram` text choices in `programs/models.py:431`?

patwey (approved):
Looks good!

1 question, don't need to re-review!

programs/migrations/0141_create_gap_tracking_programs.py
catonph: Good catch - these shouldn't really have a base program. Updated the migration files!

PR #1459, opened 2026-04-15
mfb-883: FIX - fixed category types and icons

elliottwaissbluth:
Foregoing description for the sake of time, see changed files.

Verified fixes locally.


patwey (approved):
Looks good!

Any reason you also updated the `active` field in these? I typically do import them as inactive and turn them on one at a time for testing, but that's definitely not required if you'd prefer to import them active as is. Up to you!

PR #1462, opened 2026-04-21
MFB-722: Fixed qualifying relative and added tests

RomannaBidnyk:
## Context & Motivation

- Fixes #[MFB-722](https://linear.app/myfriendben/issue/MFB-722/bug-add-support-for-qualifying-relatives-magi-household-size)

`HouseholdMember.is_dependent()` only implemented the IRS "qualifying child" test. It was missing the "qualifying relative" test.

This caused low-income adults (19+, not students, not disabled) to be incorrectly split into their own tax unit, shrinking household size and pushing families above income thresholds for MAGI-based programs like CHIP, Medicaid, and ACA Premium Tax Credit.

## Changes Made

**Code:**
- `screener/models.py` - added the qualifying relative path to `is_dependent()`, matching the ticket's proposed fix. Threshold lookup replaces the hardcoded value.
- `screener/irs_parameters.py` - new file with a year-keyed dict of IRS thresholds (2022–2026) sourced from PolicyEngine-US YAML with links to the official IRS source for each year.

**Tests added:**
- `screener/tests/test_is_dependent.py` — 16 tests for the `is_dependent()` method itself (both dependency paths + edge cases from the ticket + the MFB-307 scenario).
- `programs/programs/policyengine/tests/test_qualifying_relative_units.py` — 3 tests verifying that low-income adults stay in the main tax unit when we build the PolicyEngine request.
- `programs/programs/tx/pe/tests/test_mfb_307_tx_chip.py` — 2 tests using the MFB-307 Texas family scenario (guards against the bug coming back).
- `programs/programs/co/tests/test_qualifying_relative_impact.py `— 5 tests for CO programs affected by the fix: RTD LiVE and Property Credit Rebate (both flagged in the ticket's impact analysis).

**Tests modified:**
- `screener/tests/test_models.py` — 5 pre-existing tests now fail because of the change. Fix: added income above the threshold to each so they continue testing what they originally tested (not the new qualifying-relative path).

## Testing

- Migrations to run: None
- Manual testing steps:
 The fix works for the bug from the ticket.
   Open the Texas screener, household size 4, ZIP `78701` (Travis County):

   - Parent 1 (age 41): $43,800/yr
   - Parent 2 (age 39): $18,000/yr
   - Child 1 (age 19): $0 income, not a student, no disability
   - Child 2 (age 7): $0 income

   **Expected result:** the 7-year-old shows **CHIP** on the results page. Without this fix, they would not.

## Deployment

- Admin updates needed: None
- Notify team/users of:
  - **CO Property Credit Rebate**: dependents aren't eligible for this rebate. After the fix, low-income adults living at home count as dependents, so they'll be excluded from the rebate going forward.
  - **CO RTD LiVE**: this program checks income against the tax unit. After the fix, adult dependents are part of the parents' tax unit, so their parents' income counts. Adults with high-earning parents may lose eligibility they previously had.

## Notes for Reviewers

**Question — threshold year:** `screen.last_tax_filing_year` is `None` in production. The threshold falls back to the most recent year in the list (**2026, $5,300**). What year should last_tax_filing_year be set to? It should match the year FPL uses in the admin.
UPDATE: Resolved: switched to `screen.get_reference_date().year.` Since `last_tax_filing_year` is never populated, this is now explicit and intentional.

**Note — IRS rules:** The IRS has 4 conditions for "qualifying relative." This PR checks one of them (income below threshold). The skipped one is: "the taxpayer pays for more than half of the person's expenses." We'd need a new wizard question to ask this.

- Known limitations: `last_tax_filing_year` is unused in production; threshold falls back to the most recent year in the list
UPDATE: Threshold now uses` get_reference_date().year `- for non-frozen screens (no validations) this is today's year; for frozen screens it's the earliest validation date's year.

- Future considerations: move threshold to admin or auto-sync with FPL year; bump threshold during annual FPL rollover.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated dependent/qualifying-relative determination to use IRS gross-income thresholds across tax years 2022–2026, improving qualifying-relative eligibility and ensuring correct splitting into the appropriate tax unit.

* **Tests**
  * Added/expanded regression coverage for qualifying-relative unit splitting, qualifying-relative dependency gating, RTD LiVE and Property Credit Rebate scenarios, the Texas TxChip case (MFB-307), and additional qualifying-child/dependent edge cases (including “below,” “at,” and “above” threshold behavior).


programs/programs/tx/pe/tests/test_mfb_307_tx_chip.py
```diff
@@ -0,0 +1,48 @@
...
+        result = pe_input(self.screen, [TxChip])
+        tax_units = result["household"]["tax_units"]
+        people = result["household"]["people"]
+
+        self.assertIn(MAIN_TAX_UNIT, tax_units)
+        main_members = tax_units[MAIN_TAX_UNIT]["members"]
+        self.assertEqual(len(main_members), 4)
+        self.assertIn(str(self.child19.id), main_members)
+        self.assertNotIn(SECONDARY_TAX_UNIT, tax_units)
+        self.assertIn(str(self.child19.id), people)
```
SonaliBedge, line 47: Nit: For amount in all the tests, consider using underscores (40_000) for readability — it's just a visual separator like a thousands comma, Python ignores it at runtime.
It's not important; it's up to you if you want to do it or not.

screener/tests/test_is_dependent.py
```diff
@@ -0,0 +1,138 @@
...
+        child = HouseholdMember.objects.create(screen=self.screen, relationship="child", age=7)
+        IncomeStream.objects.create(
+            screen=self.screen, household_member=self.head, type="wages", amount=40000, frequency="yearly"
+        )
+        self.assertTrue(child.is_dependent())
+
+    def test_qualifying_child_student_under_24(self):
+        student = HouseholdMember.objects.create(screen=self.screen, relationship="child", age=21, student=True)
+        self.assertTrue(student.is_dependent())
+
```
SonaliBedge, line 42: Suggestion: This test passes automatically  — both the student and the head have $0 income, so is_qualifying_relative (0 < 5050) also returns True, meaning the test would still pass even if the qualifying child path was broken. To isolate the qualifying child path, give the student income above the qualifying relative threshold ($5,050) but below half the household income so only qualifying child can make this pass:

```
def test_qualifying_child_student_under_24(self):
    student = HouseholdMember.objects.create(screen=self.screen, relationship="child", age=21, student=True)
    IncomeStream.objects.create(
        screen=self.screen, household_member=self.head, type="wages", amount=40_000, frequency="yearly"
    )
    IncomeStream.objects.create(
        screen=self.screen, household_member=student, type="wages", amount=6_000, frequency="yearly"
    )
    self.assertTrue(student.is_dependent())

```

screener/tests/test_is_dependent.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_qualifying_child_disabled(self):
+        disabled_adult = HouseholdMember.objects.create(screen=self.screen, relationship="child", age=30, disabled=True)
+        self.assertTrue(disabled_adult.is_dependent())
+
```
SonaliBedge, line 52: Same here : 

```
def test_qualifying_child_disabled(self):
    disabled_adult = HouseholdMember.objects.create(screen=self.screen, relationship="child", age=30, disabled=True)
    IncomeStream.objects.create(
        screen=self.screen, household_member=self.head, type="wages", amount=40_000, frequency="yearly"
    )
    IncomeStream.objects.create(
        screen=self.screen, household_member=disabled_adult, type="wages", amount=6_000, frequency="yearly"
    )
    self.assertTrue(disabled_adult.is_dependent())

```

programs/programs/tx/pe/tests/test_mfb_307_tx_chip.py
RomannaBidnyk: Thank you! Updated

screener/tests/test_is_dependent.py
RomannaBidnyk: Thank you for noticing it, @SonaliBedge! Updated

screener/tests/test_is_dependent.py
RomannaBidnyk: Thank you! Updated

SonaliBedge (approved):
Great work, Romanna. Thank you all updates.

screener/irs_parameters.py
```diff
@@ -0,0 +1,34 @@
...
+
+def get_qualifying_relative_threshold(tax_year=None):
+    """
+    Look up the IRS threshold for the given tax year.
+    If tax_year is None or unknown, fall back to the most recent year in the list.
+    """
+    if tax_year and tax_year in IRS_THRESHOLDS:
+        return IRS_THRESHOLDS[tax_year]
+
+    return IRS_THRESHOLDS[max(IRS_THRESHOLDS.keys(), key=int)]
```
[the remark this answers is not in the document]
RomannaBidnyk: updated

screener/tests/test_is_dependent.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertTrue(aunt.is_dependent())
+
+    def test_grandchild_low_income(self):
+        grandchild = HouseholdMember.objects.create(screen=self.screen, relationship="grandChild", age=20)
+        IncomeStream.objects.create(
+            screen=self.screen, household_member=grandchild, type="wages", amount=1_000, frequency="yearly"
+        )
+        self.assertTrue(grandchild.is_dependent())
+
+    def test_disabled_adult_at_threshold_margin(self):
```
myang1010: This test doesn't appear to validate what its name implies. It seems to overlap with the test on line 43.

I think we could either rename the test so that its name accurately reflects what it's actually testing, or update the test data so the disabled adult is not considered a dependent by giving them income that exceeds 50% of the household's total income.

screener/models.py
```diff
@@ -744,13 +745,19 @@ def is_spouse(self) -> bool:
...
+            return False
+
+        # Path 1: Qualifying Child
+        is_qualifying_child = (self.age <= 18 or (self.student and self.age <= 23) or self.has_disability()) and (
+            self.calc_gross_income("yearly", ["all"]) <= self.screen.calc_gross_income("yearly", ["all"]) / 2
         )
 
-        return is_tax_unit_dependent
+        # Path 2: Qualifying Relative
+        threshold = get_qualifying_relative_threshold(self.screen.last_tax_filing_year)
```
myang1010: I'm thinking using `self.screen.get_reference_date().year` might be better here as it returns the earliest validation date for submitted screens, or today for non-frozen screens.

programs/programs/co/tests/test_qualifying_relative_impact.py
```diff
@@ -0,0 +1,149 @@
...
+        self.assertFalse(adult_child.is_in_tax_unit())
+
+        calc = self._make_calculator(screen)
+        e = MemberEligibility(adult_child)
+        calc.member_eligible(e)
+
+        self.assertTrue(e.eligible)
+
+
+class TestPropertyCreditRebateQualifyingRelativeImpact(TestCase):
```
myang1010, line 118: Seems this test is for a different program. Is it intentional to keep it here?

myang1010 (commented):
Hi Romanna, thanks so much for adding support for qualifying relatives. I really like the addition of `irs_parameters.py`. It makes the code much cleaner and easier to follow. I also appreciate the thorough test coverage you added, as well as Sonali's suggestion to include income scenarios that make the tests more specific.

I've left a comment regarding how we determine the "year." Please let me know your thoughts. If we decide to go with that approach, I believe the tests will need to be updated accordingly.

Thanks!

screener/models.py
RomannaBidnyk: Thanks for the suggestion! 

Worth noting that `last_tax_filing_year` is currently never populated by the frontend (always None), so we always fall back to the latest year in the dict anyway. Using `get_reference_date().year` would be more accurate and intentional. 

The trade-off is that it's less semantically correct for the IRS concept — for example, `get_reference_date().year` would return 2026 for a screen submitted today, even if the user filed 2025 taxes in April, where the 2025 threshold ($5,200) would technically be more correct. 
That said, the practical difference is only ~$100 between adjacent years. 

Happy to make the change — just want to confirm before updating the tests accordingly. Should I proceed?

myang1010: Since `last_tax_filing_year` is never populated by the frontend (always None), we already fall back to the latest year in the dict regardless. This means the concern about using the wrong threshold for someone who filed 2025 taxes in April already exists with the current code. `get_reference_date().year` doesn't make it worse. So switching to `get_reference_date().year` is the cleaner choice.

RomannaBidnyk: Thank you, @myang1010! I'll proceed with the implementation.

RomannaBidnyk: Done - switched to `get_reference_date().year` and updated tests to match.

screener/tests/test_is_dependent.py
RomannaBidnyk: Thank you! Renamed to `test_disabled_adult_above_qualifying_relative_threshold_still_dependent` — it verifies that a disabled adult earning above the qualifying relative income threshold is still a dependent via the qualifying child (disability) path.

programs/programs/co/tests/test_qualifying_relative_impact.py
RomannaBidnyk: Yes, this is intentional
Both RTD Live and Property Credit Rebate are CO programs affected by the qualifying relative fix (as per ticket's impact analysis). RTD Live is affected via `is_in_tax_unit()` and Property Credit Rebate via its not `is_dependent()` eligibility gate. Kept them in one file since they both test the same fix's impact on CO programs.

RomannaBidnyk:
Thank you for the thorough review and kind words, @myang1010! I've addressed all the comments - switched to `get_reference_date().year`, updated the tests accordingly, and renamed the overlapping test. Could you take another look when you get a chance?

programs/programs/co/tests/test_qualifying_relative_impact.py
myang1010: That's good reason to keep them together. I'd suggest move these two tests under `/co/tests` at least. If someone is working on Property Credit Rebate, they'd never think to look there.

myang1010 (commented):
Hi Romanna, thanks for the change. Overall looks good to me. Just that combined test file for two different programs. Could we move them to `co/tests/` or separate them? Thanks!

programs/programs/co/tests/test_qualifying_relative_impact.py
RomannaBidnyk: Thanks for pointing that out! Moved both test classes to `programs/programs/co/tests/test_qualifying_relative_impact.py.`

myang1010 (approved):
Thanks Romanna! Looks good to me!

PR #1535, opened 2026-05-21
MFB-564: Make it easier to auto-update years for programs-First Step

myang1010:
## Context & Motivation

This ticket aims to reduce the manual burden of year updates and improve audit-ability of year choices across programs.

This ticket will be the first step to get programs' year updated correctly. 

Leverage hhs Poverty Guidelines API will be in a following ticket.

- Fixes #[MFB-564](https://linear.app/myfriendben/issue/MFB-564/make-it-easier-to-auto-update-years-for-programs)

## Changes Made

- programs/migrations/0171_seed_dynamic_fpl_rows.py, seed "THIS_YEAR_FISCAL" and "THIS_YEAR_CALENDAR" to FederalPoveryLimit
- programs/models.py, add `year_type` to Program with three options, 
    - "hardcoded" (for programs need a specific year in the `year` field)
    - "fiscal_year"
    - "calendar_year" 
- programs/migrations/0172_add_year_type_to_program.py, add `year_type` to Program
- programs/management/commands/set_year_type.py, a command can help update multiple program's year type at once
- programs/admin.py, add `year_type` for admin to choose the year type of the program

## Testing

- Migrations to run: `python manage.py migrate`
- Configuration updates needed: 
- Environment variables/settings to add:
- Manual testing steps:

    - Step 1, Run migrations (seeds dynamic rows + adds year_type field):
      `python manage.py migrate`

    - Step 2, Assign `year_type` to programs using the management command(this also points `year` at the right FPL row automatically):
       For example:
          `python manage.py set_year_type fiscal_year --programs co_snap`
          `python manage.py set_year_type calendar_year --programs co_medicaid`
    
    - Step 3, In Django admin, open the programs you updated in step 2 and verify:
        - year_type shows the value you set
        - year now points to THIS_YEAR_FISCAL or THIS_YEAR_CALENDAR accordingly
        - Hardcoded programs are unchanged

## Deployment

- Post-deployment scripts needed:
- Production config updates:
- Admin updates needed:

    - Step 1, Run migrations (seeds dynamic rows + adds year_type field):
      `python manage.py migrate`

    - Step 2, Assign `year_type` to programs using the management command(this also points `year` at the right FPL row automatically):
       For example:
          `python manage.py set_year_type fiscal_year --programs co_snap`
          `python manage.py set_year_type calendar_year --programs co_medicaid`
    
    - Step 3, In Django admin, open the programs you updated in step 2 and verify:
        - year_type shows the value you set
        - year now points to THIS_YEAR_FISCAL or THIS_YEAR_CALENDAR accordingly
        - Hardcoded programs are unchanged

## Notes for Reviewers

- Known limitations: 

  - FPL fetching and caching will be handled in a separate PR. 

  - `readd_programs.py` still has the legacy `year="THIS YEAR"` lookup, left untouched deliberately since it's unrelated dead/broken code, not something this ticket should fix.

  - We'll need a separate ticket to handle `UrgentNeed` year change

## Summary by CodeRabbit

- **New Features**
  - Programs can now use hardcoded, fiscal-year, or calendar-year poverty limit settings.
  - Added tools to update year types for selected programs or all programs, with dry-run support.
  - Program imports and exports preserve year-type settings.
  - Admin users can edit, view, and filter programs by year type.
  - Dynamic poverty-limit records are seeded and maintained automatically.

- **Bug Fixes**
  - Dynamic year records are matched and synchronized correctly during imports and updates.
  - Snapshot imports no longer overwrite shared dynamic year data.

- **Changes**
  - The previous `fpl_fix` command now directs users to the year-type update tool.


programs/management/commands/set_year_type.py
```diff
@@ -0,0 +1,30 @@
...
+from programs.models import Program
+
+
+class Command(BaseCommand):
+    help = "Assign year_type to programs"
+
+    def add_arguments(self, parser):
+        parser.add_argument("year_type", choices=["calendar_year", "fiscal_year", "hardcoded"])
+        parser.add_argument("--programs", nargs="+", help="External names to target (e.g. co_snap il_medicaid)")
+        parser.add_argument("--all", action="store_true", help="Apply to all programs")
```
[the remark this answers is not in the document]
myang1010: I think the chance is very low. Don't plan to make this change.

programs/migrations/0154_add_year_type_to_program.py
```diff
@@ -0,0 +1,27 @@
...
+    dependencies = [
+        ("programs", "0153_seed_dynamic_fpl_rows"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="program",
+            name="year_type",
+            field=models.CharField(
+                blank=True,
```
SonaliBedge: Do we need `blank=True` here? It will allow an empty string. Do you want to allow blank values in addition to the choices below? Also, in `models.py` lines 693–702, `blank=Tru`e is omitted.

programs/management/commands/set_year_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def handle(self, *args, **options):
+        year_type = options["year_type"]
```
SonaliBedge, line 22: Nitpick: Here as well, consider adding type annotations on line 13 for better readability and maintainability. For example:
`def handle(self, *args: Any, **options: Any) -> None:`

programs/management/commands/set_year_type.py
```diff
@@ -0,0 +1,30 @@
+from django.core.management.base import BaseCommand
+from programs.models import Program
+
+
+class Command(BaseCommand):
+    help = "Assign year_type to programs"
+
+    def add_arguments(self, parser):
+        parser.add_argument("year_type", choices=["calendar_year", "fiscal_year", "hardcoded"])
```
SonaliBedge, line 11: Nick pick: Consider adding type annotations on line 8 for better readability and to help catch bugs early. For example:

`def add_arguments(self, parser: ArgumentParser) -> None:`

programs/management/commands/set_year_type.py
```diff
@@ -0,0 +1,30 @@
...
+            programs = Program.objects.filter(external_name__in=requested)
+            found = set(programs.values_list("external_name", flat=True))
+            missing = requested - found
+            if missing:
+                self.stdout.write(self.style.WARNING(f"Not found: {', '.join(missing)}"))
+        else:
+            self.stdout.write(self.style.ERROR("Provide --all or --programs <external_name ...>"))
+            return
+
+        updated = programs.update(year_type=year_type)
```
SonaliBedge: On line 29, it will update only year_type; it will not modify year.
For initial deployment, this is fine because migration 0155 runs afterwards and sets year based on year_type. But migration 0155 only runs once. After that, every future call to set_year_type hits line 29 and sets year_type without updating year. 
Example:
```
Before:  year_type="hardcoded",   year → FPL(year="2024")
Command: set_year_type calendar_year --programs co_snap
After:   year_type="calendar_year", year → FPL(year="2024")  ← still wrong

```

The program now says it's a calendar-year program but its eligibility calculator is still reading 2024 FPL data instead of THIS_YEAR_CALENDAR.

The fix replaces line 29 with logic that updates both fields together:

# replace this:
`updated = programs.update(year_type=year_type)`

# with this:
```
from programs.models import FederalPoveryLimit

update_kwargs = {"year_type": year_type}
if year_type == "calendar_year":
    update_kwargs["year"] = FederalPoveryLimit.objects.get(year="THIS_YEAR_CALENDAR")
elif year_type == "fiscal_year":
    update_kwargs["year"] = FederalPoveryLimit.objects.get(year="THIS_YEAR_FISCAL")
# hardcoded: leave year unchanged — admin sets it manually

updated = programs.update(**update_kwargs)
```

This makes the command safe to use at any point in the future, not just during initial migration setup.

SonaliBedge (commented):
Thank you for working on this ticket, Ming. I’ve added a few comments — please let me know your thoughts on them.
Also, please consider adding a test file for this change as well. Other management commands in this app already have test coverage under `programs/management/commands/tests/`.

programs/management/commands/set_year_type.py
myang1010: My plan is to update year in a separate step, so I prefer keeping this command single concern. These management commands are one-time migration utilities for backward compatibility, making it easier for admins to update existing programs. They're not intended for ongoing use. 
New programs will have `year_type` and `year` set at creation time.

programs/migrations/0154_add_year_type_to_program.py
myang1010: Removed!

myang1010:
@SonaliBedge , thanks for the review. 

Re: `Other management commands in this app already have test coverage under programs/management/commands/tests/.`, I didn't add test for this command because it's intended to be a one-time-use helper script. Given its short lifespan, I don't think test coverage would provide much value here.

PR #1559, opened 2026-06-02
MFB-981: Heat Pump Step 3 - Wire Rewiring America API + EPA equivalencies for heat pump water heater

david-j-stoll:
## Context & Motivation

- Related to Linear ticket MFB-981 (Heat Pump Step 3: Wire Rewiring America API + EPA equivalencies for heat pump water heater)
- The CESN Calculate Impact page (built in MFB-979) had a no-op form submit handler. This PR wires the backend half of that integration — proxying the Rewiring America Residential Electrification Model (REM) API so the frontend can request bill and emissions impact data without holding an API key client-side.

## Changes Made

- **`integrations/clients/rewiring_america.py`** (new): `RewiringAmericaClient` class with a `fetch_rem_impact()` method that calls `GET https://api.rewiringamerica.org/api/v1/rem/address` with Bearer auth and returns the raw response dict.
- **`benefits/settings.py`**: Added `REWIRING_AMERICA_API_KEY = config("REWIRING_AMERICA_API_KEY", default="")` to pull the key from the environment.
- **`screener/serializers.py`**: Added `RemImpactSerializer` and a helper `_convert_emissions_to_lbs()`. The serializer strips the full RA API response down to just `bill_delta` (cost delta, in $) and `emissions_delta` (emissions delta, converted from kgCO2e → lbCO2e). Only `fuel_results.total.delta` is returned — per-fuel breakdowns, rates, and emissions factors are dropped.
- **`screener/views.py`**: Added `RemImpactView` (`AllowAny`, GET) which validates required query params (`upgrade`, `address`, `heating_fuel`), calls `RewiringAmericaClient`, serializes the result, and returns a 502 with the upstream error body on RA API failure.
- **`screener/urls.py`**: Registered `GET /api/screener-options/<white_label>/rem-impact/` pointing to `RemImpactView`.
- **`screener/tests/test_co_cesn_utils.py`** (new): 10 unit tests covering `_convert_emissions_to_lbs` (value conversion, unit rename, all stat keys, zero, empty input, no mutation) and `RemImpactSerializer` (bill passthrough, emissions conversion, field stripping, missing input handling).

## Testing

- Migrations to run: None
- Environment variables/settings to add: `REWIRING_AMERICA_API_KEY` in `.env` (key already present in local `.env` files; confirm it is set in Staging/Production environments)
- Manual testing steps:
  1. Start the local server: `python manage.py runserver`
  2. Run the following curl against a valid Colorado address:
     ```
     curl "http://localhost:8000/api/screener-options/cesn/rem-impact/?upgrade=water_heater__heat_pump_uef3.35&address=200+E+Colfax+Ave+Denver%2C+CO+80203&heating_fuel=natural_gas&water_heater_fuel=natural_gas"
     ```
  3. Verify the response contains `bill_delta` (values in `$`) and `emissions_delta` (values in `lbCO2e`)
  4. Run the unit tests: `pytest screener/tests/test_co_cesn_utils.py -v`

## Deployment

- Production config updates: Confirm `REWIRING_AMERICA_API_KEY` is set in the Staging and Production environment configs
- Admin updates needed: None
- Notify team/users of: None — the endpoint is only called from the CESN Calculate Impact page, which remains gated behind the `cesn_heat_pump_journey` feature flag (still `false` in production)

## Notes for Reviewers

- Known limitations: The `RemImpactSerializer` uses `DictField()` for the nested stat objects (`bill_delta`, `emissions_delta`) since the exact inner field names (`mean`, `median`, `percentile_20`, `percentile_80`) are not validated at the serializer level — they are passed through as-is from the RA API. This keeps the serializer resilient to minor RA API changes but means the frontend is responsible for safely accessing those nested fields.
- Future considerations: Steps 4 (remaining upgrade types) and 5 (rendering the journey section on the results page) will build on this endpoint. No backend changes are anticipated for those steps.

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint providing REM (Rebate Eligibility and Modeling) impact calculations that estimate bill savings and emissions reductions.
  * Implemented rate limiting on the REM impact endpoint (30 requests per hour).
  * Emissions reduction metrics are now calculated and returned in imperial units.


screener/tests/test_co_cesn_utils.py
```diff
@@ -0,0 +1,107 @@
+"""
```
catonph: This file name is a little misleading and inconsistent with our others. We dont have a utils module for CESN. I would stick to something like `test_rem_impact.py`

catonph: also we should steer clear of "co" which is it's own WL

screener/serializers.py
```diff
@@ -569,3 +569,35 @@ def get_nps_score(self, uuid: str) -> NPSScore:
...
             raise serializers.ValidationError({"uuid": "No NPS score found for this session"})
+
+
+_KG_TO_LBS = 2.20462
+
+
+def _convert_emissions_to_lbs(emissions: dict) -> dict:
+    """Convert each stat in an emissions dict from kgCO2e to lbCO2e in-place."""
+    return {key: {"value": stat["value"] * _KG_TO_LBS, "unit": "lbCO2e"} for key, stat in emissions.items()}
+
```
catonph, line 580: This will KeyError on malformed upstream data. This assumes every stat has a "value" key. The PR's own "Known limitations" note says inner field names aren't validated and are "passed through as-is from the RA API" — but this line doesn't pass through, it indexes. If RA ever returns an emissions stat without value (or value: null), this raises KeyError/TypeError inside to_representation, which surfaces as an unhandled 500 — not caught by the view's try/except (that only wraps the client call, not serialization). Consider stat.get("value", 0) or guarding, consistent with the defensive .get() chaining you already use for fuel_results/total/delta.

screener/tests/test_co_cesn_utils.py
```diff
@@ -0,0 +1,107 @@
...
+Unit tests for CESN Calculate Impact utilities: the kgCO2e → lbCO2e conversion
+function and the RemImpactSerializer that applies it.
+"""
+
+from django.test import TestCase
+from screener.serializers import RemImpactSerializer, _convert_emissions_to_lbs
+
+_KG_TO_LBS = 2.20462
```
catonph: Consider importing this instead of duplicating it here? (it's also in screener/serializers.py)

screener/views.py
```diff
@@ -746,3 +766,52 @@ def get(self, request, white_label):
...
+    """
+    Proxies a request to the Rewiring America REM /api/v1/rem/address endpoint
+    and returns only the total cost delta (bill impact) and total emissions delta
+    that the frontend needs to render the Calculate Impact results view.
+    """
+
+    permission_classes = [permissions.AllowAny]
+    throttle_classes = [RemRateThrottle]
+
+    def get(self, request, white_label: str) -> Response:
```
catonph: Do we need white_label here?

screener/serializers.py
```diff
@@ -569,3 +569,35 @@ def get_nps_score(self, uuid: str) -> NPSScore:
...
+
+class RemImpactSerializer(serializers.Serializer):
```
catonph, line 582: RemImpactView has no test coverage — the new tests only cover the serializer + converter, but the view holds the branching logic (400 on missing params, the 502 paths, happy-path wiring). Since the view calls an external API, mock the client (we already use unittest.mock.patch in test_nps.py), e.g. 

```
@patch("screener.views.RewiringAmericaClient.fetch_rem_impact")
def test_upstream_http_error_returns_502(self, mock_fetch):
    err = requests.HTTPError()
    err.response = Mock(status_code=400, json=lambda: {"msg": "bad"})
    mock_fetch.side_effect = err
    response = self.client.get(self.url, self.params)
    self.assertEqual(response.status_code, status.HTTP_502_BAD_GATEWAY)
```

We should probably at least cover the 400 (missing params), one 502 branch, and the happy path.

catonph (changes requested):
Looks good but a few requests!

catonph (approved):
LGTM!

PR #1560, opened 2026-06-03
MFB-1112: PolicyEngine version config + gated ?pe_version override

catonph:
## Context & Motivation

Part of the PolicyEngine frontier migration ([MFB-1100](https://linear.app/myfriendben/issue/MFB-1100/migrate-to-policyengine-frontier-api)). This is the plumbing sub-issue [MFB-1112](https://linear.app/myfriendben/issue/MFB-1112/pe-frontier-global-policyengine-version-config-gated-url-override).

The PolicyEngine model version was about to be hardcoded into the request body. That couples the frontier cutover to a deploy — the moment the code hits prod, every request is on the new version, with no validation window and no instant rollback (risky for a change touching SSI/Medicaid dollar amounts). This makes the version **runtime config** instead, so the cutover is a reversible config flip rather than a deploy.

- Targets the integration branch `caton/mfb-1100-frontier-integration`, not `main`. The SSI sub-issue (MFB-1102, #1561) stacks on top of this.

## Changes Made

- **`PolicyEngineConfig` singleton** (`configuration` app) — global single-row model holding the pinned version sent as the top-level `"version"` field on `/us/calculate`. Pinned `MAJOR.MINOR.PATCH` strings only (or blank); `clean()` validates + strips, blank ⇒ omit the field (PolicyEngine's default). Superuser-only via `SecureAdmin`, with singleton UX (no add once it exists, no delete). The hot path reads via the read-only `current_version()` classmethod so a calc never writes a config row.
- **Admin change form** — a readonly "Current version" line (`active_version_display`, shows the active value or `(default — PolicyEngine current)`) above the editable field. Reset to default = clear the field and Save.
- **Shared `versions.py` module** (`programs/programs/policyengine/versions.py`) — single source of truth for what a valid PE version is and what each form means: `VERSION_RE`, the `current`/`frontier` alias constants, the `NEWEST_VERSION` sentinel, and the helpers `is_valid_version_number`, `is_valid_override`, `to_comparable_pe_version`, `determine_pe_version`, `version_supports`. The config model, the `?pe_version=` override validator, and the resolver/gating all import from here (no duplicated regex/alias lists).
- **Version resolution + injection** — `determine_pe_version()` resolves the version (per-request override → config pin → omit); `pe_input()` injects it as the top-level `"version"`. Flows to both `ApiSim` and `PrivateApiSim`.
- **Gated `?pe_version=` override** — threaded view → `all_results` → `eligibility_results` → `calc_pe_eligibility`. The view flips `screen.is_test=True` **before** the calc runs, so the resulting `EligibilitySnapshot` is excluded from exports/marts. Validated via `pe_versions.is_valid_override` (alias or version-number, else **400**). The override is the only place the floating aliases may be used; the config rejects them.
- **Webhook gating** — the referrer webhook is skipped for **any** test screen (`if not screen.is_test`), covering `?pe_version=` previews and any other test screen. ⚠️ Slightly broader than the override — see reviewer notes.
- **Engine error visibility** — `engines.py` now includes PolicyEngine's response body in the failure message (`raise_for_status()` omits it), so a failed call (e.g. a well-formed-but-nonexistent version) is diagnosable in Sentry instead of a bare "400".
- **Tests** — singleton one-row enforcement, version-format validation (rejects aliases + junk like `xcZX`, strips whitespace), read-only accessor doesn't write, admin display + `obj=None` add-form guard, override validation + 400 path, the `is_test`-before-calc ordering invariant, webhook-not-sent-for-test-screen, and direct coverage of the `versions.py` helpers.

## Testing

- Migrations to run: `python manage.py migrate configuration` (adds `0006_policyengineconfig`; metadata folded in, no separate migration).
- Configuration updates needed: none to merge. To pin a version: `PolicyEngine Version → policyengine_version` in the Django admin (e.g. `1.715.2`); clear + Save to return to default.
- Manual testing steps:
  - No row / blank version → request body has no `version` key (unchanged behavior).
  - Set `1.715.2` in admin → request body includes `"version": "1.715.2"`.
  - Save `frontier`/`current`/junk (`xcZX`) in admin → rejected with a validation error.
  - `/api/eligibility/<uuid>?pe_version=frontier` (authenticated) → screen flips to `is_test=True`, request uses the override, no webhook fired.
  - `/api/eligibility/<uuid>?pe_version=fronteir` (typo) → `400` `{"pe_version": "must be a version number like '1.715.2', or 'current'/'frontier'"}`, no snapshot, `is_test` untouched.
  - `python manage.py test configuration programs.programs.policyengine screener.tests` — all green.

## Deployment

- Post-deployment scripts needed: none.
- Production config updates: none on merge. **The cutover itself is a later, deliberate admin action** — defaults to unset, so merging/deploying changes nothing until flipped.
- Admin updates needed: none (model self-registers; perform the cutover as a superuser).
- Notify team/users of: nothing on merge.

### Frontier cutover checklist (whole migration, not just this PR)

This config is the cutover mechanism for the migration. When flipping to frontier:

1. **The pin is mandatory and must land before June 10** — it is **not** "or wait for the default flip." Frontier-gated inputs (e.g. `meets_ssi_disability_criteria` from MFB-1102) are omitted while the version is unset, and stay omitted even after PE flips its server-side default on June 10. So an unpinned prod on/after June 10 runs frontier *without* those inputs → disabled members' SSI drops to 0. Setting `policyengine_version = 1.715.2` is what actually activates the migration.
2. **Pre-flip:** confirm MFB-1102 (and other frontier sub-issues) are merged to prod first, version still unset (code dormant, no behavior change).
3. **Flip:** set `policyengine_version = 1.715.2` in the Django admin (superuser).
4. **Verify immediately after flip:** run a disabled, non-aged/non-blind test screen and confirm `ssi > 0` (and SSI-recipient Medicaid amounts in MA/CO/NC/IL/WA). ⚠️ A bad pin (valid format but nonexistent version) fails *safe but silent* — all PE programs return empty, logged to Sentry only (now with the PE reason). This verify step is what catches it before users do.
5. **Rollback if wrong:** clear the field (or revert to a prior pinned version) and save — instant, no deploy.

## Notes for Reviewers

- **Webhook gating scope:** skipping the webhook for test screens is intentionally broader than `?pe_version=` — it also suppresses the webhook for any pre-existing test screen (e.g. a `referrer_code=test` screen with a configured webhook). Desired (a test screen should never POST to a partner), but flagging since it touches the non-override path.
- **`?pe_version=` override is intentionally open to any caller** (not staff-gated): it only affects the requester's own screen, flips it to `is_test`, and never persists to other users / marts / config. The *consequential* control — the prod version pin — is superuser-only via `SecureAdmin`.
- `PolicyEngineConfig.save()` forces `pk=1` and runs `full_clean(exclude=["id"])` so a second save is an intentional singleton upsert; the read-only `current_version()` is used on the hot path to avoid writing a row during a calc.
- Future considerations: a replay-based version-diff canary is noted on MFB-1100 as a possible future backstop; not in scope here.


configuration/admin.py
```diff
@@ -33,3 +33,43 @@ def get_form(self, request, obj=None, **kwargs):
...
+
+
+class PolicyEngineConfigAdmin(SecureAdmin):
+    """Global singleton. SecureAdmin already restricts a white-label-less model to
+    superusers; we additionally enforce single-row UX (no add once it exists, no delete).
+
+    The change form shows a readonly "Current version" line above the editable field;
+    to reset to PolicyEngine's default, clear the field and save (per its help_text)."""
+
+    DEFAULT_LABEL = "(default — PolicyEngine current)"
```
david-j-stoll, line 45: Reading this as-is, it seems like this `DEFAULT_LABEL` value is meant to be the fallback for the version number. Currently, this is just a string and not a valid version number (see function at line 58 as well).  If we want to use `DEFAULT_LABEL` as a fallback for the version number, we should explicitly set this. Based on the chat thread with Ziming, this can be either an enumerated string ("frontier" or "current") or a valid version number.

configuration/models.py
```diff
@@ -13,3 +16,84 @@ class Configuration(models.Model):
...
+
+        # Blank is allowed (omits the "version" field); otherwise must be an exact
+        # version number. This also rejects the floating "frontier"/"current" aliases,
+        # which simply aren't valid here.
+        if value and not self.VERSION_RE.match(value):
+            raise ValidationError(
+                {"policyengine_version": 'Must be an exact version number like "1.715.2", or left blank.'}
+            )
+
+    def save(self, *args, **kwargs):
```
david-j-stoll, line 71: See comment above. Is there any value in storing previous version numbers to use as a fallback? Maybe not, but if so then we need to be clear that the version number has no default.

configuration/models.py
```diff
@@ -13,3 +16,84 @@ class Configuration(models.Model):
...
+    them when it promotes a release, so storing one would let our served version change
+    under us. The per-request ?pe_version= override (a test-only preview) is the only
+    place those aliases may be used. Blank means omit the "version" field entirely, i.e.
+    PolicyEngine's default.
+    """
+
+    # An exact package version, e.g. "1.715.2". The shape (not a hardcoded list) means
+    # new PolicyEngine releases need no code change. Shared with the ?pe_version=
+    # override validation in screener.views.
+    VERSION_RE = re.compile(r"^\d+\.\d+\.\d+$")
```
david-j-stoll: If self.policyengine_version can be an enumerated string (such as `frontier` or `current`) this wouldn't work.

david-j-stoll (changes requested):
Looks great, just a couple things to sanity check

configuration/models.py
catonph: I designed this so only numeric versions can be set from the django admin and stored in the DB. `frontier` and `current` are only meant to be used as url parms on test screeners and can never be injected into the DB itself. Since if someone forgets they set e.g. `frontier` here it could break in prod.

configuration/models.py
catonph: The "default" / fallback version number will be to send no version number and then PE just uses their current version. So this is just a singleton override. I thought about whether we want this as a bit of a log…..open to arguments for it

configuration/admin.py
catonph: So this is just the default *label* within the Django admin portal - it doesnt relate to anything that is sent in the code

screener/views.py
```diff
@@ -150,14 +171,30 @@ def get(self, request, id):
         ).get(uuid=id)
 
         is_admin = request.query_params.get("admin")
-        results = all_results(screen, is_admin=is_admin)
+
+        pe_version = request.query_params.get("pe_version")
+        if pe_version:
```
david-j-stoll, line 165: Last thing, is it worth it to verify whether or not the user attempting to set the pe_version is a staff member or superuser? Looks like that is something that we are tracking (looking at `benefits-api\authentication\models.py:8`. May not be a big concern in the short term but could cause problems down the line.

configuration/admin.py
catonph: (current_version() is also just a string label generator)

screener/views.py
catonph: Good call out:
- Override (?pe_version=) — intentionally open to anyone; it's a self-scoped, non-persisting, test-screen-only preview. No identity gate needed.
- Prod admin version pin through the admin portal — already set to superuser-only

david-j-stoll (approved):

PR #1561, opened 2026-06-03
MFB-1102: send meets_ssi_disability_criteria for SSI (frontier)

catonph:
## Context & Motivation

Part of the PolicyEngine frontier migration ([MFB-1100](https://linear.app/myfriendben/issue/MFB-1100/migrate-to-policyengine-frontier-api)) — sub-issue [MFB-1102](https://linear.app/myfriendben/issue/MFB-1102/pe-frontier-send-meets-ssi-disability-criteria-for-ssi-ssi-recipient). The one confirmed code break in the migration.

On the frontier model (policyengine-us 1.715.2, promoted to default June 10), SSI disability classification **no longer falls back** to `is_disabled` / reported SSI receipt. Without sending `meets_ssi_disability_criteria`, a disabled non-aged/non-blind person gets `is_ssi_disabled: false` → `ssi: 0`, with downstream effects on SSI-recipient Medicaid (non-zero in MA/CO/NC/IL/WA), SNAP categorical eligibility, and SSI-linked state supplements.

This does not fix a current bug — current prod (1.691.1) still works. It **prevents disabled members' SSI (and downstream programs) from dropping to zero once we move to frontier** (which we control via the version pin — see Deployment).

- Targets the integration branch `caton/mfb-1100-frontier-integration` (builds on MFB-1112's version config), not `main`.

## Changes Made

- **`MeetsSsiDisabilityCriteriaDependency`** (`meets_ssi_disability_criteria`, person-level), added to the `Ssi` calculator's `pe_inputs`. Value mirrors `IsDisabledDependency` (`disabled OR long_term_disability OR visually_impaired`).
- **Scope is the `Ssi` calculator only** — verified from policyengine-us source that the *only* non-test consumer of this variable (for our states) is `is_ssi_disabled`; the Medicaid `SSI_RECIPIENT` and SNAP categorical paths are downstream of `ssi`/`is_ssi_eligible` and do not read it independently, so no Medicaid/SNAP/state-supplement calculator changes are needed.
- **Including blindness is verified safe**: SSI eligibility is `is_ssi_aged OR is_blind OR is_ssi_disabled` (PE source), so the flag can only add to an OR, never reduce eligibility.
- **Version gating** — `meets_ssi_disability_criteria` does **not exist in the current model (1.691.1)**; sending it 400s the whole `/calculate` request. Added a reusable `min_pe_version` marker on the dependency base class: `pe_input()` resolves the request version once and omits any dependency whose `min_pe_version` exceeds it. This makes the change **safe to ship dormant** — the input is omitted unless the version is explicitly pinned to ≥ 1.715.2 (or `frontier`). See Deployment for the activation caveat.
- Dependency unit tests + version-gating tests (no static validation scenario — the frontier *outcome* check is homed in MFB-1103; see Testing).

## Testing

Migrations to run: none.

Coverage is intentionally split across three layers — a static validation in standard CI can't honestly test the frontier *outcome* (CI runs against the current model 1.691.1, where the old `is_disabled` fallback masks the change), so the behavioral check is homed in MFB-1103 instead.

**Layer 1 — automated, in this PR (CI on every push):**
- Dependency tests: `MeetsSsiDisabilityCriteriaDependency.value()` returns the correct boolean (disabled / long-term / visually-impaired / none) and emits field `meets_ssi_disability_criteria`.
- Version-window gating tests (`test_pe_input.py`): variable is **sent** at ≥1.715.2 / `frontier`, **omitted** on current / older / unpinned; min/max/window logic correct incl. the unknown-version asymmetry.
- Full policyengine + federal suites green (no regression from the new input or the gate).
- These prove the **mechanics** (we emit the right field, gated correctly) — not the PE-computed outcome.

**Layer 2 — manual, against the live PolicyEngine API (verified once this session):**
- **No pin / `current`:** variable omitted, status **200**, SSI computes (no 400). ✅
- **`1.715.2` WITH the variable (the fix):** present, status **200**, disabled CO member gets `ssi = 11604`. ✅
- **`1.715.2` WITHOUT the variable (the regression we prevent):** same household, variable stripped → status **200**, `ssi = 0`. ✅ This is the load-bearing proof — same version, same household, the only difference is the variable, and it's the difference between `11604` and `0`. (Confirms the fix is necessary, not cosmetic — the matching `11604` between current and frontier-with-fix could otherwise mask a no-op.)
- Sending the variable on the **current model (1.691.1)** → **400** *"Variable meets_ssi_disability_criteria … is not available …"* — why the gate omits it below 1.715.2.
- Not repeatable in CI (needs live auth + network + the version pin).

**Layer 3 — deferred to [MFB-1103](https://linear.app/myfriendben/issue/MFB-1103/pe-frontier-regression-diff-assets-snap-abawd-tx-ccs-ma-eaedc) (regression diff):**
- The repeatable behavioral check — disabled member gets non-zero SSI **on frontier**, plus the downstream SSI-recipient Medicaid amounts (MA/CO/NC/IL/WA) — runs there via `?pe_version=1.715.2` before/after diffs, the only context that actually exercises frontier.

## Deployment

- Production config updates: none on merge. Shipping this code while prod is on the current model is safe — the gate omits the variable.
- **Activation requires an explicit pin to ≥ 1.715.2** via MFB-1112's `PolicyEngineConfig`. ⚠️ This is **mandatory, not "or wait for June 10."** An unpinned request resolves to `version=None`, which the gate treats as below the floor → the variable stays **omitted even after PE flips its server-side default to frontier on June 10**. So if we reach June 10 unpinned, frontier becomes the served model but `meets_ssi_disability_criteria` is still omitted → disabled members' `ssi` drops to 0 — the exact bug this PR prevents. The pin must land **before** June 10; the default flip is not an activation path. (This is the safer design — activation is under our control, not tied to PE's release calendar.)
- Notify team/users of: the cutover sequencing (this fix activates only when the version is pinned ≥ 1.715.2)

## Notes for Reviewers

- **Hard version coupling:** unlike MFB-1112 (whose config defaults to a no-op), this variable is itself incompatible with the current model. The `min_pe_version` gate is what makes it dormant-safe — without it, merging to prod on the current model would 400 every SSI calc. Please sanity-check the gate logic in `pe_input` / `pe_versions.version_supports` (the version helpers now live in the shared `programs/programs/policyengine/versions.py` module, added in MFB-1112).
- **Reusable pattern + its boundary:** the gate is an optional `[min_pe_version, max_pe_version]` window on the dependency base class, covering all three add/remove shapes — new variable (`min`), removed variable (`max`), windowed (`both`). Unknown/unpinned version fails any floor but satisfies any ceiling (new vars stay off until opted in; still-present vars stay on until pinned past removal). It only gates SEND/OMIT — it does **not** handle a variable whose value/format changes per version (e.g. `county_str`→`county_fips`); that mechanism is deferred to MFB-1104 and the limitation is documented in `dependencies/base.py`. The min/max window comparison itself lives in `versions.version_supports`.
- **Scope confirmed from source, not assumed:** the value logic (include blindness) and breadth (Ssi-only) were both verified against the policyengine-us source rather than guessed.

### DIFF
--- a/programs/programs/policyengine/policy_engine.py
+++ b/programs/programs/policyengine/policy_engine.py
@@ -104,6 +104,11 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator], pe_version: 
             "marital_units": {},
         }
     }
+    # Resolve once: used both to gate version-specific inputs below and to set the
+    # top-level "version" field. None => no pin (PolicyEngine's current default).
+    version = pe_versions.determine_pe_version(pe_version)
+    comparable_version = pe_versions.to_comparable_pe_version(version)
+
     members: list[HouseholdMember] = screen.household_members.all()
     relationship_map = screen.relationship_map()
 
@@ -137,6 +142,17 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator], pe_version: 
 
     for program in programs:
         for Data in program.pe_inputs + program.pe_outputs:
+            # Skip inputs that the resolved model version doesn't define yet — sending
+            # an unknown variable 400s the whole request (e.g. meets_ssi_disability_criteria
+            # on 1.691.1). With no pin (comparable_version is None) we omit gated inputs
+            # too, since the unpinned default is the current model that lacks them.
+            if not pe_versions.version_supports(
+                comparable_version,
+                getattr(Data, "min_pe_version", ()),
+                getattr(Data, "max_pe_version", ()),
+            ):
+                continue
+
             period = program.pe_period
             if hasattr(program, "pe_output_period") and Data in program.pe_outputs:
                 period = program.pe_output_period
@@ -170,7 +186,6 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator], pe_version: 
         del raw_input["household"]["tax_units"][SECONDARY_TAX_UNIT]
 
     # Inject the resolved version (override > config); None means omit the field.
-    version = pe_versions.determine_pe_version(pe_version)
     if version is not None:
         raw_input["version"] = version
 
--- a/programs/programs/policyengine/policy_engine.py
+++ b/programs/programs/policyengine/policy_engine.py
@@ -144,10 +144,17 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator], pe_version: 
         for Data in program.pe_inputs + program.pe_outputs:
             # Skip inputs that the resolved model version doesn't define yet — sending
             # an unknown variable 400s the whole request (e.g. meets_ssi_disability_criteria
+<<<<<<< HEAD
             # on 1.691.1). With no pin (comparable_version is None) we omit gated inputs
             # too, since the unpinned default is the current model that lacks them.
             if not pe_versions.version_supports(
                 comparable_version,
+=======
+            # on 1.691.1). With no pin (parsed_version is None) we omit gated inputs too,
+            # since the unpinned default is the current model that lacks them.
+            if not _version_supports(
+                parsed_version,
+>>>>>>> c6aff7e3 (MFB-1102: add max_pe_version — full version-window gating (all 3 shapes))
                 getattr(Data, "min_pe_version", ()),
                 getattr(Data, "max_pe_version", ()),
             ):
--- a/programs/programs/policyengine/policy_engine.py
+++ b/programs/programs/policyengine/policy_engine.py
@@ -104,8 +104,10 @@ def pe_input(screen: Screen, programs: List[PolicyEngineCalulator], pe_version: 
             "marital_units": {},
         }
     }
-    # Resolve once: used both to gate version-specific inputs below and to set the
-    # top-level "version" field. None => no pin (PolicyEngine's current default).
+    # Two values from one resolved version, for two consumers:
+    #   version (string)            -> version to send in PE API request body (if None/omitted,
+    #                                   PE defaults to current version)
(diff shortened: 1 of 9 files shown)


programs/programs/policyengine/policy_engine.py
```diff
@@ -37,6 +37,45 @@ def resolve_pe_version(pe_version_override: Optional[str] = None) -> Optional[st
...
+
+
+def _parse_version(version: Optional[str]) -> Optional[tuple]:
+    """
+    Turn a resolved version into a comparable tuple for input gating:
+      - "1.715.2"  -> (1, 715, 2)
+      - "frontier" -> _LATEST (newest released model, supports gated inputs)
+      - "current" / None / unparseable -> None (treat as the current default model)
+    """
+    if version == "frontier":
```
david-j-stoll: Typically `latest` is the shorthand for the most recent version. We may want to verify with Policy Engine

programs/programs/policyengine/policy_engine.py
```diff
@@ -37,6 +37,45 @@ def resolve_pe_version(pe_version_override: Optional[str] = None) -> Optional[st
...
+        return _LATEST
+    if not version or version == "current":
```
david-j-stoll: We should enumerate these values for quality, granted we are allowing strings. The values would either need to explicitly be an enumerated string of "current" or "frontier" or match the version number regex. That validation can and should be consolidated so it's referenced in your other PR as well.

david-j-stoll (changes requested):

david-j-stoll (approved):

PR #1566, opened 2026-06-04
MFB-1022: LIHEAP estimates too high

david-j-stoll:
## Context & Motivation

- The `IlLiheap` calculator previously used PolicyEngine only for an income eligibility boolean (`il_liheap_income_eligible`), then returned hard-coded annual benefit amounts sourced from the 2025 IL benefit matrix. PolicyEngine now fully implements `il_liheap` and uses the 2026 benefit matrix as its source of truth, making its values more accurate and automatically kept up to date.
- This change removes the hard-coded amounts and delegates the full benefit calculation to PolicyEngine, bringing `IlLiheap` in line with how other state LIHEAP calculators (e.g. `MaHeap`) work.

## Changes Made

- `programs/programs/policyengine/calculators/dependencies/spm.py` — Added `IlLiheap(SpmUnit)` dependency class with `field = "il_liheap"` for the PE dollar-amount output
- `programs/programs/il/pe/spm.py` — Refactored `IlLiheap` calculator:
  - Changed `pe_name` from `"il_liheap_income_eligible"` to `"il_liheap"`
  - Added energy expense inputs (`HasHeatingCoolingExpenseDependency`, `HeatingExpensePersonDependency`, `ElectricityExpenseDependency`) required by PE's IL LIHEAP calculation
  - Changed `pe_outputs` to use the new `IlLiheap` dependency
  - Removed `benefit_amounts` hard-coded dict (2025 values)
  - Simplified `household_value()` to return `int(self.get_variable())` directly, delegating eligibility and benefit calculation entirely to PE
- `programs/programs/il/pe/tests/test_spm.py` — Updated `TestIlLiheap` to reflect new calculator behavior:
  - Removed tests for `benefit_amounts`, `il_liheap_income_eligible` pe_name, and the `has_expense(["rent", "mortgage"])` check
  - Added tests for new `pe_name`, updated `pe_outputs`, new energy expense `pe_inputs`, and PE-sourced benefit amount return value

## Testing

- Migrations to run: None
- Manual testing steps:
  1. Run an IL LIHEAP screen for a single-person household with income below the eligibility threshold (e.g. $20,000/year) and confirm a non-zero benefit is returned
  2. Run an IL LIHEAP screen for a household with income above the threshold (e.g. $60,000/year) and confirm the benefit is $0
  3. Run an IL LIHEAP screen for a household that already has `il_liheap` marked as a current benefit and confirm the benefit is $0
  4. Compare returned benefit amounts across household sizes against the [2026 IL LIHEAP benefit matrix](https://liheapch.acf.gov/docs/2026/benefits-matricies/IL_BenefitMatrix_Heat-Cool_2026.pdf#page=2)

## Deployment

- Admin updates needed: None
- Notify team/users of: IL LIHEAP benefit amounts will reflect 2026 values from PolicyEngine rather than the previously hard-coded 2025 matrix values — amounts will generally be higher

## Notes for Reviewers

- Known limitations: PolicyEngine's `il_liheap` benefit is capped by actual heating/energy expenses passed in the payload. The `HeatingExpensePersonDependency` assigns the full household heating cost to the head of household (same pattern used by `MaHeap`), since PE aggregates person-level heating expense up to the SPM unit for the payment cap.
- Future considerations: The `IlLiheapIncomeEligible` dependency class (`field = "il_liheap_income_eligible"`) has been left in place in `dependencies/spm.py` in case it is useful for other purposes, but it is no longer referenced by the `IlLiheap` calculator.

## Summary by CodeRabbit

* **New Features**
  * Illinois LIHEAP calculations now use the updated PolicyEngine-powered rules, with support for heating, cooling, and electricity-related expenses.
  * Added a new LIHEAP result type for Illinois so eligibility and benefit outputs are surfaced more consistently.

* **Bug Fixes**
  * Removed older household calculation logic that could produce inconsistent eligibility and benefit results.
  * Updated tests to reflect the new calculation flow and expected output values.

### DIFF
--- a/programs/programs/il/pe/spm.py
+++ b/programs/programs/il/pe/spm.py
@@ -52,6 +52,29 @@ class IlTanf(Tanf):
 
 
 class IlLiheap(PolicyEngineSpmCalulator):
+    """
+    Illinois Low Income Home Energy Assistance Program (LIHEAP).
+
+    Delegates the full eligibility and benefit calculation to PolicyEngine's
+    ``il_liheap`` variable (2026 benefit matrix), rather than hard-coding amounts.
+
+    PolicyEngine computes:
+      - Income eligibility: gross income <= higher of 60% SMI or 200% FPL.
+      - Benefit: an annual matrix amount keyed on fuel type, income bracket, and
+        household size, capped at the household's reported heating expense
+        (``min(matrix_amount, heating)``). PE models every IL household as
+        "All Electric" (or "Cash" when heat is included in rent).
+
+    Required PE input is ``heating_expense_person``: our screener captures heating
+    at the household level, so ``HeatingExpensePersonDependency`` assigns the full
+    amount to the head, which PE aggregates to the SPM unit for the cap. Because
+    the benefit is capped at this expense, a household reporting no heating/cooling
+    expense receives $0.
+
+    Households that already receive IL LIHEAP are flagged and handled by the
+    results layer (via ``already_has``), not in this calculator.
+    """
+
     pe_name = "il_liheap"
     pe_inputs = [
         dependency.household.IlStateCodeDependency,
@@ -61,8 +84,3 @@ class IlLiheap(PolicyEngineSpmCalulator):
         dependency.spm.ElectricityExpenseDependency,
     ]
     pe_outputs = [dependency.spm.IlLiheap]
-
-    def household_value(self) -> int:
-        if self.screen.has_benefit("il_liheap"):
-            return 0
-        return int(self.get_variable())
--- a/programs/programs/il/pe/spm.py
+++ b/programs/programs/il/pe/spm.py
@@ -1,10 +1,6 @@
-import logging
-
 from programs.programs.policyengine.calculators.base import PolicyEngineSpmCalulator
 import programs.programs.policyengine.calculators.dependencies as dependency
 from programs.programs.federal.pe.spm import Snap, SchoolLunch, Tanf
-
-logger = logging.getLogger(__name__)
 
 
 class IlSnap(Snap):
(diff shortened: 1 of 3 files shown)


programs/programs/il/pe/spm.py
```diff
@@ -56,72 +52,17 @@ class IlTanf(Tanf):
...
-        """
-        Calculate LIHEAP benefit based on hard-coded values by household size.
-
-        Uses PolicyEngine for income eligibility check (60% SMI or 200% FPL),
-        then returns hard-coded benefit amounts.
-
-        Returns annual benefit amount (not monthly).
-        """
-        # Check if user already has IL LIHEAP
         if self.screen.has_benefit("il_liheap"):
```
catonph: Let's not have this in the calculator - we check this "already has" piece automatically when generating results

programs/programs/il/pe/spm.py
```diff
@@ -56,72 +52,17 @@ class IlTanf(Tanf):
...
-        except KeyError as e:
-            logger.warning(f"PolicyEngine missing expected key for IL LIHEAP screen {self.screen.id}: {e}")
-            return 0
-        except (AttributeError, ValueError) as e:
-            logger.warning(f"Error calculating IL LIHEAP for screen {self.screen.id}: {type(e).__name__}: {e}")
-            return 0
-        except RuntimeError as e:
-            logger.warning(f"PolicyEngine API error for IL LIHEAP screen {self.screen.id}: {e}")
-            return 0
+        return int(self.get_variable())
```
catonph: I should probably know this but is int() required? If not, is there a reason to make it an int?

programs/programs/il/pe/spm.py
```diff
@@ -56,72 +52,17 @@ class IlTanf(Tanf):
 
 
 class IlLiheap(PolicyEngineSpmCalulator):
```
catonph, line 54: Would be nice to still have a doc string here

catonph (changes requested):
A couple small things - also do our validations need to be updated? And i see for testing "Run an IL LIHEAP screen for a household that already has il_liheap marked as a current benefit and confirm the benefit is $0" - does il_liheap have show_in_has_benefits set to true? otherwise that test isn't possible

programs/programs/il/pe/spm.py
david-j-stoll: Unfortunately this is one of those things that should be type checked. As of right now, we are not strictly typing that method, so wrapping this in an int ensures that this strictly typed function is returning an int at all times. Something to check again whenever we get to the mypy work

david-j-stoll:
> A couple small things - also do our validations need to be updated? And i see for testing "Run an IL LIHEAP screen for a household that already has il_liheap marked as a current benefit and confirm the benefit is $0" - does il_liheap have show_in_has_benefits set to true? otherwise that test isn't possible

il_liheap does currently have the `Show in has benefits step` toggle set to on. Should be all set there.

programs/programs/il/pe/spm.py
david-j-stoll: Good call!

catonph (approved):

PR #1569, opened 2026-06-05
MFB-869: CB Step 5a — Make join table the primary write target, replace sync

catonph:
## Context & Motivation

Part of the **New Current Benefits Workflow** milestone (`has_*` columns → `CurrentBenefit` join table). This is **Step 5a**, the backend-only step that makes the join table the primary write target.

The old `_sync_current_benefits()` read the `has_*` columns back, deleted all `CurrentBenefit` rows, and recreated them — a one-directional mirror (`has_*` → join table) that was a Phase 2 transitional mechanism. It can't survive the `has_*` columns being dropped, and it prevents the join table from being written independently. This ticket replaces it with direct writes.

- Implements [MFB-869](https://linear.app/myfriendben/issue/MFB-869/cb-step-5a-make-join-table-the-primary-write-target-replace-sync)
- Blocked by [MFB-719](https://linear.app/myfriendben/issue/MFB-719/cb-step-3a-switch-has-benefit-reads-to-join-table-benefits-api) (Step 3a, done) — `has_benefit()` already reads the join table
- Blocks [MFB-720](https://linear.app/myfriendben/issue/MFB-720/cb-step-6-update-api-contract-migrate-fe-to-new-shape-benefits-api) (Step 6, coordinated BE+FE deploy that drops `has_*`)

## Changes Made

**Serializer — new primary write path**
- `ScreenSerializer` now accepts `current_benefits: ["snap", "tanf"]` (write-only list of `name_abbreviated`). Each name is resolved to a `Program` via the WL-scoped `(white_label, name_abbreviated)` unique lookup and written directly to the join table. Names with no Program in the screen's white label are silently skipped.
- Wired through both `create()` and `update()`.
- **Backward compatibility:** when `current_benefits` is absent (old frontend), the serializer falls back to mirroring the legacy `has_*` columns into the join table — including the fan-out (`has_tanf` → tanf / co_tanf / il_tanf / …) and compound conditions (`ssi` via has_ssi OR sSI income; `ma_mass_health` via has_medicaid OR has_medicaid_hi).
- When both are sent, `current_benefits` is authoritative.

**Cleanup**
- Deleted `serializers._sync_current_benefits()` → replaced by `_write_current_benefits(screen, current_benefits)`.
- Deleted `Screen._build_benefit_map()` → its `has_*` → `name_abbreviated` mapping relocated to `serializers._benefit_map_from_has_columns()`, used only by the backward-compat path. Both this helper and the `has_*` write path are removed in Step 6 (MFB-720).
- `has_benefit()` still reads from the join table (unchanged from Step 3a); docstrings updated to drop references to the deleted sync function.

**Test migration**
- Migrated the test files that used the `has_* = True; sync_current_benefits()` dual-write pattern to direct join-table writes via a new `set_current_benefits()` test helper (exercises the new primary path):
  - `screener/tests/test_models.py`
  - `programs/programs/policyengine/calculators/dependencies/tests/test_spm.py`
  - `programs/programs/il/nurse_family_partnership/tests.py`
  - `programs/programs/urgent_needs/tx/test_urgent_needs_functions.py`
  - `programs/programs/wa/nslp/tests.py`
- The fan-out / compound `has_benefit` tests in `test_models.py` intentionally **keep** the `has_*` mirror pattern, since that mapping is the behavior under test until Step 6.
- `sync_current_benefits()` test helper now delegates to the serializer's backward-compat branch so it can't drift from production. (Helper itself is removed in Step 6 / MFB-720.)
- Added `screener/tests/test_screen_serializer.py` covering both write paths, precedence, WL-scoping, unknown-name skip, empty-list deselect-all, and write-only output.

## Testing

- Migrations to run: none (no schema changes — `has_*` columns and the `CurrentBenefit` table are unchanged)
- Manual testing steps:
  - `pytest screener/` → 202 passed
  - Migrated calculator/dependency suites + serializer/model tests → 195 passed
  - `black --check` clean; full project test collection (1903 tests) has no import errors
  - Old-frontend smoke: PATCH a screen with `has_snap: true` and no `current_benefits` → join table gets the snap variant rows; `already_has` still works on results
  - New-frontend smoke: PATCH with `current_benefits: ["snap"]` → join table has exactly the snap row

## Deployment

- Admin updates needed: none
- Notify team/users of: nothing required — this is backward compatible. The live frontend continues sending `has_*` and keeps working unchanged. The new `current_benefits` payload is consumed by the frontend in Step 6 (MFB-720), which is the coordinated BE+FE deploy that drops `has_*` acceptance.

## Notes for Reviewers

- **Backward compat is deliberate and temporary.** Because this ships before the Step 6 FE switch, the API must keep accepting `has_*`. The `_benefit_map_from_has_columns()` helper is the relocated `_build_benefit_map` body verbatim (model-attr references rebased to the `screen` arg) and is deleted in Step 6.
- Known limitations: the new `current_benefits` path does not reverse-map into the `has_*` columns. It doesn't need to — `has_benefit()` reads the join table, and the remaining dbt mart was already swapped to the join table in Step 4 (MFB-867). The `has_*` columns are dropped in Step 7a (MFB-878).
- Future considerations: single-benefit toggle endpoint for the results page is split out to [MFB-1094](https://linear.app/myfriendben/issue/MFB-1094) (independent of this change).

## Summary by CodeRabbit

* **New Features**
  * Screen updates now accept an explicit list of current benefits on write, improving how selected benefits are recorded.

* **Refactor**
  * Benefit storage/lookup was reworked to rely on the canonical join-table path, improving consistency for compound and derived benefits (e.g., SSI variants).

* **Tests**
  * Test suite updated and expanded to validate the new write path, white-label scoping, and backward-compat fallbacks.

### DIFF
--- a/screener/serializers.py
+++ b/screener/serializers.py
@@ -336,12 +336,25 @@ def _write_current_benefits(screen, current_benefits):
             # Primary path: frontend sent explicit program names. Resolve each in
             # this screen's white label; silently drop any this WL doesn't offer.
             requested = set(current_benefits)
-            program_ids_to_write = list(
-                Program.objects.filter(
-                    white_label=screen.white_label,
-                    name_abbreviated__in=requested,
-                ).values_list("id", flat=True)
-            )
+            resolved = Program.objects.filter(
+                white_label=screen.white_label,
+                name_abbreviated__in=requested,
+            ).values_list("id", "name_abbreviated")
+            program_ids_to_write = [program_id for program_id, _ in resolved]
+
+            # A requested name with no Program in this WL is dropped rather than
+            # erroring (a WL only writes the programs it offers). During the Step 6
+            # FE cutover that's also how a typo'd / stale name_abbreviated vanishes,
+            # so log it — visible in Heroku logs (heroku logs -a cobenefits-api),
+            # not Sentry, since it's info-level and benign.
+            dropped = requested - {name for _, name in resolved}
+            if dropped:
+                logger.info(
+                    "current_benefits: dropped %d name(s) not offered by white_label %s: %s",
+                    len(dropped),
+                    screen.white_label_id,
+                    sorted(dropped),
+                )
 
         CurrentBenefit.objects.filter(screen=screen).delete()
         if program_ids_to_write:
--- a/screener/serializers.py
+++ b/screener/serializers.py
@@ -138,7 +138,7 @@ class HouseholdMemberSerializer(serializers.ModelSerializer):
         read_only_fields = ("screen", "id")
 
 
-def _benefit_map_from_has_columns(screen) -> dict:
+def _benefit_map_from_has_columns(screen: Screen) -> dict[str, bool]:
     """
     Returns the full `name_abbreviated → bool` map for a screen derived from its
     legacy `has_*` columns, including the fan-out (one `has_*` column → many WL
@@ -309,7 +309,7 @@ def _benefit_map_from_has_columns(screen) -> dict:
 _SSI_BENEFIT_NAMES = frozenset({"ssi", "tx_ssi", "wa_ssi", "cesn_ssi"})
 
 
-def _derived_current_benefit_names(screen) -> set[str]:
+def _derived_current_benefit_names(screen: Screen) -> set[str]:
     """`name_abbreviated` values implied by screen state, independent of the
     frontend's current-benefits toggles.
 
@@ -329,7 +329,7 @@ def _derived_current_benefit_names(screen) -> set[str]:
     return derived
 
 
-def _write_current_benefits(screen, current_benefits):
+def _write_current_benefits(screen: Screen, current_benefits: list[str] | None) -> None:
     """
     Write the CurrentBenefit join table for `screen`, replacing any existing rows.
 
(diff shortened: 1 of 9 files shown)


screener/serializers.py
```diff
@@ -137,24 +138,262 @@ class Meta:
...
+    derived: set[str] = set()
+    if screen.calc_gross_income("yearly", ("sSI",)) > 0:
+        derived |= _SSI_BENEFIT_NAMES
+    return derived
 
-    Uses select_for_update() inside a transaction to serialize concurrent PATCH requests
-    on the same screen and prevent races on the delete+bulk_create. The screen is re-fetched
-    inside the atomic block so program_ids_to_write reflects the post-lock committed state.
+
+def _write_current_benefits(screen, current_benefits):
```
david-j-stoll: Can we add type hints here?

elliottwaissbluth (approved):
If the test scenarios passed like the PR description specifies, this code looks good to me.

screener/serializers.py
```diff
@@ -137,24 +138,262 @@ class Meta:
         read_only_fields = ("screen", "id")
 
 
-def _sync_current_benefits(screen):
+def _benefit_map_from_has_columns(screen) -> dict:
```
david-j-stoll: Can we add type hints here?

david-j-stoll (changes requested):
Small request! Looks good otherwise

PR #1573, opened 2026-06-08
MFB-720: CB Step 6 — drop legacy has_* current-benefits contract (benefits-api)

catonph:
## Context & Motivation

CB Step 6 of the New Current Benefits Workflow milestone — the coordinated BE+FE deploy that retires the legacy `has_*` current-benefits contract. By this step the `CurrentBenefit` join table is already the authoritative read source (Step 3a / MFB-719) and the primary write target (Step 5a / MFB-869). This PR removes the remaining backward-compat `has_*` write path so the join table is the **sole** source, and makes `current_benefits` the read/write representation on the screen.

**Ships together with** the benefits-calculator MFB-720 PR — the serializer stops accepting `has_*` benefit fields and the FE round-trips `current_benefits`, so both repos must deploy together.

- Related issue: MFB-720
- Paired FE PR: MyFriendBen/benefits-calculator#2123 (CB Step 6 — migrate FE to current_benefits Set contract)

### Contract: drop the legacy `has_*` write path
- **Deleted `serializers._benefit_map_from_has_columns()`** — the ~120-line `name_abbreviated → bool` map relocated from `Screen._build_benefit_map()` in Step 5a. It only backed the backward-compat write path.
- **Dropped the backward-compat branch in `_write_current_benefits()`** — it no longer accepts `current_benefits=None`; `current_benefits` is a required arg.
- **`ScreenSerializer.current_benefits`** is a write-only input (bounded at 256 / child 64; `required=False` — see the optional-field note below). Removed the **71 benefit `has_*` fields** from `Meta.fields`.
  - **Kept** `has_benefits` (the "did the user answer the benefits question" enum) and the member-insurance fields (`has_employer_hi`, `has_private_hi`, `has_medicaid_hi`, `has_medicare_hi`, `has_chp_hi`, `has_no_hi`, `has_va`) — a separate, insurance-keyed concern handled by MFB-1084, not part of this migration.
- `create()` / `update()` pop `current_benefits` without a `None` default.

### Read path: echo `current_benefits` on the screen, and fix the cache staleness it exposed
- **`ScreenSerializer.to_representation` injects `current_benefits`** as the sorted list of join-table `name_abbreviated`. The input field stays write-only; the read value is supplied here because the `current_benefits` attribute is the reverse relation (a manager), not a list of names. This is what lets the FE rehydrate the Step-8 selections on reload/resume.
- **Added `Screen.invalidate_current_benefits_cache()`** — clears **both** per-instance caches over the join table: the `_current_benefit_names` cached_property (backs `has_benefit()`) and the `current_benefits` relation prefetch (backs the serializer read). Replaces the prior "don't `del` this" warning with a supported hook.
- **`create()`/`update()` call it after `_write_current_benefits`** so a same-request read-back reflects the write. This also fixes `has_benefit()` returning stale results after a same-instance write — **so the cache-invalidation concern previously deferred to MFB-878 is resolved here** (MFB-878 updated accordingly).
- **Prefetch `current_benefits__program`** on the `ScreenViewSet` queryset so the read costs no extra query on retrieve/update.

### Tests & cleanup
- **Trimmed `screener/tests/helpers.py` to `seed_program` only** — deleted the transitional `sync_current_benefits` / `set_current_benefits` helpers (they only mirrored dual-write while `has_*` was a write target). `seed_program` is a general-purpose Program fixture used by 7 test modules and is retained.
- **Rewrote the `has_benefit()` tests** to write via `_write_current_benefits(screen, [...])`; kept the SSI-derived-from-income guard at the join-table level; deleted the obsolete `ma_mass_health` / `has_*` fan-out tests (those exercised the deleted map). Migrated four program test files off `set_current_benefits`.
- **New read/cache tests:** serializer read echoes join-table names (empty → `[]`); create/update responses reflect the written set; regression guard that `update` returns the **new** set with a prefetched instance (the exact bug); model tests that invalidation refreshes both `has_benefit()` and the relation, documents the stale-without-invalidation hazard, and is safe when caches are unpopulated.

### Lowercase `name_abbreviated` invariant (code review)
The FE now matches tile keys against `name_abbreviated` exactly (it dropped `.toLowerCase()`), so the BE guarantees lowercase:
- `Program.save()` lowercases `name_abbreviated`; a **CheckConstraint** (`name_abbreviated = lower(name_abbreviated)`, migration `0158`) enforces it even for `bulk_create` / `.update()` / raw SQL.
- **Prod audit:** `WHERE name_abbreviated <> lower(name_abbreviated)` and the case-collision check both returned **zero rows** — so no data backfill is needed and the constraint applies cleanly. Confirmed, not assumed.
- Lowercased pre-existing uppercase test fixtures (`test_has_benefits_programs.py`, `test_import_program_config.py`) that now violate the constraint.

### `current_benefits` is optional (defaults to empty)
- `required=False`; `create()/update()` treat an absent key as `[]`. The FE always sends it; this keeps non-FE serializer clients (validation imports, screen-pull commands) from 400ing. Replaced the earlier `required=True`.

**Intentionally retained:** `has_benefit_from_list()` (two prod callers — CO affordable residential energy, NC Head Start) and `_derived_current_benefit_names()` / `_SSI_BENEFIT_NAMES`.

## Testing

- Migrations to run: **none** (no schema change — the `has_*` columns stay on the DB; dropping them is Step 7 / MFB-878)
- Manual testing steps:
  - `black .` — clean
  - `pytest screener/ programs/programs/{wa/nslp,co/energy_calculator/affordable_residential_energy,nc/nc_head_start,il/nurse_family_partnership,policyengine/calculators/dependencies/tests/test_spm.py,urgent_needs/tx/test_urgent_needs_functions.py}` → **326 passed**
  - Verified end-to-end locally: Step-8 selections POST as `current_benefits: [...]` (no `has_*`), and the response now echoes `current_benefits` so selections persist on back-nav/reload.

## Deployment

- Admin updates needed: none
- **External webhook consumers (action required before deploy):** `screener/webhooks.py` sends the full `ScreenSerializer(screen).data` to referrer subscribers with the `send_screen` webhook function. This payload **loses the ~71 `has_*` benefit keys and gains `current_benefits`**. A prod query (referrers with `send_screen` + a `webhook_url`) found **three active subscribers** that must be asked to confirm they don't read `has_*`:
    - **Benefits in Action** (`bia`, CO)
    - **Colorado Design Insight Group** (`ccig`, CO)
    - **NC Navigators Consortium** (`ncnc`, NC)
  Their endpoints are external (partner-owned), so this can't be verified from the codebase — outreach to each is needed before deploy. The lockstep-with-FE note does **not** cover these subscribers.
  - **N+1 from the new read path is handled in-code:** `to_representation` now reads the join table, and the webhook loads the screen outside the viewset's prefetched queryset, so `screen_data()` re-fetches with `current_benefits__program` prefetched (matching the viewset / eligibility paths). No per-send extra query. The only remaining webhook action is the partner outreach above.
- Notify team/users of: **must deploy in lockstep with the paired benefits-calculator PR** — the serializer now requires `current_benefits`, no longer reads `has_*`, and echoes `current_benefits` on read. An old FE talking to this BE (or vice-versa) breaks the current-benefits step.

## Notes for Reviewers

- **Cache design:** there are two per-instance caches over the same join-table data — `_current_benefit_names` (memoized set for the ~33 `has_benefit()` calls per eligibility run) and the relation prefetch (avoids N+1). They're layered (the cached_property reads the relation), not redundant. The read-after-write hazard is confined to the **write path**; `invalidate_current_benefits_cache()` clears both there. The 33 read-only calculator callers are unaffected (they never write mid-run).
- **`update()` coupling (by design):** a PATCH deletes+recreates household members from the payload, so SSI-from-income derivation reads the **payload's** income. The FE always sends the full screen on every PATCH, so income is always present — fine, just noting the coupling.
- **WaNslp categorical reads (fixed in a prereq PR):** `WaNslp` read `has_snap` / `has_tanf` / `has_head_start` / `has_early_head_start` as raw columns for categorical eligibility. Since this PR stops writing those columns, that pathway would go dead. **#1575 migrates WaNslp to read the `CurrentBenefit` join table and must merge before/with this PR.** (Converting WaNslp to PolicyEngine entirely is MFB-1119.)
- **Known limitation:** `WaNslp` reads `has_snap` / `has_tanf` / `has_head_start` / `has_early_head_start` as **raw Screen columns** (`programs/programs/wa/nslp/calculator.py:130,132`) for categorical eligibility — a separate consumer from the current-benefits join table, untouched here. Columns remain until Step 7 (MFB-878). Converting WaNslp to PolicyEngine is filed as MFB-1119.
- **Deviation from the ticket text:** it says "delete `screener/tests/helpers.py`," but that file also exports `seed_program` (a general fixture used by 7 modules). Trimmed to `seed_program` rather than deleting — same intent (kill the transitional helpers), no collateral breakage.

## Summary by CodeRabbit

* **Refactor**
  * current_benefits is now a required list payload for create/update (absent → treated as empty) and responses echo committed current_benefits; legacy per-benefit booleans removed.
  * Program identifiers are normalized to lowercase on save and enforced at the database level.

* **Improvements**
  * Server preloads current_benefits relations to avoid per-item queries.
  * Writes report dropped frontend-requested benefits; cache invalidation API added so reads reflect recent writes.

* **Tests**
  * Many tests updated to the new current_benefits helpers and lowercase program contract; added cache-invalidation coverage.


screener/models.py
```diff
@@ -375,49 +375,43 @@ def has_benefit_from_list(self, names: list[str]):
...
         """
         Returns True if the user has declared they already receive the benefit
         identified by `name_abbreviated`. Drives the "already_has" flag on
         eligibility results so the frontend can filter the program out.
 
-        Reads from the CurrentBenefit join table — the primary write target as
-        of Step 5a (MFB-869). The join table is written on every POST/PATCH by
-        the serializer's `_write_current_benefits()`: directly from the
-        frontend's `current_benefits` list, or (backward compat, until Step 6)
-        mirrored from the legacy has_* columns.
```
david-j-stoll, line 425: Just a flag from Claude:

`12 fixtures rely on has_* benefit keys: wa_nslp, wa_wic, wa_lifeline, wa_orca_lift, tx_head_start, tx_early_head_start, tx_wap, tx_ccad, ma_head_start, ma_early_head_start, ma_middle_income_rental, il_nfp. Any whose eligibility turns on a categorical-benefit or already_has signal will regress.`

In addition, we should do a quick review of our skills to make sure that any `has_` references have been removed.

screener/serializers.py
```diff
@@ -619,9 +387,9 @@ def update(self, instance, validated_data):
         household_members = validated_data.pop("household_members")
         expenses = validated_data.pop("expenses")
         energy_calculator_screen = validated_data.pop("energy_calculator", None)
-        # Not a Screen column — pop before the bulk .update(). None (key absent) →
-        # old frontend, mirror from has_*; a list → new frontend writes directly.
-        current_benefits = validated_data.pop("current_benefits", None)
+        # Not a Screen column — pop before the bulk .update() and write separately.
+        # Absent (non-frontend clients) is treated as no current benefits.
+        current_benefits = validated_data.pop("current_benefits", [])
```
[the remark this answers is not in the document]
catonph: There's no partial-update path in this codebase that can hit this. update() on ScreenSerializer always receives the full screen — the FE PATCHes the entire screen on every step (it deletes and recreates household members from the payload, which is why income/SSI derivation reads the payload), so current_benefits is always present. required=False defaulting to [] is deliberate: it's there so non-FE serializer clients (validation imports, the pull_screen command) don't 400 — not to support partial writes. There's no DRF partial=True / partial_update call site (your own scan came back empty on that), so a benefit-clobbering partial PATCH can't occur. No change needed.

screener/models.py
catonph: Confirmed false positives except for a couple which were handled in 1575

david-j-stoll (approved):

PR #1518, opened 2026-05-14
Add WA TANF (wa_tanf)

elliottwaissbluth:
## Context & Motivation

- Implements [MFB-749](https://linear.app/myfriendben/issue/MFB-749)
- Adds Washington State TANF (Temporary Assistance for Needy Families) as a PolicyEngine-based SPM calculator

## Changes Made

- **Research files**: Added `programs/programs/wa/tanf/spec.md`, initial config JSON, and validation test cases
- **New dependencies**: `EmploymentIncomeBeforeLsrDependency` and `SelfEmploymentIncomeBeforeLsrDependency` in `member.py` — these feed PE's `tanf_gross_earned_income` so the $500 flat + 50% earned income disregard (WAC 388-450-0170) is applied automatically
- **New dependencies**: `WaTanf` output and `WaShowAllCashAssistanceProgramsDependency` (bypasses PE per-member immigration check; screener handles via `legal_status_required`) in `spm.py`
- **Calculator**: `WaTanf` class in `wa/pe/spm.py` inheriting from federal `Tanf`, registered in `wa/pe/__init__.py`
- **Dependency tests**: Full coverage for `EmploymentIncomeBeforeLsrDependency`, `SelfEmploymentIncomeBeforeLsrDependency`, `WaTanf`, and `WaShowAllCashAssistanceProgramsDependency`
- **Initial config**: Creates `wa_cash` category, program with legal statuses, documents, and DSHS CSO navigator
- **Test cases**: 3 validation scenarios — eligible HH3, eligible HH2 (edge case), ineligible (income exceeds payment standard after disregard)

## Testing

- Migrations to run: None
- Configuration updates needed: Run `python manage.py import_program_config programs/management/commands/import_program_config_data/data/wa_tanf_initial_config.json`
- Environment variables/settings to add: None
- Manual testing steps:
  1. Import config (above)
  2. Import validations: `python manage.py import_validations validations/management/commands/import_validations/data/wa_tanf.json`
  3. Activate program: set `wa_tanf` active via admin
  4. Run `python manage.py validate --white-label wa --program wa_tanf` — all 3 should pass

## Deployment

- Post-deployment scripts needed: Run import_program_config for wa_tanf_initial_config.json, then activate program
- Production config updates: None
- Admin updates needed: Activate wa_tanf program after import
- Notify team/users of: New WA TANF program available

## Notes for Reviewers

[...]

PR #1575, opened 2026-06-09
Migrate raw has_* calculator reads to the join table (WaNslp + SNAP-student + Energy EBT) — pre-req for MFB-720

catonph:
## Context & Motivation

Several calculator code paths read raw `screen.has_*` benefit columns directly. The CB Step 6 cutover (MFB-720, #1573) stops writing those columns, so after it deploys those reads silently return the DB default (`False`) forever — quietly changing eligibility output. This PR migrates **all** such reads to the `CurrentBenefit` join table (`has_benefit` / `has_benefit_from_list`), the authoritative read source since MFB-869.

Affected reads:
- **WaNslp** categorical eligibility — read `has_snap` / `has_tanf` / `has_head_start` / `has_early_head_start` (`programs/programs/wa/nslp/calculator.py`).
- **`snap_ineligible_student()`** (`programs/programs/helpers.py`) — the "household receives TANF" SNAP-student exemption (E6); consumed by the PE dependency `is_snap_ineligible_student` and the CO SNAP-student warning.
- **`NcSnapIneligibleStudentDependency`** (`policyengine/.../member.py`) — the NC SNAP-student TANF exemption.
- **Energy EBT** (`co/energy_calculator/energy_ebt/calculator.py`) — the `has_leap` "no LEAP" gate for the `cesn_energy_ebt` calculator.

The CB Step 6 cutover (MFB-720, #1573) **stops writing those columns** — the frontend now sends `current_benefits: [...]` and the serializer no longer mirrors into `has_*`. Left unchanged, WaNslp's categorical pathway silently goes dead once MFB-720 deploys: those columns stay frozen at their `False` default, so a high-income WA household that selects SNAP/TANF/Head Start on the current-benefits step would no longer be screened categorically eligible for NSLP — it would fall through to the income test only.

This was surfaced reviewing what becomes stale after MFB-720. This PR moves WaNslp's categorical reads onto the `CurrentBenefit` join table (the authoritative read source since MFB-869) and drops a redundant self-check.

- Related: MFB-720 (#1573), MFB-1119 (WaNslp → PolicyEngine conversion, which will replace this logic wholesale)

## Changes Made

- **Categorical reads → join table.** The SNAP/TANF/Head Start categorical check now reads the `CurrentBenefit` join table (`screen.has_benefit(...)`) instead of the `has_*` columns. The qualifying programs live in a `presumptive_eligibility` class attribute — `wa_snap`, `wa_tanf`, `wa_head_start` — the only relevant `name_abbreviated` values a WA screen's current benefits can contain (matching the convention other calculators use). The single-use foster-child categorical helper is inlined into `household_eligible()`. (WA offers no Early Head Start program, so that pathway is dropped — it can't fire on a real WA screen.)
- **Migrated the SNAP-student TANF reads via a structural `base_program` query.** `snap_ineligible_student()` (E6) and `NcSnapIneligibleStudentDependency` (E6) now call a new `Screen.has_base_benefit("tanf")` instead of reading `screen.has_tanf`. `has_base_benefit` reads the current-benefits join table grouped by `Program.base_program`, so it matches *any* TANF variant — tanf / co_tanf / wa_tanf / ma_tafdc / … — without a hand-maintained name list. This deliberately avoids enumerating variants: `wa_tanf` was added as a PE program in #1518/MFB-749 and would have been missed by a name list (it carries `base_program: "tanf"`, so the structural query picks it up automatically). Energy EBT's LEAP gate uses `has_benefit("cesn_leap")` — in the CESN calculator only `cesn_leap` can appear in a screen's current benefits (it's the LEAP variant with `show_in_has_benefits_step=True` on the CESN white label); `leap` (CO WL) and `nc_leap` (NC WL) are never rendered on the CESN screener, so they could never be true here.
- **Dropped the redundant "already has NSLP" guard.** `household_eligible()` previously barred households with `not has_benefit("nslp")`. That duplicates the results layer, which sets each program's `already_has` flag centrally (`screen.has_benefit(program.name_abbreviated)` in `screener/views.py`) and filters it on the frontend. The guard was also subtly wrong — it checked the bare name `"nslp"` while WA only offers `"wa_nslp"`, so it never fired on a real WA screen anyway.
- **Tests:** categorical tests seed the program and write the join table (`seed_program` + `set_current_benefits`) instead of setting `has_*` column kwargs; dropped the now-unused benefit-column defaults from `_screen_base`; removed `test_ineligible_already_has_nslp` (it exercised the removed guard and had to seed a bare `nslp` program WA doesn't offer).

## Testing

- Migrations to run: none
- `black` clean
- `pytest programs/programs/wa/nslp` → **15 passed**; full suite (fresh DB) → **1888 passed**
- Verified the calculator no longer references any `has_*` benefit column; categorical reads go through the join table, and self-filtering relies on the central `already_has` flag.

## Deployment

- **Merge ordering — must ship with or after the MFB-720 pair (#1573 BE + #2123 FE), not ahead of it.** This is about the **WaNslp categorical** read specifically. Most migrated reads are independently safe today because the legacy `has_*` mirror (`_benefit_map_from_has_columns()`) populates the join table on the pre-720 path: it writes `wa_snap` / `wa_head_start` (WaNslp categorical), and `tanf` / `nc_tanf` etc. all carry `base_program: "tanf"` (so the SNAP-student `has_base_benefit("tanf")` reads match). **WaNslp's TANF categorical read is the exception** — it calls `has_benefit("wa_tanf")` directly (by exact `name_abbreviated`, not base_program), and the mirror has no `wa_tanf` row (the variant was added in #1518/MFB-749 without one). So on the pre-720 path that single read is always `False`, where the old `screen.has_tanf` column read worked. If #1575 lands on `main` ahead of #1573/#2123, WA NSLP TANF-categorical eligibility breaks in that window.
- **The SNAP-student TANF reads are safe standalone** — `has_base_benefit("tanf")` matches by `Program.base_program`, and the pre-720 mirror writes the non-WA `tanf` variants (all `base_program: "tanf"`), so a WA TANF household trips E6 both pre-720 (mirror) and post-720 (FE sends `wa_tanf`, also `base_program: "tanf"`). No window.
- **No standalone mirror fix is warranted for the WaNslp categorical gap** — #1573 deletes `_benefit_map_from_has_columns()` wholesale, after which the new FE (#2123) sends `wa_tanf` directly in `current_benefits` and `has_benefit("wa_tanf")` resolves correctly. Adding a `wa_tanf` row to a map that's being deleted this week would be churn and would likely conflict with #1573. The durable fix is the merge ordering above.
- No coordinated FE deploy needed *beyond* the MFB-720 FE PR (#2123) already in flight.

## Notes for Reviewers

- WA offers no Early Head Start program, so that categorical pathway (and the column the old code read) is dropped — it can't fire on a real WA screen.
- This is a stopgap that makes WaNslp correct on the join table; MFB-1119 will replace WaNslp with a PolicyEngine-backed calculator entirely.
- Related cleanup opportunity (not in this PR): other calculators that self-guard against *their own* program name could likewise rely on the central `already_has` flag. (Guards against *other* programs — e.g. SunBucks barred by SNAP/TANF — are real eligibility rules and should stay.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved eligibility determination for Energy Assistance programs by refining how household benefits are assessed.
  * Enhanced SNAP student exemption logic for more accurate eligibility evaluation.

* **Refactor**
  * Updated Washington NSLP categorical eligibility assessment to align with current benefit records for more consistent program evaluation.

### DIFF
--- a/programs/programs/wa/nslp/calculator.py
+++ b/programs/programs/wa/nslp/calculator.py
@@ -12,7 +12,8 @@ class WaNslp(ProgramCalculator):
     Estimates free or reduced-price school lunch eligibility using OSPI 2025–26
     income guidelines (not raw FPL math), frequency-matched income comparison
     when a single pay frequency applies, and screenable categorical pathways
-    (SNAP/Basic Food, TANF, Head Start flags, foster-child proxy).
+    (SNAP/Basic Food, TANF, Head Start, foster-child proxy) read from the
+    household's current benefits.
 
     Value: $828/year per likely eligible student ($4.60 × 180 days, SY 2025–26).
 
@@ -126,12 +127,25 @@ class WaNslp(ProgramCalculator):
         gross = Decimal(str(self.screen.calc_gross_income("yearly", ["all"])))
         return gross <= red_ann
 
+    # name_abbreviated values that count as each categorical benefit. The current
+    # benefits step records the white-label-specific variant the user selected
+    # (e.g. wa_snap on a WA screen), so we check every variant. Read via the
+    # CurrentBenefit join table (has_benefit_from_list), not the legacy has_*
+    # columns — those are no longer written as of the Step 6 cutover (MFB-720).
+    _SNAP_NAMES = ("snap", "co_snap", "il_snap", "ma_snap", "nc_snap", "tx_snap", "cesn_snap", "wa_snap")
+    _TANF_NAMES = ("tanf", "co_tanf", "il_tanf", "nc_tanf", "tx_tanf", "cesn_tanf", "wa_tanf")
+    _HEAD_START_NAMES = ("head_start", "co_head_start", "ma_head_start", "tx_head_start", "wa_head_start")
+    _EARLY_HEAD_START_NAMES = ("early_head_start", "ma_early_head_start", "tx_early_head_start")
+
     def _household_categorical(self) -> bool:
-        if self.screen.has_snap or self.screen.has_tanf:
-            return True
-        if self.screen.has_head_start or self.screen.has_early_head_start:
-            return True
-        return False
+        return self.screen.has_benefit_from_list(
+            [
+                *self._SNAP_NAMES,
+                *self._TANF_NAMES,
+                *self._HEAD_START_NAMES,
+                *self._EARLY_HEAD_START_NAMES,
+            ]
+        )
 
     def _foster_school_age_categorical(self) -> bool:
         for m in self.screen.household_members.all():
--- a/programs/programs/wa/nslp/calculator.py
+++ b/programs/programs/wa/nslp/calculator.py
@@ -127,25 +127,13 @@ class WaNslp(ProgramCalculator):
         gross = Decimal(str(self.screen.calc_gross_income("yearly", ["all"])))
         return gross <= red_ann
 
-    # name_abbreviated values that count as each categorical benefit. The current
-    # benefits step records the white-label-specific variant the user selected
-    # (e.g. wa_snap on a WA screen), so we check every variant. Read via the
+    # The WA programs that confer NSLP categorical eligibility. Read from the
     # CurrentBenefit join table (has_benefit_from_list), not the legacy has_*
     # columns — those are no longer written as of the Step 6 cutover (MFB-720).
-    _SNAP_NAMES = ("snap", "co_snap", "il_snap", "ma_snap", "nc_snap", "tx_snap", "cesn_snap", "wa_snap")
-    _TANF_NAMES = ("tanf", "co_tanf", "il_tanf", "nc_tanf", "tx_tanf", "cesn_tanf", "wa_tanf")
-    _HEAD_START_NAMES = ("head_start", "co_head_start", "ma_head_start", "tx_head_start", "wa_head_start")
-    _EARLY_HEAD_START_NAMES = ("early_head_start", "ma_early_head_start", "tx_early_head_start")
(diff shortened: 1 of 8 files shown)


programs/programs/wa/nslp/calculator.py
```diff
@@ -126,32 +130,19 @@ def _income_at_or_below_reduced_cap(self) -> bool:
...
-        )
-
         gross_for_message = int(self.screen.calc_gross_income("yearly", ["all"]))
         income_limit_message = self._reduced_annual_limit()
 
-        categorical = self._household_categorical() or self._foster_school_age_categorical()
+        # Categorical eligibility: the household receives a qualifying benefit, or a
+        # school-age foster child is present (a foster child is categorically eligible).
+        categorical = any(self.screen.has_benefit(program) for program in self.presumptive_eligibility) or any(
+            m.relationship == "fosterChild" and self._is_school_meal_proxy_student(m)
```
david-j-stoll, line 143: Just a callout on the risk of using raw strings. I think I've seen a couple places where that's caused problems already in the code (see existing counties attached to Urgent Needs in prod. We have both "King" and "King County", but both seem to be pointing to the same meaningful geaographic area). 

I know this specific string could correspond to a key in a json. I think it's outside the scope of this PR to fix this, but something worth considering as another iteration to cleanup these relationship values

david-j-stoll (approved):

PR #1579, opened 2026-06-11
MFB:1088: Fixing bug from migration to JOIN tables

david-j-stoll:
## Context & Motivation

- Fixes MFB-1088 (TX Family Planning Program custom calculator follow-up)
- Related PR: _N/A_

The TX FPP `§4140` adjunctive income bypass — which should make an applicant with
income above 250% FPL eligible when they (or their child) are enrolled in SNAP,
WIC, or CHIP — never fired in production. A validation case (income above 250%
FPL but enrolled in SNAP, expected `$266`) returned `$0`.

Root cause: the calculator read the legacy `Screen.has_snap` / `has_wic` /
`has_chp` boolean columns. The join-table migration (MFB-720) dropped those
columns from the `ScreenSerializer` write contract, so nothing populates them
anymore — they are permanently `False` on every real screen. Current-benefit
enrollment now lives in the `CurrentBenefit` join table (`current_benefits`,
e.g. `tx_snap`), and CHIP is captured as a per-member insurance question
(`member.insurance.chp`), never as a current benefit.

This affected all three adjunctive programs (SNAP, WIC, CHIP) identically — SNAP
was simply the only one with a validation scenario, so it was the one observed.
The unit tests passed only because they mocked the dead boolean columns directly,
bypassing the real read path.

## Changes Made

- **`programs/programs/tx/fpp/calculator.py`** — `household_eligible()` now reads
  the live data surface for the §4140 bypass:
  - SNAP: `self.screen.has_benefit("tx_snap")` — `tx_snap` is a TX current-benefits
    tile (flagged in migration `0142`), so it lands in the `CurrentBenefit` join
    table under that exact name.
  - WIC: `self.screen.has_benefit("tx_wic")` (same — flagged in `0142`).
  - CHIP: `self.screen.has_insurance_types(("chp",))` — CHIP is a per-member
    insurance question ("applicant or their child"), **not** a current-benefits
    tile. `tx_chip` is a PolicyEngine *eligibility* program and is never written to
    the join table, so `has_benefit("tx_chip")` would always be `False`. This
    mirrors the sibling Healthy Texas Women calculator (`tx/htw/calculator.py`).
  - Replaced the old comment that argued the `has_*` columns were authoritative
    with one explaining they are dead post-MFB-720 and must not be used, plus why
    SNAP/WIC use the join table while CHIP uses member insurance.
  - **Fixed a critical correctness bug** introduced mid-iteration: a trailing comma
    made `adjunctive_eligible` a 1-tuple `(bool,)`, which is always truthy — so the
    250% FPL income test was bypassed for *every* household. Removed.
- **`validations/.../import_validations/data/tx_fpp.json`** — the §4140 SNAP
  scenario set `"has_snap": true`, which `ScreenSerializer` silently drops.
  Changed to `"current_benefits": ["tx_snap"]` so the imported screen actually
  records SNAP enrollment in the join table.
- **`validations/.../import_validations/test_case_schema.json`** — added a
  `current_benefits` array property (with a note that the legacy `has_*` booleans
  are no longer persisted), so the supported field is documented in the schema.
- **`programs/programs/tx/fpp/tests/test_fpp.py`** — `make_calculator()` now mocks
  the real methods (`has_benefit` for `tx_snap`/`tx_wic`, `has_insurance_types` for
  CHIP) instead of the dead `has_snap`/`has_wic`/`has_chp` attributes, so a
  regression back to reading the legacy columns — or to `has_benefit("tx_chip")`,
  which is always `False` — would fail the tests. The public test signature
  (`has_snap=`/`has_wic=`/`has_chp=`) is unchanged.

## Testing

- Migrations to run: none (no schema changes).
- Configuration updates needed: none.
- Manual testing steps:
  - Unit tests: `python -m pytest programs/programs/tx/fpp/tests/test_fpp.py`
    (26 passed locally).
  - Lint/format: `black` reports both changed files unchanged; run project
    `flake8` to confirm clean.
  - Re-import the corrected validation fixture and confirm the §4140 SNAP
    scenario now computes `$266` (eligible) instead of `$0`:
    `python manage.py import_validations validations/management/commands/import_validations/data/tx_fpp.json`
    Then view the new screen's results with `?admin=true` and confirm the
    `tx_fpp` validation passes.

## Deployment

- Post-deployment scripts needed: **Yes.** The existing prod validation screen
  (`116305ae-9f29-4865-aa0d-f19f554604f2`) was created from the old fixture, so
  its `current_benefits` join table is empty — it will **not** self-heal after the
  code deploys and will keep showing `$0`. Re-run `import_validations` with the
  corrected `tx_fpp.json` to create a fresh screen that records `tx_snap`.
  - Note: `import_validations` is **additive** — it creates a new screen per run
    and will error if a validation already exists for the same screen+program.
    Purge the stale `Validation` row(s) / screen for `tx_fpp` first to avoid
    accumulating duplicates.
- Production config updates: none.
- Admin updates needed: none beyond the validation re-import above.
- Notify team/users of: QA that previously saw the TX FPP §4140 SNAP case fail
  can re-verify; WIC/CHIP adjunctive paths were silently broken too and are now
  fixed.

## Notes for Reviewers

- Known limitations:
  - CHIP Perinatal (the 4th §4140 program) is still not collected by the screener
    and remains a data gap, as documented in the calculator docstring.
  - The unit tests remain mock-based (matching the existing file style). They now
    mirror the real method surface, but a true integration test that builds a
    `Screen` through `ScreenSerializer` with `current_benefits` and runs the
    calculator end-to-end would be the strongest guard against this class of bug.
- Future considerations:
  - The frontend still defines legacy `has_snap`/`has_wic`/`has_chp` in
    `ApiFormData`/`updateScreen` plumbing slated for removal once MFB-720 fully
    lands; confirm no other custom calculator still reads the dead `Screen.has_*`
    columns (grep found the TX FPP usage was the live offender).

## Summary by CodeRabbit

**Bug Fixes**
* Corrected adjunctive-income bypass eligibility calculation in Texas FPP to use current benefit enrollment and insurance type records for improved accuracy.

### DIFF
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -61,11 +61,20 @@ class TxFpp(ProgramCalculator):
         # §4140 adjunctive income eligibility: enrollment in SNAP, WIC, or CHIP
         # (applicant or their child) bypasses the 250% FPL income test.
         #
-        # Read the has_* columns directly rather than has_benefit(): has_benefit()
-        # resolves against the CurrentBenefit join table keyed by the white-label
-        # program name (e.g. "tx_snap"), so the generic "snap"/"wic" keys do not
-        # reliably match. The columns are the authoritative screen-level flags.
-        adjunctive_eligible = bool(self.screen.has_snap or self.screen.has_wic or self.screen.has_chp)
+        # SNAP/WIC enrollment lives in the CurrentBenefit join table, read via
+        # has_base_benefit() against the cross-white-label `base_program` grouping
+        # (tx_snap → "snap", tx_wic → "wic"). CHIP is collected as a per-member
+        # insurance question, not a current benefit, so it is read from member
+        # insurance (matches any household member — "applicant or their child").
+        #
+        # Do NOT read the legacy Screen.has_snap/has_wic/has_chp boolean columns:
+        # the join-table migration (MFB-720) dropped them from the serializer write
+        # contract, so they are never populated and are permanently False.
+        adjunctive_eligible = (
+            self.screen.has_base_benefit("snap")
+            or self.screen.has_base_benefit("wic")
+            or self.screen.has_insurance_types(("chp",))
+        )
 
         if adjunctive_eligible:
             e.condition(True, messages.presumed_eligibility())
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -61,18 +61,21 @@ class TxFpp(ProgramCalculator):
         # §4140 adjunctive income eligibility: enrollment in SNAP, WIC, or CHIP
         # (applicant or their child) bypasses the 250% FPL income test.
         #
-        # SNAP/WIC enrollment lives in the CurrentBenefit join table, read via
-        # has_base_benefit() against the cross-white-label `base_program` grouping
-        # (tx_snap → "snap", tx_wic → "wic"). CHIP is collected as a per-member
-        # insurance question, not a current benefit, so it is read from member
-        # insurance (matches any household member — "applicant or their child").
+        # SNAP/WIC enrollment lives in the CurrentBenefit join table under their
+        # TX-scoped names (tx_snap, tx_wic — the "already have this" tiles flagged
+        # in migration 0142), read via has_benefit(). CHIP is NOT a current-benefits
+        # tile: tx_chip is a PolicyEngine eligibility program, never written to the
+        # join table, so has_benefit("tx_chip") is always False. CHIP enrollment is
+        # captured as a per-member insurance question, so it is read from member
+        # insurance (matches any household member — "applicant or their child"),
+        # mirroring the sibling Healthy Texas Women calculator.
         #
         # Do NOT read the legacy Screen.has_snap/has_wic/has_chp boolean columns:
         # the join-table migration (MFB-720) dropped them from the serializer write
         # contract, so they are never populated and are permanently False.
         adjunctive_eligible = (
-            self.screen.has_base_benefit("snap")
-            or self.screen.has_base_benefit("wic")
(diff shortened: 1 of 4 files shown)


programs/programs/tx/fpp/calculator.py
```diff
@@ -61,11 +61,23 @@ def household_eligible(self, e: Eligibility):
...
+        # tile: tx_chip is a PolicyEngine eligibility program, never written to the
+        # join table, so has_benefit("tx_chip") is always False. CHIP enrollment is
+        # captured as a per-member insurance question, so it is read from member
+        # insurance (matches any household member — "applicant or their child"),
+        # mirroring the sibling Healthy Texas Women calculator.
+        #
+        # Do NOT read the legacy Screen.has_snap/has_wic/has_chp boolean columns:
+        # the join-table migration (MFB-720) dropped them from the serializer write
+        # contract, so they are never populated and are permanently False.
+        adjunctive_eligible = (
```
catonph: we have historically used presumptive_eligibility as the var name it could be nice to stay consistent

programs/programs/tx/fpp/tests/test_fpp.py
```diff
@@ -43,14 +43,29 @@ def make_calculator(
...
+    mirror those methods rather than the legacy has_snap/has_wic/has_chp columns,
+    which the serializer no longer populates (MFB-720) — so a regression back to
+    reading those dead columns (or to has_benefit("tx_chip"), which is always
+    False) would fail these tests.
+    """
     mock_program = Mock()
     mock_program.year.get_limit.return_value = fpl_limit
 
+    current_benefits = set()
+    if has_snap:
```
catonph: im hesitant to use the same language here with has_snap....is there a different way for us to pass this into the calculator mock? e.g. pass `current_benefits` instead

catonph (approved):

PR #1594, opened 2026-06-17
Replace in-memory custom Cache class with django.core.cache (Redis)

SonaliBedge:
## Context & Motivation

- Fixes : [#mfb-468](https://linear.app/myfriendben/issue/MFB-468/general-replace-in-memory-custom-cache-with-redis)
- Related PR:[ PR-1260](https://github.com/MyFriendBen/benefits-api/pull/1260) , [PR-1227](https://github.com/MyFriendBen/benefits-api/pull/1227)

Fixes Heroku R14 memory quota errors caused by the custom Cache class storing data in each dyno's process memory. Promotes Redis to the default cache backend and migrates all 13 Cache subclasses to django.core.cache.

Heroku dynos were hitting memory quota errors (R14) caused by the custom in-memory `Cache` class storing data directly in each dyno's process memory:
We're currently hitting memory limits on Heroku with the current caching approach:

 ```
2025-11-18T16:46:52 heroku[web.1]: Process running mem=551M(102.6%)
 2025-11-18T16:46:52 heroku[web.1]: Error R14 (Memory quota exceeded)
```
**Root cause :** LocMemCache stores cached data inside each dyno's own RAM. When Heroku runs multiple dynos (on scale-up, restarts, or deploys), each dyno independently fetches and stores its own copy of the same data — translations, FPL tables, Google Sheets values, PolicyEngine tokens — multiplying memory usage across dynos.

Redis was already provisioned as a Heroku add-on but was only configured as a secondary named cache, not the default. This PR promotes Redis to the default cache backend and removes the custom in-memory Cache class entirely.

Benefits of this change:

- **Fixes R14** — cached data no longer lives in dyno RAM, A separate service (Redis) , shared by all dynos
- **Shared cache across dynos** — if Dyno A caches a PolicyEngine response, Dyno B reuses it instead of fetching its own copy means one copy, everyone reads it
- **Survives restarts** — LocMemCache is wiped on every deploy; Redis persists across restarts, so the first request after a deploy does not have to refetch everything
- **Scales correctly** — as Heroku scales to more dynos, memory usage no longer multiplies e.g. 5 dynos still share one cache.

## Changes Made

- **`benefits/settings.py`** - Promoted Redis to the `"default"` cache backend when `REDIS_URL` is set (previously it was a secondary named cache). Falls back to `LocMemCache` in local dev when `REDIS_URL` is absent. `PARLER_CACHE_BACKEND` now uses the `"default"` backend in both environments. Added `IGNORE_EXCEPTIONS: True` to the Redis `OPTIONS` so a Redis outage degrades to a cache miss (falls through to the DB/source API) instead of raising and crashing every page that now routes through Redis — including Parler translations, which were previously isolated on their own dedicated `LocMemCache` specifically to avoid this risk.

- **`integrations/util/cache.py`** - Deleted the custom in-memory `Cache` base class entirely.

- **`integrations/util/__init__.py`** - Deleted (only re-exported `Cache`).

- **`integrations/services/sheets/sheets.py`** - Removed `GoogleSheetsCache` (was a bridge between `Cache` and Google Sheets; no longer needed).

- **13 `Cache` subclasses across 11 files** migrated to `django.core.cache`: `FplCache`, `TranslationCache`, `PolicyEngineBearerTokenCache`, `CFHCache`, `GoogleSheetsCache`, `Ami`, `Smi`, `IncomeLimitsCache` (×2), `CccapFplCache`, `CoHeadStartCountyEligibleCache`, `LeapValueCache`, `RAGCache`, `BoulderAmiCache`, `ACACache`, `NcHeadStartMarketRatesCache`

- **`translations/views.py`** - Fixed a broken reference to the now-removed `translation_cache` attribute on `TranslationManager`.

### Addressed review feedback

- **`integrations/services/sheets/cache.py`** (new file) - Added a `GoogleSheetsCache` abstract base class to de-duplicate the near-identical `cache.get`/process/`cache.set` boilerplate that had been copy-pasted into 9 of the Sheets-backed cache classes above. Note: this reuses the same class name as the one removed from `sheets.py` above — it's an unrelated, newly-designed class (a shared base for the Redis-backed caches), not a reintroduction of the old bridge class.
  - Centralizes a fix for a bug present in nearly every one of those 9 classes: an empty/failed Sheets fetch was being cached for the full 24h `CACHE_TIMEOUT`, so a single bad fetch would break that program's screening for every user for a full day. Now an empty result is never cached (retries next request).
  - Adds a stale-data fallback with Sentry reporting (`capture_exception`) when a fetch or process call raises — replicating the old `Cache._update_cache()`'s graceful-degradation behavior (log to Sentry, keep serving last-known-good data) that had been dropped during the initial migration to `django.core.cache`.
  - Migrated `CccapFplCache`, `RAGCache`, `BoulderAmiCache`, `ACACache`, `NcHeadStartMarketRatesCache`, `CoHeadStartCountyEligibleCache`, `LeapValueCache`, `Ami`, `Smi`, and `IncomeLimitsCache` onto this base, removing their duplicated caching logic. `CFHCache`'s connect-for-health cache was converted from a module-level function + globals into `CfhCountyValuesCache`, also on this base.

- **Merged the two duplicate `IncomeLimitsCache` classes** into one. Deleted `programs/programs/co/energy_programs_shared/income_limits_cache.py` (confirmed unused elsewhere) and repointed `energy_programs_shared/income_validation.py` at the surviving `integrations/services/income_limits.py` singleton.

- **Renamed `_get_data()` → `get_data()`** on the base class and all ~13 call sites (10 production files, plus 10 mock targets in `nc_head_start/tests.py`) — it's the primary API every calculator calls externally, so the leading underscore was the wrong convention signal.

- **Bug fixes found during review:**
  - `nurturing_futures/calculator.py` (`BoulderAmiCache`) — `a.replace(",", "".replace("$", ""))` never actually stripped `$` (the inner call evaluates to `""` first), silently zeroing every NurturingFutures income limit on malformed sheet values. Fixed to `.replace(",", "").replace("$", "")`, plus a `[0] * 8` fallback so a sheet returning no rows can't raise `IndexError`.
  - `connect_for_health/calculator.py` and `nc_aca/calculator.py` (`member_value`) — both used unguarded dict indexing (`values[county]`), which raises `KeyError` and — since the eligibility loop in `screener/views.py` only catches `DependencyError` — crashes the *entire* household's screening if a county is missing from the sheet. Fixed to `values.get(county, 0)`.
  - `rental_assistance_grant/calculator.py` (`RAGCache`) — restored the per-household-size income-limit list logic (comma-stripping, malformed-row skipping) that was dropped during an intermediate migration step.

## Testing

- Migrations to run: None
- Environment variables/settings to add: None - `REDIS_URL` is already set by the Heroku Redis add-on
- Manual testing steps:
   - Run `python -m pytest`
   - Verify translations load correctly across languages 
   - Verify benefit program calculators (CCAP, LEAP, Connect for Health, PolicyEngine) return correct eligibility and values
   - Confirm Heroku dyno memory usage drops after deploy

Note: The R14 error cannot be reproduced locally — it only occurs under Heroku's 512MB dyno memory limit.

## Deployment

- Production config updates: None - Redis add-on is already provisioned
- On first deploy, all cache entries will be cold (expected). The first request per cache key will populate Redis - no data loss risk.
- Admin updates needed: None
- Notify team/users of: Expect a brief spike in Google Sheets API calls and PolicyEngine API calls on first deploy as caches warm up.

## Notes for Reviewers

- `PARLER_CACHE_BACKEND = "default"` means Parler translations now use Redis in production (previously use LocMemCache per-dyno). This is an improvement- translations are now shared and consistent across all dynos. 
- All Redis cache keys are namespaces user `KEY_PREFIX - "benefits"` in settings, preventing collisions with other services.
- energy_programs_shared/income_limits_cache.py uses cache key "energy_income_limits_data" while integrations/services/income_limits.py uses "income_limits_data" — different keys intentionally, to avoid cache collisions between the two classes.

## Summary by CodeRabbit

* **Caching & Performance**
  * Default caching now switches to Redis when configured; otherwise it uses in-memory caching.
  * Spreadsheet-backed program inputs (income limits, FPL/AMI, county eligibility/rates, and related thresholds) now share a consistent cache-first loading flow.

* **Reliability & Data Freshness**
  * Translations refresh/invalidate more reliably after changes and on error/retry scenarios.
  * Policy-engine bearer tokens use improved caching with automatic invalidation after authorization failures.
  * County-based calculations are more resilient to malformed or missing sheet rows (missing values fall back safely).

### CODE REVIEW

File: programs/programs/nc/nc_aca/calculator.py [hunk 1]
```diff
@@ -1,19 +1,40 @@
...
+            return data
+        data = self._process()
+        cache.set(self.CACHE_KEY, data, timeout=self.CACHE_TIMEOUT)
+        return data
 
-        return {d[0].strip() + " County": float(d[1].replace(",", "")) for d in data}
+    def _process(self):
+        data = GoogleSheets(self.sheet_id, self.range_name).data()
+
+        # return {d[0].strip() + " County": float(d[1].replace(",", "")) for d in data}
```
RomannaBidnyk: Looks like this might be a leftover comment from the original implementation — probably safe to remove?

programs/programs/co/nurturing_futures/calculator.py
```diff
@@ -1,18 +1,38 @@
...
+    def _process(self):
+        data = GoogleSheets(self.sheet_id, self.range_name).data()
+
+        if not data or len(data) == 0:
+            return []
+
+        result = []
+        for a in data[0]:
+            try:
+                cleaned_value = a.replace(",", "".replace("$", ""))
```
RomannaBidnyk: Was the intention here to strip both commas and dollar signs from the value? If yes, there might be a bug in the new code added in this PR:

`cleaned_value = a.replace(",", "".replace("$", ""))`

It looks like` "".replace("$", "") `evaluates to `""` first, so this might end up being the same as `a.replace(",", "") `— meaning the `$` never actually gets removed. If that's correct, a value like `"$120,000"` would become `"$120000"`, fail `int()`, and silently default to `0`, which could make every NurturingFutures income limit 0.

Probably worth double checking — if the goal was to strip both, should it be:

`cleaned_value = a.replace(",", "").replace("$", "")`

Please correct me if my understanding is wrong!

programs/programs/co/nurturing_futures/calculator.py
```diff
@@ -1,18 +1,38 @@
...
+        data = self._process()
+        cache.set(self.CACHE_KEY, data, timeout=self.CACHE_TIMEOUT)
+        return data
 
-        return [int(a.replace(",", "").replace("$", "")) for a in data[0]]
+    def _process(self):
+        data = GoogleSheets(self.sheet_id, self.range_name).data()
+
+        if not data or len(data) == 0:
+            return []
```
RomannaBidnyk, line 20: As far as I understand, if Google Sheets returns no rows, `_process()` returns an empty list `[]`, and that empty list gets stored in Redis for 24 hours with no check.

Then when a Boulder County household is screened, the code does `_get_data()[household_size - 1] `— indexing into whatever came back. If it's an empty list, that would raise `IndexError` (for example [][1] for a household of 2) on every NurturingFutures screening for the next 24 hours until the cache expires naturally.

Please correct me if I'm wrong, but it looks like the old code handled this by returning a safe fallback like [0, 0, 0, 0, 0, 0, 0, 0] when fetching failed, so at least it wouldn't crash. Should we add something similar here — like skipping the cache write if the result is empty?

programs/programs/co/connect_for_health/calculator.py
```diff
@@ -1,32 +1,44 @@
...
+    values = {}
+    for row in data:
+        try:
+            county_key = row[_CFH_COUNTY_COLUMN].strip() + " County"
+            premium_value = float(row[_CFH_AVERAGE_COLUMN])
+            values[county_key] = premium_value
+        except (KeyError, ValueError, AttributeError):
+            continue  # Skip malformed rows
+
+    cache.set(_CFH_CACHE_KEY, values, timeout=_CFH_CACHE_TIMEOUT)
```
RomannaBidnyk: As far as I understand, if Google Sheets returns no rows,` _get_cfh_county_values() `loops over nothing and returns an empty dict `{}`, which then gets stored in Redis for 24 hours with no guard.

Then in `member_value()` the code does `values[county]` — if `values` is an empty dict, that raises `KeyError` on every Connect for Health screening in Colorado until the cache expires.

Please correct me if I'm wrong — should we add a check before caching, like skipping the cache write if the result is empty?

benefits/settings.py
```diff
@@ -316,46 +316,43 @@
...
             },
-            "SOCKET_CONNECT_TIMEOUT": 5,  # seconds
-            "SOCKET_TIMEOUT": 5,  # seconds
         },
-        "KEY_PREFIX": "benefits",  # Namespace for all cache keys
-        "TIMEOUT": 300,  # Default timeout: 5 minutes
     }
 
-# Configure Parler to use the default (LocMemCache) backend
 PARLER_CACHE_BACKEND = "default"
```
RomannaBidnyk, line 358: Looking at the settings change — as far as I understand, in the old code `PARLER_CACHE_BACKEND = "default"` and `"default"` was `LocMemCache`, so Parler translations never touched Redis at all. There was a comment explaining: "This avoids Redis connection issues and keeps translations fast".
In the new code `"default"` becomes Redis in production, so Parler now hits Redis too.

If Redis is slow or temporarily unavailable, any page that needs to display translated text would wait 5 seconds (the socket timeout) and then crash — whereas before, translations always worked regardless of Redis state.

Please correct me if I'm wrong - it looks like the original isolation was intentional. Two possible fixes come to mind — either keep a separate LocMemCache backend just for Parler and point `PARLER_CACHE_BACKEND = "parler`" at it, or add `"IGNORE_EXCEPTIONS": True` to the Redis OPTIONS so cache failures degrade gracefully instead of crashing.

RomannaBidnyk (changes requested):
Great work on this PR, @SonaliBedge ! This is a significant refactor touching 21 files and 13 cache classes, and the overall approach is clean and well-structured. The read-through cache pattern is consistent across all the migrated classes, the settings change is well thought out with a proper local dev fallback, and the translation cache invalidation across all dynos which is a real improvement over the old design.

This was a great learning experience for me to review — I learned a lot about how caching works in Django, what Redis is and why it matters in a multi-dyno setup, and how the different parts of this codebase connect together. Thank you for the detailed PR description, it made it much easier to follow the reasoning behind the changes.

I left a few comments with questions and things that looked potentially off to me — please correct me if I misunderstood anything, I'm still learning!

programs/programs/co/nurturing_futures/calculator.py
SonaliBedge: Both issues are fixed : 

If `_process()` comes back empty (e.g. Google Sheets returns no rows), the empty result is no longer written to the cache — so the next screening retries the fetch instead of being stuck for 24 hours.
In the meantime, `_get_data()` returns `DEFAULT_AMI_LIMITS` (eight zeros, one per household size 1-8), so indexing with `household_size - 1` can't raise IndexError. An income limit of 0 just makes the household fail the income condition instead of crashing the whole screening — the same fail-safe behavior the old code had.

programs/programs/co/connect_for_health/calculator.py
SonaliBedge: Both fixed : 

[Lines 33-36](vscode-webview://1vmjck8fae2qiob77bjdj2c6l3foilgvdjau0921ummg1gtf6oj1/MFB_Repo/benefits-api/programs/programs/co/connect_for_health/calculator.py#L33-L36): an empty values dict is no longer cached, so the next screening retries the Sheets fetch instead of being stuck for 24 hours.
[Line 72](vscode-webview://1vmjck8fae2qiob77bjdj2c6l3foilgvdjau0921ummg1gtf6oj1/MFB_Repo/benefits-api/programs/programs/co/connect_for_health/calculator.py#L72): values.get(county, 0) instead of values[county] — a missing county now yields a $0 value rather than raising KeyError.

benefits/settings.py
SonaliBedge: Added "IGNORE_EXCEPTIONS": True to the Redis cache OPTIONS in [settings.py:328-338](vscode-webview://1vmjck8fae2qiob77bjdj2c6l3foilgvdjau0921ummg1gtf6oj1/MFB_Repo/benefits-api/benefits/settings.py#L328-L338). With this, django-redis catches connection/timeout errors internally and treats them as a cache miss (get() returns None, set() is a no-op) rather than propagating an exception.

Net effect for Parler and every other cache.get()/cache.set() call added in this PR (FPL, GoogleSheets-backed caches, PolicyEngine token cache, etc.): if Redis is down, requests still pay up to the 5s socket timeout, but they render/succeed by falling through to the DB or source API instead of crashing. That's a real cost worth flagging — it's slow, not free — but it converts a hard outage into a degraded-but-working state, and keeps translations on Redis rather than reintroducing the in-memory cache that was likely driving the original R14 memory errors.

SonaliBedge: Thank you, @RomannaBidnyk, for reviewing my PR. I've fixed all the issues. Could you please take another look?

RomannaBidnyk: Thanks for the explanation @SonaliBedge — this looks great to me. Since it swaps the default cache backend across the whole app, I'd like a second reviewer to take a look at this one as well

RomannaBidnyk (approved):
Really nice work on this, @SonaliBedge!
All my earlier comments have been addressed.
Approving from my side!

myang1010 (commented):
Hi Sonali, thank you for the thoughtful work on migrating the cache to Redis. And thank you Romanna for thorough review and helpful feedback! 

All the changes make sense to me. Just a few suggestions,
1.  Repeated `_get_data` pattern. Consider keep the `GoogleSheetsCache` class, then replace class body with `CACHE_KEY` and `CACHE_TIMEOUT` class vars, a `get_data()` and `_process()` abstract methods would eliminate this.

2. Multiple calculators call `*._get_data()`. The single underscore signals "private" by convention. These should be renamed to `get_data()`.  
 
3. KeyError risk in nc_aca. `nc_aca/calculator.py` uses `values[self.screen.county] * 12` (direct access), while the `connect_for_health` calculator was updated to use `values.get(county, 0)`. If the county isn't in the sheet data, this will crash.

4. Error handling. The old `Cache._update_cache()` caught all exceptions and reported them to Sentry while silently serving stale data. The new `_process()` methods let exceptions propagate to callers. Consider reuse the error handling part in `get_data()`.

Thanks and let me know your thoughts!

programs/programs/co/energy_assistance/calculator.py
```diff
@@ -1,21 +1,26 @@
...
 from programs.co_county_zips import counties_from_screen
 import math
 
 
 class LeapValueCache(GoogleSheetsCache):
-    expire_time = 60 * 60 * 24
-    default = []
+    CACHE_KEY = "leap_value_data"
     sheet_id = "1W8WbJsb5Mgb4CUkte2SCuDnqigqkmaO3LC0KSfhEdGg"
     range_name = "'FFY 2025'!A2:G65"
```
myang1010, line 13: `LeapValueCache._process()` returns a `[]` but the base class `_empty_fallback()` returns `{}`. How about adding a `_empty_fallback()` that returns `[]` here?

myang1010 (commented):
Hi Sonali, thanks for the updates! The new changes look good to me! Just a small thing listed below.

programs/programs/co/energy_assistance/calculator.py
SonaliBedge: Thank you, Ming, for reviewing it again, I made the changes above.

integrations/services/sheets/cache.py
```diff
@@ -0,0 +1,70 @@
...
+            data = self._process(self._fetch_raw())
+        except Exception as exc:
+            capture_exception(exc)
+            stale_data = cache.get(self._stale_cache_key)
+            return stale_data if stale_data is not None else self._empty_fallback()
+
+        if not data:
+            # don't cache empty data/falsy result - retry on the next request, return it directly
+            # instead of locking every dependent screening out for 24h.
+            return self._empty_fallback()
```
andreaHG: The exception path checks the stale cache before falling back to `_empty_fallback()`, but the `if not data` path goes directly to `_empty_fallback()`.

If a sheet temporarily returns no rows without raising an exception, `_process()` may produce an empty result and we would discard an otherwise usable last-known-good value. Would it make sense for both unsuccessful refresh paths to use the same stale-or-fallback behavior?

For example:
```py

    def _stale_or_fallback(self) -> T:
        """
        Serve the last known-good (stale) value if one is cached, otherwise fall back
        to `_empty_fallback()`. Used whenever a fresh fetch fails or comes back empty.
        """
        stale_data = cache.get(self._stale_cache_key)
        return stale_data if stale_data is not None else self._empty_fallback()

```

Then both paths could call: `return self._stale_or_fallback()`

integrations/services/sheets/cache.py
andreaHG, line 1: This is now shared fallback and failure-handling logic for roughly 10 program caches. Would it be worth adding direct unit tests for get_data() rather than relying on calculator tests to exercise it indirectly?

At minimum, it seems useful to cover:

- a successful fetch with non-empty data
- an empty result, with and without stale data
- an exception, with and without stale data
- a cache hit (returns the cached value without re-fetching)

andreaHG (commented):
Looks great! 

I left two small, non-blocking comments below: one about a gap in the stale-fallback behavior for the empty-result path, and one suggesting direct test coverage for the new base class. Otherwise, this looks solid to merge.

integrations/services/sheets/cache.py
SonaliBedge: Thank you, Andrea, for reviewing PR, I fixed stale-fallback gap and added unit tests get_data()

andreaHG (approved):

PR #1595, opened 2026-06-18
MFB-1036: Add Google Places autocomplete to Calculate Impact address field

david-j-stoll:
# MFB-1036 (Backend): Google Places Autocomplete proxy endpoint

## Context & Motivation

- Fixes [MFB-1036](https://linear.app/myfriendben/issue/MFB-1036/heat-pump-step-2b1-add-google-places-autocomplete-to-calculate-impact)
- Related PR: frontend changes in `frontend.md`

The Bill Impact Calculator on the Heat Pump Journey collects a home address to pass to the Rewiring America REM API. Rather than loading the Google Maps JS SDK in the browser (which would expose the API key in the client bundle), we proxy autocomplete requests through the Django backend — the same pattern used for the REM API proxy (`RemImpactView`). The API key lives only on the server.

## Changes Made

- **`integrations/clients/google_places.py`** (new) — `GooglePlacesClient` with a single `autocomplete_address(input_text)` method. Calls `maps.googleapis.com/maps/api/place/autocomplete/json` with `types=address` and `components=country:us`, strips the trailing `, USA` suffix from each prediction's description, and returns a `list[AddressPrediction]` (`TypedDict` with `description` and `place_id` string fields). Reads `settings.GOOGLE_MAPS_API_KEY`. Results are cached using Django's `default` cache backend with a 7-day TTL; the cache key is `places_autocomplete_<md5(input.lower())>` so repeated and case-variant queries skip the upstream API entirely. If Google's JSON response carries a status other than `OK` or `ZERO_RESULTS` (e.g. `REQUEST_DENIED` from a missing or misconfigured key), a `WARNING` is logged and a `requests.RequestException` is raised — the error response is not cached, and the caller surfaces a 502.

- **`benefits/settings.py`** — Added `GOOGLE_MAPS_API_KEY = config("GOOGLE_MAPS_API_KEY", default="")`. Added `"places": "200/hour"` to `DEFAULT_THROTTLE_RATES` (throttles per hashed IP, matching the existing `rem` throttle pattern).

- **`screener/views.py`** — Extracted `HashedIPAnonRateThrottle` base class with the shared `get_cache_key` logic (hashes the client IP via SHA-256 so raw IPs are never stored in the cache). `RemRateThrottle`, `NPSRateThrottle`, and `PlacesRateThrottle` now each subclass it with just a `scope` declaration. Added `PlacesAutocompleteView` (`AllowAny`, returns empty list on blank input, 502 on upstream HTTP or network errors). The `HTTPError` handler safely reads `e.response.status_code` only when `e.response is not None`, falling back to `"unknown"` to avoid `AttributeError` in edge cases where the exception carries no response object. Imported `GooglePlacesClient`.

- **`screener/urls.py`** — Registered `places/autocomplete/` at `api/places/autocomplete/`. No white-label segment — this endpoint is not tenant-specific.

- **`benefits-api/.env`** — Added `GOOGLE_MAPS_API_KEY=` as an empty placeholder so other developers know the variable is required.

- **`screener/tests/test_places_autocomplete.py`** (new) — 16 tests covering `GooglePlacesClient` (param passing, USA suffix stripping, empty predictions, already-clean descriptions, cache hit skips API call, cache key is case-insensitive, error status raises and is not cached) and `PlacesAutocompleteView` (missing/blank input → 200 empty list, Google API error status → 502, HTTP error → 502, network error → 502, happy path, input trimming).

## Testing

- **Migrations to run:** None.
- **Configuration updates needed:** Set `GOOGLE_MAPS_API_KEY` locally in `benefits-api/.env` with a valid Google Places API key. See the API key note below before provisioning.
- **Environment variables/settings to add:** `GOOGLE_MAPS_API_KEY` in `benefits-api/.env` (local) and as a Heroku config var for staging/production.
- **Manual testing steps:**
  1. Add a valid `GOOGLE_MAPS_API_KEY` to `benefits-api/.env`.
  2. Start the Django dev server.
  3. `curl "http://localhost:8000/api/places/autocomplete/?input=123+Main"` — confirm a JSON array of `{description, place_id}` objects is returned.
  4. Confirm descriptions do not include `, USA`.
  5. `curl "http://localhost:8000/api/places/autocomplete/"` (no input) — confirm `[]` is returned.
  6. Repeat step 3 — the second call returns from cache without hitting the Places API (confirm by checking server logs for outbound requests).
  7. Run `python manage.py test screener.tests.test_places_autocomplete --no-input` — all 16 tests should pass.

## Deployment

- **Post-deployment scripts needed:** None.
- **Production config updates:** Add `GOOGLE_MAPS_API_KEY` as a Heroku config var on both staging and production. This is required before the frontend autocomplete feature will function — the frontend will still render (with a non-functional address field) if the key is absent, but API calls will return 502.
- **Admin updates needed:** None.
- **Notify team/users of:** Notify Caton that the API key must be provisioned with **no HTTP referrer restrictions** (referrer restrictions are enforced by checking the browser `Referer` header, which server-to-server requests do not send — Google will return `REQUEST_DENIED` for a referrer-restricted key). The correct setup is either unrestricted or IP-restricted, with an **API restriction** scoped to the Places API in Google Cloud Console.

## Notes for Reviewers

- **Caching backend:** The cache uses Django's `default` backend — `LocMemCache` (per-process, in-memory) in local dev, and also in production unless `REDIS_URL` is set. With multiple Heroku dynos each dyno maintains its own warm cache independently, which still eliminates repeat calls within a worker. The `"redis"` backend (cross-dyno) can be wired in later if needed by swapping `cache` for `caches["redis"]` in `google_places.py`.
- **Known limitations:** `AnonRateThrottle` throttles by IP, so users behind a shared IP (office, school) share the `200/hour` limit. This is consistent with how `RemRateThrottle` works and acceptable at current scale.
- **Future considerations:** If USPS delivery-point validation (`dpvConfirmation`) is required, the Google Address Validation API can be added as a second endpoint. Because the autocomplete session would then be linked to a validation request, the autocomplete portion would become free per Google's billing rules.

### DIFF
--- a/screener/views.py
+++ b/screener/views.py
@@ -4,6 +4,7 @@ from typing import Optional
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 from integrations.clients.rewiring_america import RewiringAmericaClient
+from integrations.clients.google_places import GooglePlacesClient
 from integrations.services.communications import MessageUser
 from programs.programs.policyengine import versions as pe_versions
 from programs.models import Referrer
@@ -702,6 +703,23 @@ class NPSRateThrottle(throttling.AnonRateThrottle):
         return self.cache_format % {"scope": self.scope, "ident": hashed}
 
 
+class PlacesRateThrottle(throttling.AnonRateThrottle):
+    """
+    Rate throttle for Google Places Autocomplete proxy requests.
+    Rate is configured via DEFAULT_THROTTLE_RATES["places"] in settings.
+    Uses a hashed IP as the cache key to avoid storing raw IPs.
+    """
+
+    scope = "places"
+
+    def get_cache_key(self, request, view):
+        ident = self.get_ident(request)
+        if ident is None:
+            return None
+        hashed = hashlib.sha256(ident.encode()).hexdigest()
+        return self.cache_format % {"scope": self.scope, "ident": hashed}
+
+
 class NPSScoreView(views.APIView):
     """
     API endpoint for submitting NPS (Net Promoter Score) feedback.
@@ -862,3 +880,36 @@ class RemImpactView(views.APIView):
 
         serializer = RemImpactSerializer(raw)
         return Response(serializer.data)
+
+
+class PlacesAutocompleteView(views.APIView):
+    """
+    Proxies address autocomplete requests to the Google Places API,
+    keeping the API key server-side. Returns US street address predictions
+    restricted to street-level addresses.
+    """
+
+    permission_classes = [permissions.AllowAny]
+    throttle_classes = [PlacesRateThrottle]
+
+    def get(self, request, **_kwargs) -> Response:
+        input_text = request.query_params.get("input", "").strip()
+
+        if not input_text:
+            return Response([], status=status.HTTP_200_OK)
+
+        try:
+            client = GooglePlacesClient()
+            predictions = client.autocomplete_address(input_text)
+        except requests.HTTPError as e:
+            return Response(
+                {"error": f"Google Places API error: {e.response.status_code}"},
+                status=status.HTTP_502_BAD_GATEWAY,
+            )
+        except requests.RequestException as e:
+            return Response(
+                {"error": "Google Places request failed.", "detail": str(e)},
+                status=status.HTTP_502_BAD_GATEWAY,
+            )
+
+        return Response(predictions)
--- a/screener/views.py
+++ b/screener/views.py
@@ -902,8 +902,9 @@ class PlacesAutocompleteView(views.APIView):
             client = GooglePlacesClient()
             predictions = client.autocomplete_address(input_text)
         except requests.HTTPError as e:
+            http_status = e.response.status_code if e.response is not None else "unknown"
             return Response(
-                {"error": f"Google Places API error: {e.response.status_code}"},
+                {"error": f"Google Places API error: {http_status}"},
                 status=status.HTTP_502_BAD_GATEWAY,
             )
         except requests.RequestException as e:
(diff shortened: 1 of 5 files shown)


screener/views.py
```diff
@@ -702,6 +703,23 @@ def get_cache_key(self, request, view):
...
+class PlacesRateThrottle(throttling.AnonRateThrottle):
+    """
+    Rate throttle for Google Places Autocomplete proxy requests.
+    Rate is configured via DEFAULT_THROTTLE_RATES["places"] in settings.
+    Uses a hashed IP as the cache key to avoid storing raw IPs.
+    """
+
+    scope = "places"
+
+    def get_cache_key(self, request, view):
```
catonph: `get_cache_key()` here is now byte-for-byte identical to `RemRateThrottle.get_cache_key` and `NPSRateThrottle.get_cache_key`. This is the third copy. Could we extract a shared base, e.g.:

```
class HashedIPAnonRateThrottle(throttling.AnonRateThrottle):
    """AnonRateThrottle that keys on a hashed IP so raw IPs aren't stored."""
    def get_cache_key(self, request, view):
        ident = self.get_ident(request)
        if ident is None:
            return None
        hashed = hashlib.sha256(ident.encode()).hexdigest()
        return self.cache_format % {"scope": self.scope, "ident": hashed}
```

and have PlacesRateThrottle/RemRateThrottle/NPSRateThrottle subclass it with just scope = "...".

catonph (approved):
Approving but probably worth extracting the get_cache_key code

PR #1611, opened 2026-06-30
MFB-1094: Single-benefit toggle endpoint for results page benefits-api

david-j-stoll:
## Context & Motivation

Adds a lightweight endpoint for adding/removing a single `CurrentBenefit` row on a
screen, to be consumed by the results-page "already have this" toggle (MFB-871).

Reusing the Screen PATCH path for this would revalidate and rewrite the entire
screen (household members, expenses, income streams, etc.) on every toggle — a lot
of overhead and surface area for what is a single-row operation. This dedicated
endpoint writes one row.

- Implements: MFB-1094 (CB Step 7b — Single-benefit toggle endpoint for results page)
- Unblocks: MFB-871 (CB Step 8 — Results page "already have this" toggle)
- Related PR: N/A

## Changes Made

- **`screener/urls.py`** — New route `PATCH /api/screens/<uuid>/current-benefits/`
  → `ScreenCurrentBenefitsView`.
- **`screener/views.py`** — New `ScreenCurrentBenefitsView` (APIView):
  - Resolves `program` by `name_abbreviated` scoped to the screen's `white_label`;
    an unknown name (or one offered only by another white label) returns 404, so a
    toggle can't write a cross-WL program.
  - Idempotent add (`get_or_create`) / remove (`filter().delete()`) of the single
    `CurrentBenefit` row.
  - Read-modify-write wrapped in a transaction with `select_for_update()` on the
    screen, so concurrent toggles (or a toggle racing a Screen PATCH) can't
    interleave on the join table.
  - Returns the updated list: `{"current_benefits": [...sorted name_abbreviated...]}`.
  - Permissions: `DjangoModelPermissions` (PATCH → `screener.change_screen`),
    matching the Screen PATCH path.
- **`screener/serializers.py`** —
  - New `CurrentBenefitToggleSerializer` validating `name_abbreviated` (string) and `has` (bool).
  - Aligned the `name_abbreviated` length caps with the DB column. Both the new
    `CurrentBenefitToggleSerializer.name_abbreviated` field and the existing bulk-path
    `ScreenSerializer.current_benefits` list child were capped at 64, while the
    `Program.name_abbreviated` column is `max_length=120` — so a DB-valid name of
    65–120 chars would have been rejected with a 400 before the white-label-scoped
    resolve. Both caps are now `120` to match the column. (The bulk-path 64 was a
    pre-existing inconsistency, not introduced by this PR; fixed here while in the area.)
- **`screener/tests/test_current_benefits_toggle.py`** — 14 tests:
  - Behavior: add, remove, idempotent repeat-add / repeat-remove, wrong-WL 404,
    unknown-program 404, unknown-screen 404, missing-field 400s, and the
    unauthenticated / missing-perm rejection cases.
  - Length cap: the `name_abbreviated` cap equals the `Program.name_abbreviated` column
    (drift guard), a name at the column max clears validation (→ 404, not a length
    400), and a name over the max is rejected (400).
- **`screener/tests/test_screen_serializer.py`** — new `CurrentBenefitsNameLengthTests`
  covering the bulk path's `current_benefits` child cap: cap equals the column
  (drift guard), a name at the column max passes validation, a name over it is rejected.
  The drift-guard tests read `Program._meta.get_field("name_abbreviated").max_length`
  rather than hardcoding 120, so the serializer caps and the column can't silently diverge again.

Note on the API path: the ticket drafted the route as `/api/v2/screens/...`, but
there is no `v2` namespace anywhere in the codebase and nothing (BE or FE) consumes
it yet. The route is mounted at `/api/screens/<uuid>/current-benefits/` to stay
consistent with the existing `/api/screens/...` endpoints.

## Testing

- Migrations to run: none (no schema changes)
- Manual testing steps:
  1. `python manage.py test screener.tests.test_current_benefits_toggle screener.tests.test_screen_serializer`
  2. Against a running server, authenticated as a user with `screener.change_screen`:
     - Add: `PATCH /api/screens/<uuid>/current-benefits/` body `{"name_abbreviated": "snap", "has": true}`
       → 200, response `{"current_benefits": ["snap", ...]}`, row present.
     - Remove: same URL, body `{"name_abbreviated": "snap", "has": false}`
       → 200, `snap` absent from the returned list.
     - Repeat either call → idempotent (no error, no duplicate row).
     - Unknown program name or a program from another white label → 404.

## Deployment

- Admin updates needed: none
- Notify team/users of: nothing required. The endpoint has no consumer until
  MFB-871 (results-page toggle) ships — note that ticket's spec references the
  `/api/v2/...` path and should be updated to `/api/screens/<uuid>/current-benefits/`.

## Notes for Reviewers

- Known limitations:
  - Unlike the bulk write path (`_write_current_benefits`), this endpoint does NOT
    OR in derived benefits (e.g. SSI implied by an sSI income stream) — it's a
    targeted single-row operation. The next full Screen PATCH reapplies those
    compound rules. This is intentional per the ticket (single-row add/remove).
- Behavioral difference from the bulk path worth a look:
  - An unknown / cross-WL `program` here returns **404** (hard fail), whereas the
    bulk `_write_current_benefits()` path **silently drops** unrecognized names and
    logs them to Sentry. Intentional — a one-shot toggle on an unresolvable program
    is unambiguously a client error, so the FE (MFB-871) should treat 404 as "bad
    program key", not retry.
- Scope note:
  - This PR also bumps the bulk-path `ScreenSerializer.current_benefits` child cap
    from 64 → 120 (see Changes Made). That touches the existing screener PATCH
    validation, not just the new endpoint — flagging so it isn't a surprise in the diff.
- Future considerations:
  - If the results page ever needs to toggle multiple benefits at once, this could
    be extended to accept a list, but the current single-row contract matches the
    MFB-871 UI (one toggle per card).

## Summary by CodeRabbit

* **New Features**
  * Added an “already have this” toggle for screen results, allowing users to add or remove a single current benefit via a screen PATCH action.
  * Screen results now update instantly after each change.

* **Bug Fixes**
  * Improved validation and error handling for missing/invalid inputs, unknown benefit names, and incorrect scoping; prevents duplicate entries and keeps repeated add/remove actions idempotent.

* **Tests**
  * Added coverage for toggle behavior, permissions, validation boundaries, and failure cases.

### DIFF
--- a/screener/serializers.py
+++ b/screener/serializers.py
@@ -242,9 +242,11 @@ class ScreenSerializer(serializers.ModelSerializer):
     # max_length bounds an otherwise-unbounded payload: there are ~130 distinct
     # name_abbreviated values across all white labels, so 256 is a comfortable
     # ceiling. Unknown names are dropped in _write_current_benefits(), so element
-    # content needn't be validated here.
+    # content needn't be validated here. The per-element cap mirrors the
+    # Program.name_abbreviated column (max_length=120) so no DB-valid name is
+    # rejected here.
     current_benefits = serializers.ListField(
-        child=serializers.CharField(max_length=64),
+        child=serializers.CharField(max_length=120),
         required=False,
         write_only=True,
         max_length=256,
@@ -426,9 +428,12 @@ class CurrentBenefitToggleSerializer(serializers.Serializer):
     `program` is a `name_abbreviated`; `has` selects add (True) vs remove (False).
     Resolving the name to a Program and writing the join-table row happen in the
     view, which has the screen (and thus its white_label) in hand.
+
+    `program`'s max_length mirrors the Program.name_abbreviated column (120) so no
+    DB-valid name is rejected before the white-label-scoped lookup in the view.
     """
 
-    program = serializers.CharField(max_length=64)
+    program = serializers.CharField(max_length=120)
     has = serializers.BooleanField()
 
 
--- a/screener/serializers.py
+++ b/screener/serializers.py
@@ -420,6 +420,18 @@ class ScreenSerializer(serializers.ModelSerializer):
         return instance
 
 
+class CurrentBenefitToggleSerializer(serializers.Serializer):
+    """Input for the single-benefit toggle endpoint (PATCH .../current-benefits/).
+
+    `program` is a `name_abbreviated`; `has` selects add (True) vs remove (False).
+    Resolving the name to a Program and writing the join-table row happen in the
+    view, which has the screen (and thus its white_label) in hand.
+    """
+
+    program = serializers.CharField(max_length=64)
+    has = serializers.BooleanField()
+
+
 class NavigatorSerializer(serializers.Serializer):
     name = TranslationSerializer()
     phone_number = serializers.CharField()
--- a/screener/serializers.py
+++ b/screener/serializers.py
@@ -423,17 +423,17 @@ class ScreenSerializer(serializers.ModelSerializer):
 
 
 class CurrentBenefitToggleSerializer(serializers.Serializer):
-    """Input for the single-benefit toggle endpoint (PATCH .../current-benefits/).
+    """Input for the single-benefit toggle endpoint (PATCH /screens/<uuid:screen_uuid>/current-benefits/).
 
-    `program` is a `name_abbreviated`; `has` selects add (True) vs remove (False).
+    `name_abbreviated` is a program name_abbreviated value; `has` selects add (True) vs remove (False).
     Resolving the name to a Program and writing the join-table row happen in the
     view, which has the screen (and thus its white_label) in hand.
 
-    `program`'s max_length mirrors the Program.name_abbreviated column (120) so no
+    `name_abbreviated`'s max_length mirrors the Program.name_abbreviated column (120) so no
     DB-valid name is rejected before the white-label-scoped lookup in the view.
     """
 
-    program = serializers.CharField(max_length=120)
+    name_abbreviated = serializers.CharField(max_length=120)
     has = serializers.BooleanField()
 
 
--- a/screener/tests/test_current_benefits_toggle.py
+++ b/screener/tests/test_current_benefits_toggle.py
@@ -2,7 +2,7 @@
 Tests for the single-benefit toggle endpoint:
 
     PATCH /api/screens/<uuid>/current-benefits/
-    body: {"program": "snap", "has": true|false}
+    body: {"name_abbreviated": "snap", "has": true|false}
 
 A lightweight add/remove of one CurrentBenefit row consumed by the results-page
 "already have this" toggle. Resolution is white-label-scoped and the
@@ -19,7 +19,7 @@ from screener.models import CurrentBenefit, Screen, WhiteLabel
 from screener.serializers import CurrentBenefitToggleSerializer
 from screener.tests.helpers import seed_program
 
-# Program.name_abbreviated column max_length — the `program` field cap must match it.
+# Program.name_abbreviated column max_length — the `name_abbreviated` field cap must match it.
 NAME_ABBREVIATED_MAX_LENGTH = Program._meta.get_field("name_abbreviated").max_length
 
 
@@ -48,7 +48,7 @@ class CurrentBenefitsToggleTests(APITestCase):
 
     def test_add_new_benefit(self):
         """has=true creates the row and the response reflects the updated list."""
-        response = self.client.patch(self.url, {"program": "snap", "has": True}, format="json")
+        response = self.client.patch(self.url, {"name_abbreviated": "snap", "has": True}, format="json")
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data, {"current_benefits": ["snap"]})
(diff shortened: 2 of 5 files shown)


screener/serializers.py
```diff
@@ -420,6 +420,18 @@ def update(self, instance, validated_data):
...
 
+class CurrentBenefitToggleSerializer(serializers.Serializer):
+    """Input for the single-benefit toggle endpoint (PATCH .../current-benefits/).
+
+    `program` is a `name_abbreviated`; `has` selects add (True) vs remove (False).
+    Resolving the name to a Program and writing the join-table row happen in the
+    view, which has the screen (and thus its white_label) in hand.
+    """
+
+    program = serializers.CharField(max_length=64)
```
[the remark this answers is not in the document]
david-j-stoll: This was resolved in the most recent commit

screener/urls.py
```diff
@@ -13,6 +13,12 @@
     path("api-auth/", include("rest_framework.urls", namespace="rest_framework")),
     path("eligibility/<id>", views.EligibilityTranslationView.as_view(), name="translated screen eligibility endpoint"),
     path("screens/<uuid:screen_uuid>/nps/", views.NPSScoreView.as_view(), name="nps-score"),
+    # Single-benefit toggle for the results-page "already have this" control (MFB-1094 / MFB-871).
```
catonph: Let's remove ticket numbers here

screener/serializers.py
```diff
@@ -420,6 +422,21 @@ def update(self, instance, validated_data):
         return instance
 
 
+class CurrentBenefitToggleSerializer(serializers.Serializer):
+    """Input for the single-benefit toggle endpoint (PATCH .../current-benefits/).
```
catonph: Let's write out the full URL to avoid confusion here

screener/serializers.py
```diff
@@ -420,6 +422,21 @@ def update(self, instance, validated_data):
...
+
+    `program` is a `name_abbreviated`; `has` selects add (True) vs remove (False).
+    Resolving the name to a Program and writing the join-table row happen in the
+    view, which has the screen (and thus its white_label) in hand.
+
+    `program`'s max_length mirrors the Program.name_abbreviated column (120) so no
+    DB-valid name is rejected before the white-label-scoped lookup in the view.
+    """
+
+    program = serializers.CharField(max_length=120)
```
catonph: I think calling this name_abbreviated makes more sense as I look at the other serializers and that this points to a string and not a Program instance

screener/tests/test_current_benefits_toggle.py
```diff
@@ -0,0 +1,163 @@
+"""
+Tests for the single-benefit toggle endpoint:
+
+    PATCH /api/screens/<uuid>/current-benefits/
+    body: {"program": "snap", "has": true|false}
+
+A lightweight add/remove of one CurrentBenefit row consumed by the results-page
+"already have this" toggle (MFB-871). Resolution is white-label-scoped and the
```
catonph: Remove ticket number here

screener/tests/test_screen_serializer.py
```diff
@@ -9,10 +9,16 @@
...
 
+from programs.models import Program
 from screener.models import CurrentBenefit, HouseholdMember, IncomeStream, Screen, WhiteLabel
 from screener.serializers import ScreenSerializer, _write_current_benefits
 from screener.tests.helpers import seed_program
 
+# The Program.name_abbreviated column max_length. The serializer caps on a
+# current-benefit name must match this so no DB-valid name is rejected before the
+# white-label-scoped resolve. Read from the model so the tests track the column.
+NAME_ABBREVIATED_MAX_LENGTH = Program._meta.get_field("name_abbreviated").max_length
```
catonph: Do we need to look this up every time? Can we just set a reasonable max

screener/views.py
```diff
@@ -103,6 +106,60 @@ def update(self, request, pk=None):
...
+
+    # Same gate as the Screen PATCH path: DjangoModelPermissions maps PATCH to
+    # `screener.change_screen`, so whoever may edit the screen may toggle its
+    # benefits. The queryset is required for DjangoModelPermissions to resolve
+    # the model the permission is checked against.
+    permission_classes = [permissions.DjangoModelPermissions]
+    queryset = Screen.objects.all()
+
+    def patch(self, request, screen_uuid):
+        screen = get_object_or_404(Screen, uuid=screen_uuid)
```
catonph: This means we are fetching the screen row twice (once here and once on line 151)

We can collapse it to a single locked fetch by passing a select_for_update() queryset to get_object_or_404 inside the transaction:

```
with transaction.atomic():
    screen = get_object_or_404(Screen.objects.select_for_update(), uuid=screen_uuid)
    program = get_object_or_404(Program, white_label=screen.white_label, name_abbreviated=program_name)
    if has:
        CurrentBenefit.objects.get_or_create(screen=screen, program=program)
    else:
        CurrentBenefit.objects.filter(screen=screen, program=program).delete()
    current_benefits = sorted(
        CurrentBenefit.objects.filter(screen=screen).values_list("program__name_abbreviated", flat=True)
    )
```
get_object_or_404 accepts a QuerySet as its first arg, so this gets the lock + the 404 in one query. Bonus: the Program lookup now resolves inside the lock instead of against a pre-lock read of screen.white_label. The screen-missing 404 just rolls back an empty transaction — no behavioral change for the client.

Note: the bulk path _write_current_benefits uses the same two-step pattern, so consolidating here diverges from it slightly. Fine to leave that one alone given this endpoint is perf-motivated, but flagging for consistency.

catonph (changes requested):
A few things to take a look at

screener/tests/test_screen_serializer.py
david-j-stoll: As this is a unit test, I agree. Setting it at 120

screener/views.py
david-j-stoll: Good call, making this change!

catonph (approved):
LGTM!

PR #1631, opened 2026-07-14
MFB-1307: Improve benefit value for WA HCV program

david-j-stoll:
## Context & Motivation

- Linear ticket: [MFB-1307 — Improve benefit value for WA HCV program](https://linear.app/myfriendben/issue/MFB-1307/improve-benefit-value-for-wa-hcv-program)
- Related PR: [#1625 — MFB-1121: TX Section 8 Housing](https://github.com/MyFriendBen/benefits-api/pull/1625)

The WA HCV calculator (`programs/programs/wa/hcv/calculator.py`) shipped **before** the HUD client gained ZIP-level Small Area FMR (SAFMR) support, so its `household_value()` still calls `hud_client.get_screen_fmr()`, which always returns the **metro-wide** Fair Market Rent. Seattle-Bellevue is a HUD **mandatory-SAFMR metro** — the correct HCV payment standard there is the ZIP-level SAFMR, not the metro FMR. Since most WA HCV applicants (and most of the program's test scenarios) fall in King/Snohomish County ZIPs inside the Seattle-Bellevue HMFA, the current metro-only lookup materially misestimates the subsidy for them.

PR #1625 already added `hud_client.get_screen_payment_standard()` and seeded `Seattle-Bellevue, WA HUD Metro FMR Area` into `MANDATORY_SAFMR_AREA_NAMES` **specifically so WA HCV could reuse it**. This PR makes WA HCV consume that capability, back-ports two refinements from the newer TX HCV calculator, and — discovered during QA — corrects how the WA/TX HCV and WA WAP calculators handle "already receiving Section 8."

### Benefit value (WA HCV)
- **Payment standard now honors mandatory SAFMR** (`programs/programs/wa/hcv/calculator.py`): `household_value()` calls `hud_client.get_screen_payment_standard(...)` instead of `get_screen_fmr(...)`. Returns ZIP-level SAFMR for Seattle-Bellevue HMFA households; falls back to metro/county FMR elsewhere and when a ZIP is missing from HUD's SAFMR table. `zipcode` is already in `dependencies`.
- **Gross rent capped at reported rent** (24 CFR § 982.505 "lower of Payment Standard or Gross Rent"): `gross_rent = min(payment_standard, reported_rent)` when reported, else the payment standard alone.
- **Resilience back-port from `TxHcv`:** both `household_eligible` and `household_value` catch unexpected (non-`HudIncomeClientError`) exceptions, `logger.exception(...)` them with context, and degrade to *not eligible* / *$0* — an unexpected error can no longer 500 the eligibility run.

### Section 8 duplicate-subsidy handling (WA/TX)

Background: the old `has_benefit("section_8")` check **never matched** — "Section 8" is stored per white label with `base_program = "section_8"` (`wa_hcv`, `tx_hcv`, `co_section_8`, `ma_section_8`, …), so the bare `name_abbreviated` `"section_8"` exists nowhere and the check was silently always-False.

We're adopting the workflow rule that **a calculator must not check whether the household already has its own program** — that duplicate case is handled generically by the results-layer `already_has` filter (`screen.has_benefit(program.name_abbreviated)` in `screener/views.py`), which removes a program from results when the household reports it. Calculators only check *other* programs to inform eligibility.

- **`wa_hcv` and `tx_hcv`:** these programs **are** the `section_8` program in their white labels, so an "already receiving Section 8" gate is a self-check. **Removed it entirely** — the duplicate-voucher case is now handled by the generic `already_has` results-layer filter, not the calculator. (An otherwise-eligible household that reports Section 8 now computes as *eligible* in the calculator and is filtered from results by `already_has`.)
- **`wa_wap`:** here Section 8 is a *different* program that informs weatherization eligibility (a categorical-eligibility bypass), so the check stays. Changed the concrete `has_benefit("wa_hcv")` → `has_base_benefit("section_8")` so it matches every white-label variant.

### Tests & specs
- `programs/programs/wa/hcv/tests/test_wa_hcv.py`: `patch_hud_client` now patches `get_screen_payment_standard`; `make_calculator` gained `reported_rent` and dropped `has_section_8`; added reported-rent-cap and unexpected-exception tests; replaced the Section 8 exclusion test with one asserting the calculator does **not** self-exclude (duplicate case left to `already_has`).
- `programs/programs/tx/hcv/tests/test_tx_hcv.py`: dropped `has_section_8`; the former "Section 8 → ineligible" tests (incl. scenarios 10/11) now assert the calculator returns **eligible**, with the duplicate case delegated to `already_has`.
- `programs/programs/wa/wap/tests/test_wa_wap.py`: rewired Section 8 mock to `has_base_benefit("section_8")` + regression test for the categorical pathway.
- `programs/programs/wa/hcv/spec.md`: Criterion 4 and Scenario 7 now describe the duplicate case as handled by the `already_has` results-layer filter (not a calculator check); **corrected Scenario 4** income to $75,000/yr (its "just above" income of $57,504 was below the live FY2026 3-person Seattle-Bellevue VLI of $74,000, so it wrongly passed as Eligible — spec-accuracy fix, no calculator change).
- `programs/programs/tx/hcv/spec.md`: Criterion 2 and Scenarios 10/11 updated the same way (duplicate case → `already_has`, not a calculator self-check).
- `programs/programs/wa/wap/spec.md`: Section 8 categorical reference updated to `has_base_benefit("section_8")`.

## Testing

- Migrations / config / env vars: none (`HUD_API_TOKEN` already required)
- `pytest programs/programs/wa/hcv/ programs/programs/tx/hcv/ programs/programs/wa/wap/` → 140 passed
- **API QA (local, all 13 WA HCV spec scenarios): 13/13 pass** after the Scenario 4 spec correction. Pregnancy 2-person rule and SAFMR-based values verified; the Section 8 scenario is now verified as "calculator eligible, filtered by `already_has`."
- **Manual per-white-label check — go through each impacted white label's screener (WA and TX) and confirm behavior when the Section 8 / HCV benefit is selected on the "benefits you already have" step:**
  - **WA → `wa_hcv`:** select Housing Choice Voucher → `wa_hcv` is filtered from results by `already_has`; with it unselected, an otherwise-eligible household qualifies.
  - **WA → `wa_wap`:** select Housing Choice Voucher → `wa_wap` becomes categorically eligible even above 200% FPL (Section 8 → WA categorical pathway).
  - **TX → `tx_hcv`:** select Housing Choice Voucher → `tx_hcv` is filtered from results by `already_has`. (Depends on the config change in Deployment — the option must be selectable first.)
  - For each, confirm the inverse too: with the benefit **not** selected, the program is shown / the categorical branch does not fire.

## Deployment

- No migrations or code-side config. Notify team: WA HCV estimated values will shift (toward more accurate, ZIP-specific amounts) for Seattle-Bellevue households; the WA WAP Section 8 categorical pathway now actually takes effect.
- **Admin config to verify — every Section 8 / HCV program must be selectable on the "has current benefits" step (`show_in_has_benefits_step = True`), or the household can't report Section 8 and neither the `already_has` filter (WA/TX HCV) nor the `wa_wap` categorical pathway can receive real user input:**
  - **WA `wa_hcv`** — currently `True` ✅ (no change needed) — backs the `already_has` filter.
  - **WA `wa_wap`** — currently `True` ✅ (backs its own "already weatherized" self-exclusion input).
  - **TX `tx_hcv`** — currently **`False` ⚠️ must be flipped to `True`** (Django admin → Programs → `tx_hcv` → "Show in has benefits step"); otherwise TX users can't declare Section 8 receipt and `already_has` can never filter out a duplicate `tx_hcv`.
  - Confirm the same flag for any other white label whose HCV / Section 8 program should be filterable via `already_has`.

## Acceptance Criteria

1. **SAFMR routing** — Seattle-Bellevue HMFA households use the ZIP-level SAFMR payment standard, not metro FMR.
2. **Non-SAFMR fallback** — WA counties outside mandatory-SAFMR metros (e.g. Walla Walla) use metro/county FMR; value unchanged.
3. **Missing-ZIP fallback** — a ZIP absent from HUD's SAFMR table falls back to metro FMR, no error.
4. **Gross-rent cap** — reported rent below the payment standard caps HAP; no reported rent → payment standard.
5. **HAP floor & annualization unchanged** — floored at $0, returned as `HAP × 12`; TTP logic unchanged.
6. **Eligibility resilience** — unexpected exception in `household_eligible` → *not eligible* + `income_limit_unknown()`, logged, never 500s.
7. **Value resilience** — unexpected exception in `household_value` → `$0`, logged; `HudIncomeClientError` still → `$0` quietly.
8. **No HCV self-check** — `wa_hcv`/`tx_hcv` do not gate on already having Section 8; the duplicate case is handled by the `already_has` results-layer filter. `wa_wap`'s Section 8 categorical pathway fires via `has_base_benefit("section_8")`.
9. **Tests pass** — the three suites pass (140), incl. new SAFMR, cap, resilience, and no-self-check tests.

## Notes for Reviewers

- **Workflow rule:** calculators no longer self-check "does the household already have this program?" — that's the generic `already_has` results-layer filter's job. `wa_hcv`/`tx_hcv` are themselves the `section_8` program, so their old duplicate-subsidy gate was a self-check and was removed. `wa_wap` keeps its check because there Section 8 is a *different* program informing eligibility.
- **Scope:** the Section 8 change (wa_hcv/tx_hcv/wa_wap) is broader than "benefit value," but it was found while QA'ing this program and touches the same `wa_hcv` calculator. Folded in at the maintainer's request.
- **Follow-up (separate ticket):** `wa_lifeline` (`WaLifeline`, pure PolicyEngine) implements **none** of the screener-level categorical overrides its spec prescribes — Section 8, TANF (non-tribal), and WIC. That's a missing-feature gap, not a wrong-check, and is intentionally out of scope here.
- Known limitations: payment standard still assumed at 100% of FMR/SAFMR; medical/childcare deductions and utility allowances remain out of scope per the WA HCV spec.
- WA `BEDROOM_MAP` (1-person → 1BR) and `$50` min rent intentionally differ from TX and are left as-is; TX income-exclusion parity (24 CFR § 5.609(b)) remains a possible follow-up.

## Summary by CodeRabbit

- **New Features**
  - HCV benefit value calculations now use payment standards with reported rent capped to the payment standard (then compute HAP from the capped amount).

- **Bug Fixes**
  - Eligibility/value behavior is more resilient to HUD lookup failures, avoiding unexpected errors and returning safe “not eligible” or zero-value outcomes.
  - Section 8/HCV recognition is consistent across program variants, and duplicate-benefit filtering aligns with the results-layer handling.

- **Documentation**
  - Updated HCV and WAP eligibility criteria and scenario numbering to match the revised duplicate-benefit and rent-payment-standard rules.

- **Tests**
  - Adjusted calculators and acceptance scenarios for the updated inputs and error-handling expectations.

### DIFF
--- a/programs/programs/tx/hcv/calculator.py
+++ b/programs/programs/tx/hcv/calculator.py
@@ -132,7 +132,10 @@ class TxHcv(ProgramCalculator):
 
     def household_eligible(self, e: Eligibility):
         # Criterion 2: not already receiving Section 8 / HCV assistance.
-        e.condition(not self.screen.has_benefit("section_8"), messages.must_not_have_benefit("Section 8"))
+        # "Section 8" is the HCV program itself (base_program "section_8"): tx_hcv,
+        # wa_hcv, co_section_8, etc. Use has_base_benefit so it matches every
+        # white-label variant
+        e.condition(not self.screen.has_base_benefit("section_8"), messages.must_not_have_benefit("Section 8"))
 
         # Criterion 9: HOTMA net-asset limit — only a gate when reported over $100k.
         assets = self.screen.household_assets if self.screen.household_assets is not None else 0
--- a/programs/programs/tx/hcv/calculator.py
+++ b/programs/programs/tx/hcv/calculator.py
@@ -131,11 +131,12 @@ class TxHcv(ProgramCalculator):
         return head.age >= self.min_head_age
 
     def household_eligible(self, e: Eligibility):
-        # Criterion 2: not already receiving Section 8 / HCV assistance.
-        # "Section 8" is the HCV program itself (base_program "section_8"): tx_hcv,
-        # wa_hcv, co_section_8, etc. Use has_base_benefit so it matches every
-        # white-label variant
-        e.condition(not self.screen.has_base_benefit("section_8"), messages.must_not_have_benefit("Section 8"))
+        # Criterion 2 (not already receiving Section 8 / HCV) is intentionally not
+        # checked here: tx_hcv *is* the "section_8" program, so "already has Section
+        # 8" == "already has tx_hcv". That duplicate-subsidy case is handled
+        # generically by the results-layer `already_has` filter
+        # (screen.has_benefit(program.name_abbreviated)), not the calculator —
+        # calculators only check *other* programs to inform eligibility.
 
         # Criterion 9: HOTMA net-asset limit — only a gate when reported over $100k.
         assets = self.screen.household_assets if self.screen.household_assets is not None else 0
--- a/programs/programs/tx/hcv/calculator.py
+++ b/programs/programs/tx/hcv/calculator.py
@@ -131,13 +131,6 @@ class TxHcv(ProgramCalculator):
         return head.age >= self.min_head_age
 
     def household_eligible(self, e: Eligibility):
-        # Criterion 2 (not already receiving Section 8 / HCV) is intentionally not
-        # checked here: tx_hcv *is* the "section_8" program, so "already has Section
-        # 8" == "already has tx_hcv". That duplicate-subsidy case is handled
-        # generically by the results-layer `already_has` filter
-        # (screen.has_benefit(program.name_abbreviated)), not the calculator —
-        # calculators only check *other* programs to inform eligibility.
-
         # Criterion 9: HOTMA net-asset limit — only a gate when reported over $100k.
         assets = self.screen.household_assets if self.screen.household_assets is not None else 0
         e.condition(assets <= self.asset_limit, messages.assets(self.asset_limit))
--- a/programs/programs/tx/hcv/spec.md
+++ b/programs/programs/tx/hcv/spec.md
@@ -20,9 +20,9 @@
 
 2. **Applicant must not currently be receiving HCV/Section 8 assistance.**
 
-   - **Screener fields:** `tx_hcv` current benefit — "Section 8" is the HCV program itself (base_program `section_8`), checked via `has_base_benefit("section_8")` so it matches every white-label variant (`tx_hcv`, `wa_hcv`, `co_section_8`, …).
+   - **Screener fields:** `tx_hcv` current benefit — "Section 8" *is* this program (base_program `section_8`), so "already receiving Section 8" means "already has `tx_hcv`".
    - **Source:** 24 CFR 982.551(n)
-   - **Implementation:** `has_base_benefit("section_8") == true` → ineligible, as a hard override. A household already receiving Section 8/HCV assistance cannot receive a second voucher.
+   - **Implementation:** Not a calculator check. The generic results-layer `already_has` workflow filters `tx_hcv` out of results when the household reports it (via `screen.has_benefit("tx_hcv")`). The calculator does not self-exclude — it only checks *other* programs to inform eligibility.
 
 3. **Head of household, spouse, or co-head must have the legal capacity to enter a lease under Texas law** — in practice, age 18, unless the applicant is an emancipated minor. ⚠️ *data gap — handled conservatively, not by inclusivity default*
 
@@ -413,9 +413,9 @@ SSN is a document requirement (the config's `tx_ssn` document), not an eligibili
 
 ### Scenario 10: Single Adult Already Receiving Section 8/HCV Assistance in Harris County
 
-**What we're checking**: The duplicate-benefit exclusion (Criterion 2) for a household already receiving Section 8.
+**What we're checking**: The duplicate-benefit exclusion (Criterion 2) for a household already receiving Section 8. `tx_hcv` *is* the Section 8 program, so this is not a calculator check — the generic results-layer `already_has` filter removes `tx_hcv` from results when the household reports it. The calculator itself still returns eligible (income permitting).
 
-**Expected**: Not eligible
+**Expected**: Eligible from the calculator; filtered out of results by `already_has` (not shown to the user).
 
 **Steps**:
 - **Location**: Enter ZIP code `77001`, Select county `Harris`
@@ -423,15 +423,15 @@ SSN is a document requirement (the config's `tx_ssn` document), not an eligibili
 - **Person 1**: Birth month/year: `March 1986` (age 40), Relationship: Head of Household, Has income: No
 - **Current Benefits**: Indicate household is **already receiving Section 8 / Housing Choice Voucher assistance** — select the `tx_hcv` current benefit
 
-**Why this matters**: 24 CFR 982.551(n) prohibits duplicate HCV assistance. The JSON test case must include `tx_hcv` in the household's `current_benefits` list to trigger this exclusion — the check is `has_base_benefit("section_8")` (see Criterion 2).
+**Why this matters**: 24 CFR 982.551(n) prohibits duplicate HCV assistance, but because `tx_hcv` is itself the Section 8 program the duplicate case is handled generically by the `already_has` results-layer filter (`screen.has_benefit("tx_hcv")`), not a calculator self-check (see Criterion 2).
 
 ---
 
 ### Scenario 11: Currently Receiving HCV/Section 8 Assistance - Duplicate Benefit Exclusion in Dallas County
 
-**What we're checking**: The same duplicate-benefit exclusion for a multi-person household.
+**What we're checking**: The same duplicate-benefit exclusion for a multi-person household — handled by the `already_has` results-layer filter, not a calculator self-check.
 
-**Expected**: Not eligible
+**Expected**: Eligible from the calculator; filtered out of results by `already_has` (not shown to the user).
 
 **Steps**:
 - **Location**: Enter ZIP code `75201`, Select county `Dallas`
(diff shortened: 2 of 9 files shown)


programs/programs/tx/hcv/calculator.py
```diff
@@ -131,8 +131,12 @@ def _head_age_ok(self) -> bool:
...
 
     def household_eligible(self, e: Eligibility):
-        # Criterion 2: not already receiving Section 8 / HCV assistance.
-        e.condition(not self.screen.has_benefit("section_8"), messages.must_not_have_benefit("Section 8"))
+        # Criterion 2 (not already receiving Section 8 / HCV) is intentionally not
+        # checked here: tx_hcv *is* the "section_8" program, so "already has Section
+        # 8" == "already has tx_hcv". That duplicate-subsidy case is handled
+        # generically by the results-layer `already_has` filter
+        # (screen.has_benefit(program.name_abbreviated)), not the calculator —
+        # calculators only check *other* programs to inform eligibility.
```
catonph: This comment feels like overkill - this isnt' novel behavior but how all our calculators work

programs/programs/wa/hcv/calculator.py
```diff
@@ -68,9 +72,11 @@ def _get_year_period(self) -> str:
...
 
     def household_eligible(self, e: Eligibility):
-        has_section_8 = self.screen.has_benefit("section_8")
-        e.condition(not has_section_8, messages.must_not_have_benefit("Section 8"))
-
+        # No self-exclusion for already receiving Section 8: wa_hcv *is* the
+        # "section_8" program, so "already has Section 8" == "already has wa_hcv".
+        # That duplicate-subsidy case is handled generically by the results-layer
+        # `already_has` filter (screen.has_benefit(program.name_abbreviated)), not
+        # here — calculators only check *other* programs to inform eligibility.
```
catonph: same comment here - don't need this long explanation when this is how every `household_eligible()` function behaves

programs/programs/wa/hcv/spec.md
```diff
@@ -53,8 +53,9 @@
 4. **Applicant household must not currently be receiving Section 8 assistance from any PHA.**
    - Applies to both tenant-based (HCV) and project-based Section 8 — duplicate Section 8 subsidies are prohibited.
    - Screener fields:
-     - `section_8` current benefit (checked via `has_benefit("section_8")`)
-   - Note: Verified at application via HUD's EIV Existing Tenant Search. In the screener, this maps directly to the household's current benefits — when `has_benefit("section_8")` is true, the household is currently receiving Section 8 and is excluded from a duplicate voucher. There is no single clean CFR/USC citation for the prohibition itself — it is operationalized through the EIV procedural check at admission.
+     - `wa_hcv` current benefit — "Section 8" *is* this program (base_program `section_8`), so "already receiving Section 8" means "already has `wa_hcv`".
+   - **Implementation**: Not a calculator check. The generic results-layer `already_has` workflow filters `wa_hcv` out of results when the household reports it (via `screen.has_benefit("wa_hcv")`). The calculator does not self-exclude — it only checks *other* programs to inform eligibility.
```
catonph: This comment doesn't make a ton of sense.....because HCV will never be in step 8 we really dont apply already_has to it.....and then "The calculator does not self-exclude — it only checks *other* programs to inform eligibility." is true generally for our calculators but is that relevant here?

programs/programs/wa/hcv/spec.md
```diff
@@ -364,10 +365,10 @@ Each scenario in the Test Scenarios section below is an acceptance criterion. Th
 
 ---
 
-### Scenario 7: Currently Receiving Section 8 (Criterion 4 Ineligible)
+### Scenario 7: Currently Receiving Section 8 (Criterion 4 — handled by `already_has`)
```
catonph: Probably shouldn't have a test case for this since it can't happen

programs/programs/tx/hcv/spec.md
```diff
@@ -413,33 +413,33 @@ SSN is a document requirement (the config's `tx_ssn` document), not an eligibili
 
 ### Scenario 10: Single Adult Already Receiving Section 8/HCV Assistance in Harris County
```
catonph: Probably shouldn't have a test case for this since it can't actually happen in the user flow

programs/programs/wa/wap/calculator.py
```diff
@@ -39,7 +39,10 @@ def household_eligible(self, e: Eligibility):
             self.screen.has_benefit("wa_snap")
             or self.screen.has_benefit("wa_tanf")
             or self.screen.has_benefit("wa_ssi")
-            or self.screen.has_benefit("wa_hcv")
+            # Section 8 = the HCV program (base_program "section_8", e.g. wa_hcv).
+            # has_base_benefit matches every variant; has_benefit("section_8") is a
+            # dead check because the bare name_abbreviated "section_8" exists nowhere.
+            or self.screen.has_base_benefit("section_8")
```
catonph, line 41: This comment is a bit of overkill - explaining what we used to do which doesn't need to live in a code comment

programs/programs/wa/hcv/spec.md
david-j-stoll: Why would hcv never be in step 8? We use it to determine eligibility for wa_wap: https://github.com/MyFriendBen/benefits-api/blob/main/programs/programs/wa/wap/calculator.py#L42 <-- changing this to base benefit as discussed Friday

catonph (approved):
LGTM!

PR #1642, opened 2026-07-16
MFB-1316: Re-migrate TX FPP to a hybrid PolicyEngine calculator

david-j-stoll:
## Context & Motivation

- Linear: **MFB-1316** — https://linear.app/myfriendben/issue/MFB-1316 (auto-linked via the `david/mfb-1316` branch)
- Related PR: MFB-1088 (the original custom migration this partially reverts)

`tx_fpp` (Texas Family Planning Program) was a **fully custom** calculator (MFB-1088) that re-implemented the income test as a flat `gross_income ≤ 2.5 × FPL`. That copy has **drifted** from PolicyEngine's countable-income model as PE refined it — most recently in the `policyengine-us` 1.755.0 → 1.768.1 update, which revised the TX FPP income-sources list and countable-income formula (child-earnings exemption, child-support paid/received handling, dependent-care deduction). None of those reach our custom copy.

This PR re-migrates `tx_fpp` to a **hybrid PolicyEngine calculator**: PE owns the income determination again, while MFB keeps the two overlays PE cannot model (they depend on data PE never sees). PE's future income fixes then flow in automatically instead of being hand-maintained.

## Changes Made

- **New calculator** `programs/programs/tx/pe/member.py` — `TxFpp(PolicyEngineMembersCalculator)`. Reads PE's eligibility **sub-variables** (not the bundled `tx_fpp_benefit`, which is $0 unless income-eligible and would zero out the §4140 bypass):
  - `tx_fpp_age_eligible` (person) — §4130 age, 64 or younger.
  - `tx_fpp_income_eligible` (spm_unit) — countable income (1 TAC 382.109) ≤ 250% FPG.
- **MFB overlays preserved** (unchanged from MFB-1088): §4140 adjunctive bypass — SNAP (`has_benefit("tx_snap")`) / WIC (`has_benefit("tx_wic")`) / CHIP (`has_insurance_types(("chp",))`); §4100 insurance rule — only full Medicaid disqualifies, Emergency Medicaid and other coverage stay eligible.
- **New output dependencies**: `TxFppAgeEligible` (`dependencies/member.py`), `TxFppIncomeEligible` (`dependencies/spm.py`).
[...]

PR #1644, opened 2026-07-17
MFB-1325: Apply the FPP Policy Manual countable-income rules in the custom TX FPP calculator

david-j-stoll:
## Context & Motivation

- Linear: **MFB-1325** — https://linear.app/myfriendben/issue/MFB-1325 (auto-linked via the `david/mfb-1325` branch)
- Related PR: #1642 (MFB-1316) — the hybrid-PolicyEngine re-migration that was explored and set aside in favor of this approach

The custom TX FPP calculator (from MFB-1088) measured income as a flat `gross_income("all") ≤ 250% FPL`, which diverges from the FPP **Definition of Income**. This PR brings the calculator's income test in line with the **TX FPP Policy Manual** — "Definition of Income" (Rev 24-2, Oct. 15, 2024) and §4130/§4140 — which is the source of truth here. (PolicyEngine models the same rules and was the initial reference, but each rule below was verified against the manual directly.)

Note: no change was strictly *required* — the custom calc never calls PolicyEngine — so this is a deliberate **accuracy** improvement (the old flat total was stricter than the FPP policy), not a forced fix.

## Changes Made

- **`_countable_income()` helper** (`tx/fpp/calculator.py`) replacing the flat `gross_income("all")` test, per the FPP Definition of Income:
  - **a child's earned income is exempt** — earned income summed only for members age ≥ 18 (`child_age_threshold = 18`, per §4140 / 1 TAC §382.109; members with no recorded age are not treated as adults);
  - **child support paid deducted** — legally obligated child support paid by a household member (§4140 Step 2), from the `childSupport` expense;
  - **child support received disregard** — Definition of Income: *"Count income after deducting $75 from the total monthly child support payments the household receives"* → `max(0, received − $75/mo)`;
  - countable income floored at 0.
- New constants `child_age_threshold = 18` and `child_support_received_disregard_monthly = 75`.
- Added `"income_type"` to `dependencies` (the calc now branches on income type), plus `"expense_type"` / `"expense_amount"` (the calc reads the `childSupport` expense; declaring these makes an incomplete expense row skip the program via `DependencyError` instead of crashing `calc_expenses` — hardening against the raw API, which the frontend never triggers).
- **Unchanged/preserved:** age ≤ 64 (no minimum age, §4130), §4100 Medicaid-only exclusion, §4140 SNAP/WIC/CHIP adjunctive bypass, flat $266.84/eligible-member value (TX HHS Women's Health Programs Report FY2024).
- Expanded tests to **47** (`tx/fpp/tests/test_fpp.py`) and updated `tx/fpp/spec.md` (added ticket refs; replaced the deprecated "Validation Scenarios" section with a calculator-derived "Test Scenarios" section that the unit tests mirror).

## Testing

- Migrations to run: none (no model/schema changes)
- Manual testing steps:
  - `pytest programs/programs/tx/fpp/tests/test_fpp.py` → 47 passed
  - `pytest programs/programs/tx` → passed (full TX suite)
  - This calc does **not** call PolicyEngine, so unit tests are the primary signal. Optional end-to-end check: submit a TX screen with (a) a working under-18 member, (b) child support paid, and (c) child support received near the 250% FPL line, and confirm `tx_fpp` reflects the exemption/deduction/disregard.

## Deployment

- Admin updates needed: none
- Notify team/users of: TX FPP income eligibility now uses countable income per the FPP Definition of Income. A small set of near-250%-FPL TX households — those with a working minor, child support paid, or modest child support received — may now qualify where they previously didn't. These are accuracy corrections (the old flat-income test was stricter), not regressions.

## Notes for Reviewers

- Verification: each income rule in this PR was checked against the FPP Policy Manual (§4130, §4140, and the "Definition of Income" table, Rev 24-2) — not merely against PolicyEngine.
- **Known limitation (pre-existing, not introduced here):** the Definition of Income "Types of Income" table also marks several types **Exempt** that `_countable_income()` still counts via the broad `unearned` bucket — notably **SSI, TANF, dividends/interest (`investment`), EITC**, and various assistance/adoption/foster-care/in-kind payments. The `exclude=` argument on `calc_gross_income` supports fixing this (an FPP-exempt income-type list); it's a bounded follow-up, deliberately out of scope here. Effect: the calc over-counts income for a narrow set of households and can under-approve at the margin.
- Future considerations:
  - Address the exempt-income-types gap above (follow-up).
  - The §4140 adjunctive list in the manual is CHIP **Perinatal**/SNAP/WIC; we use regular CHIP insurance as a proxy (existing data gap, noted in the spec).
  - Re-check against the FPP Policy Manual if it's revised (current basis: Rev 24-2).

## Summary by CodeRabbit

* **New Features**
  * Updated FPP eligibility to use policy-defined countable income (earned/unearned, with floors applied).
  * Earned income is excluded for members below the configured age threshold.
  * Added child support rules: $75/month disregard for received and deduction for paid.
  * Updated adjunctive bypass behavior for TX-specific SNAP/WIC identifiers and CHIP handling.
* **Bug Fixes**
  * Corrected eligibility outcomes for edge cases involving unknown ages and child support aggregation.
  * Prevented bypass from incorrectly qualifying otherwise ineligible households.
* **Documentation**
  * Refreshed the TX FPP spec, field mappings, and benefit value wording.
* **Tests**
  * Expanded unit and end-to-end test coverage for countable-income components and bypass scenarios.

### DIFF
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -118,21 +118,19 @@ class TxFpp(ProgramCalculator):
         that table (an ``exclude`` list of the FPP-exempt income types) is a tracked
         follow-up, not implemented here.
         """
-        # Earned income counts only for adults (under-18 earnings are exempt). Members
-        # with no recorded age are not counted as adults.
+        # Adults only — under-18 (and unknown-age) earnings are exempt.
         adult_earned = sum(
             member.calc_gross_income("yearly", ["earned"])
             for member in self.screen.household_members.all()
             if member.age is not None and member.age >= self.child_age_threshold
         )
 
-        # Unearned income counts for all ages. Child support received is pulled out of
-        # the unearned bucket and re-added net of its (annualized) disregard.
+        # Child support received is pulled from the unearned bucket and re-added net of its
+        # annualized disregard.
         unearned = self.screen.calc_gross_income("yearly", ["unearned"], exclude=["childSupport"])
         child_support_received = self.screen.calc_gross_income("yearly", ["childSupport"])
         countable_child_support = max(0, child_support_received - self.child_support_received_disregard_monthly * 12)
 
-        # Child support paid is deducted.
         child_support_paid = self.screen.calc_expenses("yearly", ["childSupport"])
 
         return int(max(0, adult_earned + unearned + countable_child_support - child_support_paid))
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -59,6 +59,11 @@ class TxFpp(ProgramCalculator):
         "income_type",
         "income_amount",
         "income_frequency",
+        # _countable_income() reads the childSupport expense via calc_expenses(); declare the
+        # expense dependencies so an incomplete expense row makes can_calc() skip this program
+        # (DependencyError) instead of crashing calc_expenses() and 500-ing the whole request.
+        "expense_type",
+        "expense_amount",
         "household_size",
     ]
 
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -19,6 +19,11 @@ class TxFpp(ProgramCalculator):
       adjunctive income eligibility via enrollment in SNAP, WIC, or CHIP — applicant
       or their child (§4140). CHIP Perinatal, the 4th §4140 program, is not collected
       by the screener and is tracked as a data gap.
+      The income test uses a *countable* income that mirrors PolicyEngine's
+      gov.states.tx.fpp countable-income model at the version we serve (policyengine-us
+      1.768.1) — see ``_countable_income``. It is NOT a flat gross-income total; PE is
+      treated as the source of truth (not independently verified against 1 TAC §382.109
+      / the FPP Policy Manual). If PE changes its FPP countable-income logic, re-sync.
     - Not enrolled in (full) Medicaid (§4100). FPP serves those who earn too much for
       regular Medicaid. Emergency Medicaid recipients are classified as underinsured
       and remain eligible, so they are NOT excluded. Other coverage (employer, private,
@@ -38,9 +43,22 @@ class TxFpp(ProgramCalculator):
     max_age = 64
     fpl_percent = 2.5
     member_amount = 266.84
+
+    # Earnings of members below this age are exempt from countable income — mirrors
+    # PolicyEngine (1 TAC §382.109(3)(A); FPP Policy Manual 4140 treats 18-year-olds as
+    # adults). PE is treated as the source of truth here; not independently verified.
+    child_age_threshold = 18
+
+    # Monthly disregard applied to child support *received* before it counts as income —
+    # only the amount above this counts. $75/month per the Texas FPP Policy Manual
+    # (Definition of Income, Rev 24-2), mirroring PolicyEngine's
+    # gov.states.tx.fpp.income.child_support_disregard (unchanged since 2016-07-01).
+    child_support_received_disregard_monthly = 75
+
     dependencies: ClassVar[list[str]] = [
         "age",
         "insurance",
+        "income_type",
         "income_amount",
         "income_frequency",
         "household_size",
@@ -83,6 +101,40 @@ class TxFpp(ProgramCalculator):
             e.condition(True, messages.presumed_eligibility())
             return
 
-        gross_income = int(self.screen.calc_gross_income("yearly", ["all"]))
+        countable_income = self._countable_income()
         income_limit = int(self.fpl_percent * self.program.year.get_limit(self.screen.household_size))
-        e.condition(gross_income <= income_limit, messages.income(gross_income, income_limit))
+        e.condition(countable_income <= income_limit, messages.income(countable_income, income_limit))
+
+    def _countable_income(self) -> int:
+        """TX FPP countable income, mirroring PolicyEngine's gov.states.tx.fpp
+        (``tx_fpp_countable_income``) at the version we serve (policyengine-us 1.768.1).
+
+        Differs from a flat gross-income total in three ways (PE treated as source of
+        truth; not independently verified against 1 TAC §382.109 / the FPP Policy Manual):
+          - earnings of members under ``child_age_threshold`` are exempt,
+          - child support *paid* is deducted,
+          - child support *received* counts only above a monthly disregard.
+
+        The dependent-care deduction PE added in 1.771.2 (frontier) is intentionally NOT
+        applied — it is absent from the 1.768.1 model we serve.
+        """
+        # Earned income counts only for adults (under-18 earnings are exempt). Members
+        # with no recorded age are not counted as adults.
+        adult_earned = sum(
+            member.calc_gross_income("yearly", ["earned"])
+            for member in self.screen.household_members.all()
+            if member.age is not None and member.age >= self.child_age_threshold
+        )
+
+        # Unearned income counts for all ages. Child support received is pulled out of
+        # the unearned bucket and re-added net of its (annualized) disregard.
+        unearned = self.screen.calc_gross_income("yearly", ["unearned"], exclude=["childSupport"])
+        child_support_received = self.screen.calc_gross_income("yearly", ["childSupport"])
+        countable_child_support = max(
+            0, child_support_received - self.child_support_received_disregard_monthly * 12
+        )
+
+        # Child support paid is deducted.
+        child_support_paid = self.screen.calc_expenses("yearly", ["childSupport"])
+
+        return int(max(0, adult_earned + unearned + countable_child_support - child_support_paid))
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -130,9 +130,7 @@ class TxFpp(ProgramCalculator):
         # the unearned bucket and re-added net of its (annualized) disregard.
         unearned = self.screen.calc_gross_income("yearly", ["unearned"], exclude=["childSupport"])
         child_support_received = self.screen.calc_gross_income("yearly", ["childSupport"])
-        countable_child_support = max(
-            0, child_support_received - self.child_support_received_disregard_monthly * 12
-        )
+        countable_child_support = max(0, child_support_received - self.child_support_received_disregard_monthly * 12)
 
         # Child support paid is deducted.
         child_support_paid = self.screen.calc_expenses("yearly", ["childSupport"])
--- a/programs/programs/tx/fpp/calculator.py
+++ b/programs/programs/tx/fpp/calculator.py
@@ -19,11 +19,9 @@ class TxFpp(ProgramCalculator):
       adjunctive income eligibility via enrollment in SNAP, WIC, or CHIP — applicant
       or their child (§4140). CHIP Perinatal, the 4th §4140 program, is not collected
       by the screener and is tracked as a data gap.
-      The income test uses a *countable* income that mirrors PolicyEngine's
-      gov.states.tx.fpp countable-income model at the version we serve (policyengine-us
-      1.768.1) — see ``_countable_income``. It is NOT a flat gross-income total; PE is
-      treated as the source of truth (not independently verified against 1 TAC §382.109
-      / the FPP Policy Manual). If PE changes its FPP countable-income logic, re-sync.
+      The income test uses a *countable* income (see ``_countable_income``), not a flat
+      gross-income total, per the FPP Policy Manual "Definition of Income" (Rev 24-2,
+      Oct. 15, 2024) and §4140. Re-check against the manual if it is revised.
     - Not enrolled in (full) Medicaid (§4100). FPP serves those who earn too much for
       regular Medicaid. Emergency Medicaid recipients are classified as underinsured
       and remain eligible, so they are NOT excluded. Other coverage (employer, private,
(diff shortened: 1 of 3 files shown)


programs/programs/tx/fpp/spec.md
```diff
@@ -3,7 +3,7 @@
 - **Program:** Family Planning Program (FPP)
 - **State:** TX (`tx`)
 - **name_abbreviated:** `tx_fpp`
-- **Ticket:** MFB-1088
+- **Tickets:** MFB-1088 (custom migration), MFB-1325 (income mirror of PolicyEngine)
```
catonph: No ticket references in the spec.md

programs/programs/tx/fpp/calculator.py
```diff
@@ -83,6 +99,40 @@ def household_eligible(self, e: Eligibility):
...
+        # Unearned income counts for all ages. Child support received is pulled out of
+        # the unearned bucket and re-added net of its (annualized) disregard.
+        unearned = self.screen.calc_gross_income("yearly", ["unearned"], exclude=["childSupport"])
+        child_support_received = self.screen.calc_gross_income("yearly", ["childSupport"])
+        countable_child_support = max(0, child_support_received - self.child_support_received_disregard_monthly * 12)
+
+        # Child support paid is deducted.
+        child_support_paid = self.screen.calc_expenses("yearly", ["childSupport"])
+
+        return int(max(0, adult_earned + unearned + countable_child_support - child_support_paid))
```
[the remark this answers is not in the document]
david-j-stoll: This is already documented as a follow-up

programs/programs/tx/fpp/tests/test_fpp.py
```diff
@@ -31,46 +39,53 @@ def make_member(age=30, medicaid=False, emergency_medicaid=False, employer=False
...
     mock_program.year.get_limit.return_value = fpl_limit
 
     benefits = set(current_benefits or [])
 
+    def _gross(freq, types, exclude=[]):
+        if "unearned" in types:
+            return unearned
+        if "childSupport" in types:
+            return child_support_received
+        return 0
```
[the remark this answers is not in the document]
david-j-stoll: This is already resolved

programs/programs/tx/fpp/tests/test_fpp.py
```diff
@@ -142,34 +165,124 @@ def test_employer_insurance_does_not_disqualify(self):
...
+        """Both child-support branches apply together: $20k unearned + ($10k received −
+        $900 disregard) − $5k paid = $24,100."""
+        calc = make_calculator(
+            unearned=20_000, child_support_received=10_000, child_support_paid=5_000, fpl_limit=15_000
+        )
+        self.assertEqual(calc._countable_income(), 24_100)
+
+    def test_countable_income_combines_all_components(self):
+        """adult earned + unearned + disregarded child support − child support paid.
+        $18k adult earned + $10k unearned + ($5k received − $900) − $2k paid = $30,100."""
```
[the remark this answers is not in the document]
david-j-stoll: Not blocking, skipping for now

programs/programs/tx/fpp/spec.md
```diff
@@ -63,8 +85,8 @@ prior strict "no insurance" filter to a Medicaid-only exclusion.
 | Criterion | Fields | Notes |
 |-----------|--------|-------|
 | Age ≤ 64 | `birth_year`, `birth_month` | per-member; None-safe |
-| Income ≤ 250% FPL | `calc_gross_income`, `household_size`, `program.year` | gross income, all types |
-| §4140 bypass | `has_benefit("snap")`, `has_benefit("wic")`, `has_benefit("chp")` | CHIP Perinatal not captured (gap) |
+| Countable income ≤ 250% FPL | per-member `calc_gross_income(["earned"])` (adults only), screen `["unearned"]` / `["childSupport"]`, `calc_expenses(["childSupport"])`, `household_size`, `program.year` | per FPP Definition of Income (Rev 24-2) + §4140; $75/mo child-support-received disregard; exempt income types not yet excluded (see limitation) |
+| §4140 bypass | `has_benefit("tx_snap")`, `has_benefit("tx_wic")`, `has_insurance_types(("chp",))` | CHIP Perinatal not captured (gap) |
 | Medicaid exclusion | `member.insurance.medicaid` | Emergency Medicaid excluded from match |
 
 ## Validation Scenarios
```
catonph: Lets remove this section

programs/programs/tx/fpp/spec.md
```diff
@@ -19,17 +18,40 @@ prior strict "no insurance" filter to a Medicaid-only exclusion.
...
-     regardless of the 250% FPL test:
-     - SNAP — `screen.has_benefit("snap")`
-     - WIC — `screen.has_benefit("wic")`
-     - CHIP (applicant or their child) — `screen.has_benefit("chp")`
+     regardless of the countable-income test:
+     - SNAP — `screen.has_benefit("tx_snap")`
+     - WIC — `screen.has_benefit("tx_wic")`
+     - CHIP (applicant or their child) — `screen.has_insurance_types(("chp",))` (per-member
+       insurance; `tx_chip` is a PE eligibility program, never a current-benefit tile, so
+       `has_benefit("tx_chip")` is always False)
```
catonph, line 55: We should probably add a test case or two to cover this new definition

programs/programs/tx/fpp/calculator.py
catonph, line 141: Non-block but some of these comments seem to be repetitive

elliottwaissbluth:
# Code Review: MFB-1325 — TX FPP Countable-Income Rules

**PR**: [#1644](https://github.com/MyFriendBen/benefits-api/pull/1644) — Apply the FPP Policy Manual countable-income rules in the custom TX FPP calculator
**Ticket**: [MFB-1325](https://linear.app/myfriendben/issue/MFB-1325)
**Files**: `programs/programs/tx/fpp/calculator.py`, `spec.md`, `tests/test_fpp.py` (+348 / −62)
**Reviewed**: 2026-07-22 (8 finder angles → 27 candidates → 12 independently verified: 5 confirmed, 3 plausible, 4 refuted)

---

## Summary

The PR is a well-documented accuracy improvement: it replaces the flat `gross_income("all") ≤ 250% FPL` test with a `_countable_income()` that exempts under-18 earnings, deducts child support paid, and applies the $75/mo disregard to child support received — each rule tied to the FPP Policy Manual. The mechanics were verified against the real `Screen`/`HouseholdMember` models and check out: `childSupport` is a valid income *and* expense type, `exclude=` works as assumed, earned+unearned partitions "all" exactly, and no income type is silently dropped.

Two things should be fixed before merge: a **missing expense dependency** that can 500 the entire eligibility request, and a **test mock that can't detect a child-support double-count regression**.

---

## Findings

| # | Severity | Verdict | Finding |
|---|----------|---------|---------|
| 1 | 🔴 High | CONFIRMED | Reads expenses but declares no expense dependency → potential 500 for the whole request |
| 2 | 🟠 Medium | PLAUSIBLE | Under-18 applicant's own wages exempted by the age-only filter |
| 3 | 🟠 Medium | CONFIRMED | Test mock ignores `exclude` — child-support double-count regression would stay green |
| 4 | 🟡 Low | CONFIRMED | No validation scenario exercises any of the three new income rules |
| 5 | 🟡 Low | CONFIRMED | Exempt income types hardcoded inline instead of a declared class list |
| 6 | 🟡 Low | PLAUSIBLE | Countable-income logic private to TxFpp; TxHtw may need it later |
| 7 | ⚪ Info | CONFIRMED | ~3x income queries in non-prefetched batch paths |
| 8 | ⚪ Info | PLAUSIBLE | Constant-echo tests and attribute-mutating control tests |

---

### 1. Missing expense dependency — crash path for the whole eligibility request 🔴
`calculator.py:56` (`dependencies` list) — **CONFIRMED**

`_countable_income()` now calls `self.screen.calc_expenses("yearly", ["childSupport"])`, but `dependencies` adds only `"income_type"` — no `"expense_type"`/`"expense_amount"` (the names `Expense.missing_fields()` emits). The failure chain is fully constructible:

- `Expense.amount` and `Expense.frequency` are `null=True`, and the unmodified `ModelSerializer` accepts null — an incomplete `childSupport` expense row can be persisted via the API.
- `Screen.missing_fields()` reports `expense_amount`, but since TxFpp doesn't declare it, `can_calc()` passes.
- `Expense.yearly()` then evaluates `None * 12` → `TypeError` (or `UnboundLocalError` for `frequency=None`, which `missing_fields()` doesn't even track).
- The per-program loop in `screener/views.py:482-485` catches **only** `DependencyError`, so the exception fails the entire `/eligibility_results` request — one bad expense row takes down results for every program.

**Fix**: declare the expense dependencies (precedent: `co/denver_property_tax_relief` declares `"expense_type"`).

### 2. Age-only child exemption exempts an under-18 applicant's own wages 🟠
`calculator.py:125` — **PLAUSIBLE**

The exemption filters purely by `member.age >= 18`. The manual exempts "a child's earned income" — a household-composition concept. A 16-year-old **head of household** earning $45k (reachable: neither `HouseholdMemberSerializer.validate` nor the frontend birth-year schema constrains the head's age) gets all earnings exempted → countable income $0 → FPP-eligible at any wage.

The PR documents a defensible age-based reading (1 TAC §382.109), so this is a policy-interpretation gap rather than an unambiguous bug. Cheap hardening: don't exempt head/spouse earnings, or reuse the relationship-aware pattern from `ks/k40h/calculator.py:154`.

### 3. Mock ignores `exclude` — double-count regression is untestable 🟠
`tests/test_fpp.py:74` (`_gross`) — **CONFIRMED**

In the real `HouseholdMember.calc_gross_income` (`screener/models.py:466`), `childSupport` falls in the `unearned` bucket and only `exclude` removes it. The mock returns the `unearned` fixture for any `["unearned"]` call regardless of `exclude`, baking the exclusion into the test world. If a future edit drops `exclude=["childSupport"]` from `calculator.py:130`, production counts child support twice (e.g. $20k unearned + $10k received → $39,100 countable instead of $29,100, wrongly denying near-limit households) while **every test stays green** — no test asserts the call args.

**Fix**: make the mock subtract `child_support_received` only when `"childSupport"` is in `exclude`, or assert the call arguments.

### 4. Validation scenarios don't exercise any new rule 🟡
`validations/.../data/tx_fpp.json` — **CONFIRMED**

All 11 scenarios have `"expenses": []`, no `childSupport` income stream, and no earning minor (the only under-19 member has `has_income: false`). Validations run through the real `ScreenSerializer`, so they're the only integration-l

david-j-stoll:
Thanks for the thorough review, @elliottwaissbluth. Here's how we addressed each finding.

## Fixed

**#1 — Missing expense dependency (🔴 → reframed as hardening).** Added `expense_type` + `expense_amount` to `dependencies`, guarded by `test_declares_expense_dependencies`. One correction on severity, though: the crash is reachable only through the raw `/api/screens` endpoint (`ExpenseSerializer` is an unvalidated `ModelSerializer` and `amount`/`frequency` are `null=True`). The production frontend (`benefits-calculator`) **cannot** trigger it — its Zod `expenseRowSchema` requires `expenseAmount: z.number().int().min(0)` and a `frequency` enum, and it filters to `amount > 0` before submit, so no incomplete row is ever sent. So this is defense-in-depth against the API contract (matches the `denver_property_tax_relief` precedent), not a live production crash today.

**#3 — Test mock ignored `exclude`.** Fixed the `_gross` mock so `childSupport` is included in the `["unearned"]` result unless it's in `exclude`, mirroring the real model. Also added an explicit guard, `test_unearned_query_excludes_child_support`, asserting the calculator passes `exclude=["childSupport"]` — so a dropped exclude (the double-count regression) now fails the suite regardless of mock fidelity.

## Not changing (with reasons)

**#2 — Under-18 applicant's own wages (PLAUSIBLE → refuted against the spec).** Validated against `spec.md`, which is our source of truth. The FPP Definition of Income exempts "a child's earned income," and the spec defines "child" purely by age (< 18) with an explicit **no minimum age** for applicants — there is no relationship qualifier and nothing restricting a minor from applying. So under the FPP rule a 13–17-year-old (MFB's minimum user age is 13) *is* a child, and their earned income is correctly exempt whether or not they're head of household. The relationship-aware filter cited is `ks/k40h`'s income definition — a different program's rule; importing it would make TX FPP diverge from its own manual. Keeping the age-based exemption as documented.

**#4 — Validation scenarios don't exercise the new rules (CONFIRMED → obsolete).** We're retiring the `import_validations` JSON flow; coverage now lives in unit tests. The spec's "Validation Scenarios" section has been replaced with a calculator-derived "Test Scenarios" section that the unit tests mirror, so all three new income rules are covered there.

**#5 — Exempt income types as a declared class list.** The only inline literal today is `["childSupport"]`; the broader exempt-types list (SSI/TANF/EITC/…) is already tracked as the documented follow-up. We'll introduce the class-level list when we implement that follow-up rather than add a constant for a single value now.

**#6 — Extract countable-income logic for TxHtw reuse.** No current caller — TxHtw doesn't use it today. We'd rather not pre-extract a shared abstraction speculatively; happy to lift it out when a second consumer actually needs it.

**#7 — ~3× income queries in non-prefetched paths.** Acknowledged, but this is the same per-member `calc_gross_income` pattern used across existing calculators (e.g. `k40h`) and isn't a regression specific to this PR. Left as-is.

**#8 — Constant-echo and attribute-mutating control tests.** The control tests (setting the disregard / age threshold to 0) are intentional contrast tests that prove the rule actually drives the outcome. Keeping them.

---

Test count is now **47** (was 45), `black` clean. PR description updated to match.

elliottwaissbluth (approved):

PR #1662, opened 2026-07-28
Reland Redis cache (MFB-468) with translation-endpoint fixes

david-j-stoll:
## Context & Motivation

Relands #1594 (reverted in #1661) with the defects that took staging's `/api/translations/` down fixed, plus the Redis test coverage that would have caught them.

- Resolves [MFB-468](https://linear.app/myfriendben/issue/MFB-468/general-replace-in-memory-custom-cache-with-redis)
- Relands: #1594 · Revert being undone: #1661
- Blocked-by ticket MFB-457 (Redis addon) is already provisioned on both apps

### Why #1594 passed CI and review but failed in staging

**1. Redis was 100% non-functional, silently.** Heroku serves `rediss://` with a self-signed cert. redis-py defaults to `ssl_cert_reqs="required"` and rejected every connection; `IGNORE_EXCEPTIONS: True` converted that into a permanent cache miss. Staging's Redis had `DBSIZE` **0** — not one key was ever written.

```
cache.set -> None            # not True
raw ping FAILED: ConnectionError [SSL: CERTIFICATE_VERIFY_FAILED]
               certificate verify failed: self-signed certificate in certificate chain
```

**2. CI never exercised Redis.** `pr-validation.yaml` sets no `REDIS_URL`, so `settings.py` always took the `LocMemCache` branch. The whole `django_redis` path — serialization, key handling, pooling, the TLS config — had zero coverage.

With the cache permanently missing, **every** request rebuilt the full translation map. That rebuild walked parler's descriptors, and each `translation.text` access writes back via `_cache_translation()` — so one request issued ~156k cache writes (8,583 active labels × 18 languages). At ~22 s it crossed Heroku's 30 s router timeout, and `views.py` retried the entire build inline on failure, doubling it. `WEB_CONCURRENCY=1` meant the stalled worker also 503'd `/api/configuration/` and `/api/screener-options/`.

**3. It could not have worked on the current plan anyway.** Both apps run Heroku Redis **mini: 25 MB, `noeviction`**. The single combined key is **15.31 MB** — 61% of the instance — plus ~156k parler keys, into a Redis that *fails writes* rather than evicting when full.

## Changes Made

**Translation cache** (`translations/models.py`)
- Build from one `values_list()` query over the translated table instead of parler's descriptors. Verified byte-identical across all **156,150** values.
- One cache key **per language** instead of a combined blob, and versioned (`translation_data:v2:{lang}`) so a future format change is self-invalidating rather than needing a coordinated flush.

**Settings** (`benefits/settings.py`, `benefits/cache_config.py`)
- `ssl_cert_reqs=None` for `rediss://` **only** — passing it to a plain `redis://` pool raises `TypeError`, hence the scheme check. Extracted into an importable helper so it is unit-testable without a TLS server.
- zlib `COMPRESSOR` enabled.
- `PARLER_ENABLE_CACHING = False` — nothing reads parler's per-row entries, and writing them is what made a cold rebuild unserveable.
- `max_connections` 50 → 15; the mini plan allows **20 in total** across all dynos and workers.
- `DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = True` so a swallowed error is visible instead of appearing only as latency.

**View** (`translations/views.py`) — stop retrying the rebuild inside the request handler.

### Measured impact

| | #1594 | this PR |
|---|---|---|
| warm request | 12.3 s local / never warmed on staging | **0.002 s** |
| cold rebuild | 12.3 s local / ~22 s staging | **0.59 s** |
| peak heap per rebuild | 186 MB | **71 MB** |
| Redis keys | 1 blob + ~156k parler | **18** |
| Redis memory | 15.31 MB blob, over the 25 MB ceiling | **6.52 MB, 0 evictions** |
| per-request read | 15.31 MB | **0.38 MB** |

The heap number matters directly: MFB-468 was filed for `R14 Memory quota exceeded` (`mem=551M(102.6%)`), and a 186 MB spike per rebuild on a 512 MB dyno is consistent with the R15 kills staging has seen under concurrent cold rebuilds.

## Testing

- Migrations to run: **none**
- Configuration updates needed: **none** — `REDIS_URL` is already set on both apps by the addon
- Environment variables/settings to add: none for deploy. Locally, `REDIS_URL` is optional; without it you get `LocMemCache`.

**Automated**

- `redis:8` service container added to `pr-validation.yaml` with `REDIS_URL`, so CI exercises `django_redis` instead of silently falling back.
- `benefits/tests/test_redis_backend.py` — runs against a real Redis, skips when unreachable so local runs stay green. Round-trips, bulk `get_many`/`set_many`/`delete_many`, and asserts compression is actually applied.
- `benefits/tests/test_cache_config.py` — pins `rediss://` vs `redis://` scheme handling and the connection ceiling. Includes a test that a plain pool *rejects* `ssl_cert_reqs`, documenting why the check exists.
- `translations/tests.py` — per-language keying, key versioning, `use_fallback` semantics, inactive-label exclusion, single-query rebuild, invalidation across all languages.
- Cache-asserting tests now pin the backend via `benefits.test_cache`, and an autouse `conftest.py` fixture clears the cache between tests (Redis state outlives a test; LocMem's did not).

Suite is green in all three configurations:

| configuration | result |
|---|---|
| Redis up (CI) | **2534 passed**, 19 skipped |
| Redis configured but **down** | **2526 passed**, 27 skipped — previously **10 failed** |
| no `REDIS_URL` (LocMem) | **2526 passed**, 27 skipped |

**Manual** — verified against a local Redis configured to mirror the mini plan (`--maxmemory 25mb --maxmemory-policy allkeys-lru`): 18 keys, 6.52 MB, zero evictions, warm request 0.002 s.

## Deployment

- Post-deployment scripts needed: **none**. The cache populates lazily on the first request; a cold rebuild is now 0.59 s.
- Production config updates: recommend setting `maxmemory-policy` to `allkeys-lru` on both addons — currently `noeviction`, which turns a full cache into silent write failures rather than shedding old keys. Suggested staging-first:
  ```
  heroku redis:maxmemory --policy allkeys-lru --app cobenefits-api-staging
  heroku redis:maxmemory --policy allkeys-lru --app cobenefits-api
  ```
  Worth confirming nothing on these instances needs durability first — I checked only the cache callers.
- Admin updates needed: none.
- Notify team/users of: nothing user-facing. Staging is currently healthy on the revert; this restores the Redis migration.

## Notes for Reviewers

The two commits are deliberately separate — `e423a3fd` is a clean revert-of-the-revert (verified byte-identical to `2950d2ed`), and `93215616` is the fixes on top. Reviewing the second commit alone shows exactly what changed relative to the original PR.

**Known limitations**

- `IGNORE_EXCEPTIONS: True` is kept, so a Redis outage still degrades rather than erroring — but it now logs. A boot-time cache health check would surface misconfiguration even earlier; not included here to keep the diff scoped.
- CI's Redis is not memory-constrained to 25 MB (GitHub service containers can't override the container command, and the test dataset is too small to hit a ceiling anyway). The sizing work was verified locally against a mini-equivalent Redis.
- Running the suite with `REDIS_URL` pointing at a Redis you also use for development will flush that cache database, since the isolation fixture calls `cache.clear()`.

**Future considerations**

- `export_translations()` (`translations/models.py:106`) still uses the same per-language parler walk. Nothing user-facing calls it, so it is out of scope here, but it will be slow for the same reason.
- If more caches move onto this Redis, the mini plan's 25 MB and 20-connection limits are the next constraints to hit.

## Summary by CodeRabbit

* **New Features**
  * Added Redis-backed caching with TLS support, compression, connection retry handling, and safe fallback to local caching.
  * Added resilient Google Sheets caching with fresh-data and stale-data fallback support.
  * Improved translation caching with language-specific entries, default-language fallback, and automatic invalidation.
  * PolicyEngine authentication tokens now refresh automatically after unauthorized responses.

* **Bug Fixes**
  * Improved handling of malformed spreadsheet data and missing counties or income values.
  * Added consistent cache isolation across tests and expanded Redis integration coverage.


david-j-stoll:
Closing as duplicate

PR #1663, opened 2026-07-28
MFB-468: Replace in-memory custom Cache with Redis (reland of #1594)

david-j-stoll:
## Context & Motivation

- Fixes [MFB-468](https://linear.app/myfriendben/issue/MFB-468/general-replace-in-memory-custom-cache-with-redis) — no GitHub issue; tracked in Linear
- Related PR: relands #1594, which was reverted by #1661

MFB-468 exists because `LocMemCache` was exhausting the dyno memory quota (`R14`, `mem=551M(102.6%)`), so the cache needs to live outside dyno RAM. #1594 did that but took staging's `/api/translations/` down for ~3h with `H12` timeouts and was reverted. This relands it with the defects fixed and the test coverage that would have caught them.

Two defects caused the outage, and both were invisible to CI and review:

1. **Redis was 100% non-functional, silently.** Heroku serves `rediss://` with a self-signed cert. redis-py defaults to `ssl_cert_reqs="required"` and rejected every connection, while `IGNORE_EXCEPTIONS: True` turned that into a permanent cache miss. Staging's Redis had `DBSIZE` **0** — not one key was ever written.
2. **CI never exercised Redis.** `pr-validation.yaml` sets no `REDIS_URL`, so `settings.py` always took the `LocMemCache` branch. The whole `django_redis` path — TLS config, serialization, key handling, pooling — had zero coverage.

With the cache permanently missing, every request rebuilt the full translation map. That rebuild walked parler's descriptors, and each `translation.text` access writes back via `_cache_translation()`, so one request issued ~156k cache writes (8.6k active labels × 18 languages). At ~22s it crossed Heroku's 30s router timeout, and `views.py` retried the whole build inline, doubling it. `WEB_CONCURRENCY=1` meant the stalled worker also 503'd `/api/configuration/` and `/api/screener-options/`.

Separately, #1594 could not have worked on the current plan regardless: both apps run Heroku Redis **mini (25 MB, formerly `noeviction`)**, and the single combined cache key is **15.31 MB** — 61% of the instance — plus ~156k parler keys, into a Redis that *fails writes* rather than evicting when full.

## Changes Made

**Translation cache** (`translations/models.py`)
- Build the map from one `values_list()` query over the translated table instead of parler's descriptors. Output verified byte-identical across all 156,150 values.
- Cache one key **per language** instead of a single combined blob, and version the key (`translation_data:v2:{lang}`) so a future format change is self-invalidating rather than needing a coordinated flush.

**Settings** (`benefits/settings.py`, new `benefits/cache_config.py`)
- `ssl_cert_reqs=None` for `rediss://` **only** — passing it to a plain `redis://` pool raises `TypeError`, hence the scheme check. Extracted into an importable helper so it is unit-testable without a TLS server.
- Enable the zlib `COMPRESSOR` (~3.5x on translation payloads).
- `PARLER_ENABLE_CACHING = False` — nothing reads parler's per-row entries, and writing them is what made a cold rebuild unserveable.
- `max_connections` 50 → 15; Heroku Redis mini allows **20 in total** across all dynos and workers.
- `DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = True` so a swallowed error is visible instead of surfacing only as latency.

**View** (`translations/views.py`) — stop retrying the rebuild inside the request handler, and stop invalidating on failure. A failed rebuild writes nothing, so there was no bad entry to clear, while clearing discarded every still-warm language and forced the next request into a full rebuild against the same failing dependency (reproduced: 17 warm keys -> 0; now 17 -> 17). Also dropped `str(error)` from the 503 body, which exposed internal exception text.

**Post-review fixes** — eight further issues surfaced in review, each verified before and after:

- **FPL cache removed** (`programs/models.py`). `_get_fpl_data()` was doing a Redis round-trip to read a module-level constant across **56 call sites**. The pre-existing `FplCache.update()` opened with `return self.default`, leaving its ASPE API fetch unreachable, so the cached value was only ever that constant. It also had an identity hazard: a miss returned `_FPL_DEFAULTS` itself while a hit returned a deserialized copy, so a caller mutating the result corrupted the constant *only on the miss path*. Now returns the constant directly; `_invalidate_fpl_cache` and the dead `try/except` are gone, along with the now-unused `cache` and `requests` imports.
- **Per-language TTL jitter** (`translations/models.py`). All 18 keys shared an identical 24h TTL, so they expired simultaneously and made the concurrent-cold-rebuild case a *daily certainty*. Now spread over a 2h window via `zlib.crc32(lang)` — not builtin `hash()`, which is salted per process and would make dynos disagree about expiry. Verified on Redis: 18 distinct TTLs spread across 101 minutes. This required per-language `cache.set` instead of `set_many` (~31ms extra on a rebuild that now happens at most daily).
- **Missing-county lookups now report** (`connect_for_health`, `nc_aca`). Both had changed `values[county]` to `values.get(county, 0)`. Because the frontend filters out programs whose value is not `> 0`, a renamed county or an empty sheet fetch made the program vanish from results with no error anywhere. Both now emit `capture_message` alongside the 0.
- **`BoulderAmiCache` all-zeros poisoning** (`nurturing_futures`). `[0] * 8` is truthy, so an all-malformed sheet row passed `get_data()`'s `if not data` guard and was cached for 24h *and* written to the 7-day stale key — making every household ineligible, silently. Now returns `[]` so it routes through stale/fallback, and reports.
- **`RAGCache`** now reports counties whose income limits all parse to 0, batched into one event rather than one per row.
- **Dead `PARLER_CACHE_BACKEND` removed** — never read by parler 2.3, which imports the default cache directly in `parler/cache.py`.
- **A guard test that could not fail was replaced.** `parler.appsettings` reads `PARLER_ENABLE_CACHING` once at import, so `override_settings` never reaches it; the test passed even with the setting flipped to `True`. Now asserts directly against `parler.appsettings` and counts writes patched at `parler.models._cache_translation`.
- **Test helper moved** to `benefits/tests/cache_override.py` so pytest stops collecting it as an empty module, and `translations/tests.py` now imports it instead of keeping a duplicate.

**Test infrastructure**
- `redis:8` service container + `REDIS_URL` in `pr-validation.yaml`, so CI exercises `django_redis` instead of silently falling back to `LocMemCache`.
- New `benefits/tests/test_redis_backend.py` — runs against a real Redis, skips when unreachable. Round-trips, bulk `get_many`/`set_many`/`delete_many`, and asserts compression is actually applied.
- New `benefits/tests/test_cache_config.py` — pins `rediss://` vs `redis://` scheme handling and the connection ceiling, including a test that a plain pool *rejects* `ssl_cert_reqs`, documenting why the check exists.
- `translations/tests.py` (was empty boilerplate) — per-language keying, key versioning, `use_fallback` semantics, inactive-label exclusion, single-query rebuild, invalidation across all languages.
- New `benefits/tests/cache_override.py` — shared cache override, applied to the five test modules that assert on real cache reads/writes so they no longer depend on ambient `REDIS_URL`.
- `conftest.py` — autouse fixture clearing the cache between tests (Redis state outlives a test; `LocMemCache`'s did not).

### Measured impact

Verified on staging (v853, Standard-1X, 8,608 active labels), not just locally:

| | #1594 | this PR |
|---|---|---|
| `/api/translations/?lang=en-us` | 503 (`H12`, 30s) | **200 in 541 ms** |
| cold rebuild | ~22 s | **2.72 s** |
| warm read | never warmed | **0.010 s** |
| first read after dyno restart | full rebuild | **0.024 s** (cache survives) |
| dyno RSS for one rebuild | 357 MB (**70%** of 512 MB) | **187 MB (37%)** |
| Redis keys | 1 blob + ~156k parler | **18** |
| Redis dataset | 15.31 MB, over the 25 MB ceiling | **5.71 MB**, 0 evictions |
| per-request read | 15.31 MB | **0.38 MB** |

The RSS row is the ticket's actual premise: the old walk cost **+205 MB for a single rebuild**, so two concurrent rebuilds exceed a 512 MB quota — matching the `R14`/`R15` history. The new path costs **+35 MB**.

## Testing

- Migrations to run: **none** (this branch adds no migrations; it is rebased on `main`, so MFB-1019's `0163_drop_program_value_type` / `0158_drop_snapshot_value_type` are included and already applied in every environment)
- Configuration updates needed: **none for the app**. `REDIS_URL` is already set on both apps by the Heroku Redis addon. See Deployment for two addon-level items, one already applied.
- Environment variables/settings to add: none required. `REDIS_URL` is optional locally — without it you get `LocMemCache`. To match staging/prod constraints, run Redis as `redis-server --port 6379 --maxmemory 25mb --maxmemory-policy allkeys-lru`
- Manual testing steps:
  1. **Run `pytest --create-db` the first time.** `pytest.ini` sets `--reuse-db --nomigrations`, so a test DB built before MFB-1019 still has `value_type NOT NULL` and produces ~74 failures / ~187 errors that look alarming but are purely a stale schema.
  2. With Redis running, request `/api/translations/?lang=en-us` twice — the second should be milliseconds.
  3. `redis-cli -n 1 --scan --pattern 'benefits*'` should show 18 keys (`translation_data:v2:{lang}`), not one blob.
  4. `redis-cli info memory` — expect a few MB of dataset and `evicted_keys:0`.
  5. Edit a translation in the admin, re-request, and confirm the change appears (invalidation clears all languages).
  6. Restart the app and confirm the first request is still fast — the cache should survive.

Suite is green in all three cache configurations (re-measured after the rebase onto `main`, which brought in ~500 new tests):

| configuration | result |
|---|---|
| Redis up (matches CI) | **3114 passed**, 0 skipped |
| Redis configured but **down** | **3106 passed**, 8 skipped |
| no `REDIS_URL` (LocMem) | **3106 passed**, 8 skipped |

The 8 skips are `benefits/tests/test_redis_backend.py`, which skips itself when Redis is unreachable. These runs had `HUD_API_TOKEN` set locally, so the HUD integration tests executed rather than skipping.

Before this PR, 10 tests failed whenever `REDIS_URL` pointed at a Redis that was not running.

**Every new test was verified to fail when its fix is reverted**, not merely to pass. That mattered: one guard test in an earlier revision of this PR passed regardless of the setting it was supposed to guard. Sabotage runs confirm each test catches its regression.

Coverage added for code that had none:

| area | covers |
|---|---|
| `benefits/tests/test_cache_config.py` | `rediss://` vs `redis://` scheme handling, connection ceiling, and that a plain pool *rejects* `ssl_cert_reqs` |
| `benefits/tests/test_redis_backend.py` | real-Redis round-trips, bulk `get_many`/`set_many`/`delete_many`, compression actually applied (skips when unreachable) |
| `translations/tests.py` | per-language keying, key versioning, TTL stagger + cross-process stability, fallback semantics, single-query rebuild, invalidation, parler-avoidance with a control |
| `translations/tests.py` (new class) | view error path: warm languages survive a failed rebuild, failed rebuild caches nothing, no exception text in the body, no traceback outside DEBUG |
| `programs/tests.py` | FPL: no cache calls, stable identity, unknown period still raises `KeyError`, `get_limit` extrapolation, every year defines all 8 household sizes |
| `programs/programs/nc/nc_aca/tests.py` *(new)* | cache parsing, unknown county -> 0 + report, empty sheet cache reported |
| `programs/programs/co/nurturing_futures/tests.py` *(new)* | all-zeros not cacheable + reported, partial malformed kept + reported, good row silent |
| `programs/programs/co/rental_assistance_grant/tests.py` *(new)* | all-zero county reported, batched, partial not reported |
| `programs/programs/co/connect_for_health/tests/test_connect_for_health.py` | cache parsing, unknown county -> 0 + report, empty cache reported |

`nc_aca`, `nurturing_futures`, and `rental_assistance_grant` had **no test files at all** before this.

## Deployment

- Post-deployment scripts needed: **none**. The cache populates lazily on first request; a cold rebuild is ~2.7s on staging. No warm step required, and deliberately no cache-refresh-on-restart hook (see Notes).
- Production config updates:
  1. **`maxmemory-policy` → `allkeys-lru` — ALREADY APPLIED** to both `cobenefits-api-staging` and `cobenefits-api`. Both were `noeviction`, where a full cache *fails writes* instead of shedding old keys, which combined with `IGNORE_EXCEPTIONS` degrades silently. Verified nothing on these instances needs durability: sessions are database-backed (Django default, not cache), and there is no celery/rq/channels. Note DRF throttle counters do live in this cache and can now be evicted early under pressure — that is best-effort abuse prevention, and it was already failing to write under `noeviction`.
     ```
     heroku redis:maxmemory --policy allkeys-lru --app cobenefits-api-staging   # done
     heroku redis:maxmemory --policy allkeys-lru --app cobenefits-api           # done
     ```
  2. **Recommended: enable `log-runtime-metrics` on production.** Currently disabled on both apps, which means there is no dyno memory telemetry at all — the `R14` that motivated this ticket is invisible until something crashes. It is a free Heroku Labs flag (no addon, and neither app has a volume-priced log drain), reversible, and takes effect on the next dyno start. It was enabled on staging to produce the RSS figures above and has since been disabled again.
     ```
     heroku labs:enable log-runtime-metrics -a cobenefits-api
     heroku restart -a cobenefits-api
     ```
     Worth keeping on permanently rather than toggling per-investigation, since its value is catching memory growth *before* an incident. Reverse with `heroku labs:disable log-runtime-metrics -a cobenefits-api` + restart.
- Admin updates needed: none
- Notify team/users of: nothing user-facing. Note for whoever merges — merging triggers the staging auto-deploy, which will replace whatever is currently on staging. Staging is presently running this branch (v853) deployed directly for verification, so a merge simply makes that permanent.

## Notes for Reviewers

The branch is rebased on `main` and the commits are split on purpose:

1. **`Reapply "Replace in-memory custom Cache class..."`** — a clean revert-of-the-revert, restoring #1594 verbatim.
2. **`Reland Redis cache (MFB-468) with translation-endpoint fixes`** — **reviewing this commit alone shows exactly what changed relative to #1594.** Of #1594's 23 files, only 4 are modified.
3. **`addressing CR comments`** — the eight post-review fixes listed above.
4. **`Adapt CO test suites arriving from main to the new cache API`** — rebase fallout only, no production-code change. See below.

`programs/models.py` was the sole file both this work and MFB-1019 touched, and MFB-1019's changes are preserved verbatim (no `value_type` references reintroduced; `makemigrations --check` reports no drift).

### What the rebase had to resolve (commit 4)

`main` gained #1679 (`Replace skipped CO unit tests`) and siblings after this branch was cut, which added unit suites written against the **pre-reland** cache API. Adapting them is mechanical except for one behavioural case:

- **`FplCache` no longer exists** — commit 3 removed it, so four new suites (`cfhc`, `co_head_start`, `omnisalud`, `dental_health_care_seniors`) had a broken `from programs.models import FplCache` import that would have failed collection for the whole suite. They now read the `_FPL_DEFAULTS` constant that `FederalPoveryLimit.as_dict()` reads, preserving the original intent of importing the table rather than copying it.
- **`Cache.fetch()` → `GoogleSheetsCache.get_data()`** — `test_co_head_start.py` patched `"fetch"` (which `patch.object` rejects once the attribute is gone) and asserted on `counties.default` (now `_empty_fallback()`). The pinned *behaviour* is unchanged: a sheet outage still degrades to `{}` and makes every CO household ineligible.
- **`main` moved `connect_for_health/tests.py` → `tests/test_connect_for_health.py`**, a modify/delete conflict — and a module/package name collision, so the old `tests.py` could not be kept. Commit 3's `_process` row-parsing tests moved into main's package, and main's two cases pinning an uncovered county as a `KeyError` were rewritten: **that KeyError is the bug commit 3 fixes**, so they now assert the $0 value plus the `capture_message` report.

No other file changed on both sides.

- Known limitations:
  - **Redis is now the only cache layer in front of Google Sheets and Auth0, and this PR does not change that.** With `IGNORE_EXCEPTIONS`, an unreachable Redis makes `cache.get` a permanent miss, so every `GoogleSheetsCache.get_data()` hits the Sheets API and the PolicyEngine token fetch hits Auth0 on every call. There are **11 production `GoogleSheetsCache` subclasses**, several firing per screen, and the stale-key fallback is no help because it lives in the same Redis. Once Sheets rate-limits (300 reads/min/project), `_process` raises and `_empty_fallback()` returns `{}`, so programs go ineligible or $0. Not hypothetical: staging ran at `DBSIZE 0` for the entire #1594 window, which is exactly this state — it simply lacked the traffic to hit the quota. Pre-#1594 these were process-local reads, so a Redis outage structurally could not reach them. **Deliberately out of scope here** — the fix is a process-local layer in front of Redis for these specific caches and warrants its own ticket. The `capture_message` additions above at least make the resulting $0 values visible instead of silent.
  - **Cold rebuild under concurrent load is untested.** All rebuild timings were measured against a single dyno with no competing traffic. `WEB_CONCURRENCY=1` on staging with a single sync gunicorn worker is what turned slowness into an outage originally, since requests queue behind each other. 2.72s leaves a wide margin under the 30s router timeout, the cache now survives restarts, and the TTL jitter added in review means languages no longer all lapse at the same instant — but the concurrent case has not been reproduced, and jitter does not remove the cold-start case after a flush or a cache-version bump. Worth a short load test against `/api/translations/` immediately after a cache flush before assuming the failure mode is fully closed. Note production is **Standard-2X (1024 MB)** vs staging's Standard-1X (512 MB) and has no explicit `WEB_CONCURRENCY`, so its headroom differs from what was measured.
  - `IGNORE_EXCEPTIONS: True` is retained, so a Redis outage still degrades rather than erroring — it now logs. A boot-time cache health check would surface misconfiguration earlier; left out to keep the diff scoped.
  - CI's Redis is not memory-constrained to 25 MB. GitHub service containers cannot override the container command, and the test dataset is far too small to hit a ceiling. Sizing was verified locally against a mini-equivalent Redis and confirmed on staging.
  - Running the suite with `REDIS_URL` pointing at a Redis you also use for development will flush that cache database, since the isolation fixture calls `cache.clear()`.
  - Local memory figures proved only indicative, not accurate — local RSS baseline was 244 MB vs staging's 152 MB, and the per-rebuild deltas differed materially. Staging was the only place these could be measured honestly.
- Future considerations:
  - A full invalidate-and-rebuild transiently roughly doubles the Redis footprint (peak 15.03 MB observed on staging) because old and new key sets briefly coexist. Safe at 25 MB today; it is the real headroom constraint as more caches move onto this instance, not the 5.71 MB steady state.
  - `export_translations()` (`translations/models.py`) still uses the same per-language parler walk and will be slow for the same reason. Nothing user-facing calls it, so it is out of scope here.
  - Deliberately **not** added: cache refresh on dyno restart. Surviving restarts is the point of moving to Redis (verified: 0.024s first read after a restart), Heroku cycles dynos roughly daily, and a restart hook fires in every worker simultaneously — reproducing the concurrent-cold-rebuild pattern behind the historical R15 kills. The versioned cache key covers the legitimate case, which is format changes.
  - If more caches move onto this Redis, the mini plan's 25 MB and 20-connection limits are the next constraints.

## Summary by CodeRabbit

* **New Features**
  * Added Redis-backed caching with connection pooling, compression, retry handling, and graceful fallbacks.
  * Added resilient Google Sheets caching with stale-data and fallback support.
  * Improved translation caching with per-language entries and automatic invalidation.
  * Added safer handling for missing or invalid county data in benefit calculations.

* **Bug Fixes**
  * Prevented cache outages from disrupting eligibility and value calculations.
  * Improved malformed spreadsheet data handling and reporting.
  * Translation service errors now return a clear temporary-unavailable response.


SonaliBedge:
Thank you so much, @david-j-stoll, for working on this and fixing the issue. Everything looks good on my end  — reviewed the full diff and ran the test suite locally with pytest --create-db. All tests passed with no failures or errors (2546 passed, 44 skipped — skip count differs slightly from the PR's numbers, but that's expected local-environment variance like missing credentials for some integration tests, not an issue).
Verified against a real Redis instance (not LocMemCache) with the help of claude: traced the SSL→IGNORE_EXCEPTIONS root cause in redis-py/django-redis source, confirmed the assertNumQueries regression guards and the FLUSHDB claim in conftest.py, ran the full suite clean (2554 passed/0 failed/0 errors) on a fresh schema, and confirmed exactly 4 files changed relative to #1594 with no value_type reintroduced — LGTM.

catonph (approved):
Lets get merge conflicts sorted but otherwise approved!

catonph:
Two things from production profiling today that bear on this PR.

**1. `max_connections=15` is per-process, and prod now runs 2 workers.**

`CONNECTION_POOL_KWARGS` builds a `redis.ConnectionPool` per Python process, so the ceiling is `max_connections × gunicorn workers`. Confirmed the plan limit today:

```
$ heroku redis:info --app cobenefits-api
Plan:                   Mini
Maxmemory:              25 MB
Maxmemory Policy:       allkeys-lru
Plan Connection Limit:  20
```

The PR body notes production "has no explicit `WEB_CONCURRENCY`" — that changed today. It was defaulting to `(2 × CPU) + 1` = **4 workers**, and I set `WEB_CONCURRENCY=2` to stop unrelated R14 errors. So:

| | workers | × 15 | vs limit |
|---|---|---|---|
| before today | 4 | 60 | 3× over |
| now | 2 | 30 | 1.5× over |
| under limit | 2 | 2 × 9 = 18 | ok |

Pools are lazy so 30 is a ceiling rather than steady state, and a cache workload may never reach it. But with `IGNORE_EXCEPTIONS: True` a pool-exhaustion error surfaces as a silent cache miss — the same shape as the TLS bug this PR fixes. Might be worth deriving it, e.g. `max(2, 18 // int(os.environ.get("WEB_CONCURRENCY", 2)))`, or just dropping to 8.

**2. `Maxmemory Policy` is `allkeys-lru`, not `noeviction`.**

The body says mini is "a Redis that *fails writes* rather than evicting when full." It's actually `allkeys-lru` (see above), so it evicts instead of erroring. Lower risk than described — worth updating that paragraph since it's cited as a reason #1594 "could not have worked regardless."

---

**Separately — this PR likely fixes the results page too, and that isn't measured.**

All the timings here are `/api/translations/`. I profiled `/api/eligibility/` against production data today:

```
Results page: 7.10s total
  PolicyEngine API:       3.51s   (external; raised with PE separately)
  Google Sheets fetches:  3.30s   <- the caches this PR migrates
  Database (141 queries): 0.11s
```

Breakdown: `integrations/util/cache.py:47(fetch)` 3.285s over 26 calls, `sheets/sheets.py:35(data)` 3.263s over 6, `income_limits.py:45(update)` 1.02s. `denver_trash_rebate` alone is 1.02s via `get_screen_ami()`.

Those all route through `IncomeLimitsCache`, `CFHCache`, `BoulderAmiCache`, `RAGCache`, `LeapValueCache`, `CccapFplCache` — all migrated here. If they warm from Redis this should cut ~3.3s off the results page, roughly halving it. Worth measuring `/api/eligibility/` on staging alongside `/api/translations/` so the win is captured (and so a regression would be caught).

Also relevant to the concurrent-cold-rebuild caveat in the body: `WEB_CONCURRENCY=2` means 2 independent cold caches per restart under today's in-memory design, which strengthens the case for landing this.

PR #1670, opened 2026-07-30
MFB-1427: send already-received programs to BenBot, and stop sending it programs the user can't see

elliottwaissbluth:
# MFB-1427: send already-received programs to BenBot, and stop sending it programs the user can't see

## Context & Motivation

- Linear: [MFB-1427](https://linear.app/myfriendben/issue/MFB-1427/benbot-add-programs-already-receiving-as-context)
- Related PRs: MyFriendBen/ai-service#9 (consumes the new context) · MyFriendBen/benefits-calculator#2173 (sends what's on screen)

BenBot has been telling people to apply for benefits they already receive.

**The root cause is in this repo.** `screener/assistant.py:_build_context()` pulled every snapshot row where `eligible=True` and sent them all as `eligible_programs`. It never checked `already_has` — while the results page does (`filterPrograms.ts:isProgramBasicallyVisible` excludes already-received and $0-value programs). So BenBot always saw a strictly larger list than the user did, and since it may only recommend from the list it's given, that list has to equal what's on screen.

Enrollment lives in two places, which is why this was easy to miss: `CurrentBenefit` (household-level, `screen.has_benefit()`) covers most programs, while medicaid/CHP/medicare/VA/emergency-medicaid/family-planning are **member**-level in `Insurance.insurance_map()` and never appear in `CurrentBenefit` — deliberately, per the note in `serializers._derived_current_benefit_names`. Both are now handled.

## Changes Made

**`eligible_programs` now mirrors the results page**

Three gates on the snapshot filter, matching the frontend:

- `not screen.has_benefit(...)` — the actual MFB-1427 bug
- displayed value `> 0` — the page hides $0 programs. Safe in one direction only: snapshot value is `household_value + Σ all member values` and the page's `programValue()` is a subset of that same sum, so `0` ⟹ the page hid it too. (Programs displaying "Varies" already return a sentinel `1` for exactly this reason — `il/access_dupage/calculator.py:9-16`.)
- `not _has_member_insurance(...)` — closes the Medicaid-to-someone-already-on-Medicaid case. **Fallback-only:** the client list is authoritative precisely because per-member insurance runs client-side, and this gate is coarser (hides where the page reduces), so applying it on top of a client list would drop a program the user is looking at — the MFB-1427 failure in reverse.

**New `current_programs`**

Covers **both** enrollment systems — `CurrentBenefit` plus the insurance-derived names, unioned. Without the union, a household on Medicaid had it removed from `eligible_programs` *and* absent here, so ai-service's closed-world rule would forbid naming it at all: "why did my Medicaid renewal letter arrive?" got a refusal.

Includes benefits with no snapshot row (not modelled by the white label's calculators, or reported since the last snapshot). Carries only `external_name` and `name` — no value, no apply URL. Disjoint from `eligible_programs` by construction.

**New `visible_programs` (optional, from the client)**

Several results-page filters run **only** in the browser — citizenship/legal status, `excludes_programs` mutual exclusions, per-member insurance — and can't be reproduced from the snapshot. Rather than re-derive six filters in Python and watch them drift, the results page posts what it's rendering and we intersect.

This also closed a **live** bug beyond hide/show. Each entry carries the `value` as displayed (`programValue()`, which nets out already-covered members), preferred over the snapshot's all-members figure. Previously BenBot could quote a number higher than the card showed for a partially-enrolled household.

Handled as untrusted input throughout: names can only *narrow*; both the object and bare-string shapes accepted; bounded at 256 entries (with a `capture_message` when truncating, since truncation hides real programs); duplicate names are first-wins so a caller can't send a program twice to pick which value applies; bools/NaN/inf/negatives/out-of-range values fall back to the snapshot. Absent ≠ empty: omitting the key selects the server-side fallback, `[]` means the page is genuinely showing nothing.

**Security hardening on the client-supplied value.** `AssistantStartView` is `AllowAny`, the screen UUID is also the results-page URL, and ai-service resumes by `screen_uuid` and overwrites the stored context — so a third party who has seen a link can rewrite a live conversation's context. Two bounds close that:

- `value` is bounded by `min(client, snapshot)`. The page can only ever *reduce* the snapshot figure, so anything higher is definitionally not a displayed value.
- A name list that matches no eligible row falls back to the server-side filters. Without it, `visible_programs: ["zzz"]` empties the list and ai-service renders *and persists* "this person has NO eligible programs". `_visible_programs` already had this instinct at the parse layer; the intersection needed it too.
- Both endpoints now carry IP throttles (`assistant_start` 30/hr, `assistant_message` 120/hr) via the existing `HashedIPAnonRateThrottle`, matching the REM/Places proxies — same AllowAny-proxy shape, larger cost (60s worker hold plus a paid LLM).

**`results_url` now populated**

Never sent before. The assistant's guardrails offer "your results page" as the fallback when it has nothing to recommend, so on an empty eligible list the model had *no* legitimate exit — exactly when one invents a hotline. Same shape as the results email (`integrations/services/communications/message.py:110`), with `rstrip('/')` since `FRONTEND_DOMAIN` is env-supplied.

**Query, i18n and hardening fixes in the code this touches**

- `CONTEXT_PREFETCH` (`current_benefits__program`, `household_members__insurance`) on the start view. The enrollment checks run per program, and the insurance one also walks members plus each member's reverse OneToOne — an N+1 on two axes. Tests assert query count is flat in both program *and* member count.
- `_translated()` tries the **active** language first, then falls back to `LANGUAGE_CODE`. Both halves matter: non-default rows are created with `text=""` and because the row exists parler's `hide_untranslated` never fires (so a Spanish request would hand BenBot raw slugs), but pinning straight to `LANGUAGE_CODE` — as an earlier revision did — silently served everyone the English `apply_button_link` when state portals have `/es` pages.
- `except AttributeError` rather than `except Exception`, which was swallowing `OperationalError` and would silently degrade every name.
- `prefetch_related("name__translations")` — `Program.name` is an FK to `Translation`, but `Translation` is a parler model whose text lives in a separate table, so `select_related` alone still resolved `.text` per row.
- DB-sourced strings are whitespace-flattened and length-capped on the way out. They're interpolated into ai-service's *system* prompt and `Translation` rows are admin-editable; ai-service sanitizes again at render, this is the cheap choke point.
- Snapshot names fall back to `name_abbreviated` when blank/placeholder, otherwise the prompt renders `- (wa_snap)`.
- White-label scoped the `current_programs` query (with `.distinct()`, now that the `Q()` union can match by both arms).
- `_body(request)` normalizes once — `request.data` is a list for a JSON array body, so `.get` isn't safe to assume and an array body 500'd.
- `current_programs` sorts on `.casefold()`, since abbreviation fallbacks are lowercase and would otherwise sort after every translated name.

**Tests** — new `screener/tests/test_assistant_context.py`, 56 tests, including an `APITestCase` that drives the real endpoint.

## Testing

- Migrations to run: **none**
- Configuration updates needed: none. The `benbot` feature flag must be on for the white label you test with, or every assistant endpoint returns `403 assistant_disabled`.
- Environment variables/settings to add: none new. Two throttle rates added to `DEFAULT_THROTTLE_RATES` in settings.
- Manual testing steps:

```bash
pytest screener/tests/test_assistant_context.py -q   # 56 passed
pytest screener/ -q                                  # 341 passed
black --check . -l 120 --extend-exclude=".*/new_white_label/.*"
```

End-to-end. Run ai-service in live mode with `LOG_PAYLOADS=true`, then give a screen a current benefit via the existing toggle (needs `screener.change_screen`):

```bash
curl -X PATCH http://localhost:8000/api/screens/<uuid>/current-benefits/ \
  -H "Authorization: Token <token>" -H "Content-Type: application/json" \
  -d '{"name_abbreviated": "co_snap", "has": true}'
```

Open BenBot and check the logged `create_conversation` payload:

- the toggled program is **absent** from `eligible_programs`
- it's **present** in `current_programs`, with a name and no value/apply URL
- `visible_programs` arrived with a `value` on each entry
- `results_url` is populated

Then ask BenBot to apply for the program you toggled — it should decline and ask what changed.

Or drive Layer 1 directly:

```bash
curl -X POST http://localhost:8000/api/screens/<uuid>/assistant/conversations/ \
  -H "Content-Type: application/json" \
  -d '{"locale": "en-US", "visible_programs": [{"name_abbreviated": "co_wic", "value": 1224}]}'
```

Worth trying: omit `visible_programs` (server fallback) · send `[]` (empty eligible list) · send an ineligible name (dropped) · send `["zzz"]` (falls back, doesn't blank) · send a `value` above the snapshot's (clamped down).

## Deployment

- Admin updates needed: none
- Notify team/users of: nothing beyond the ticket

**Deploy ai-service#9 first.** It ignores `current_programs` if this ships alone, so the already-receiving context would be silently dropped — the filtering half would still work, the "answer questions about what they have" half wouldn't.

benefits-calculator#2173 can ship in either order: without it we fall back to the server-side filters, which is a correct-but-coarser list.

## Notes for Reviewers

**Where to look first:** `_build_context`, `_displayed_value` and `_visible_programs` in `screener/assistant.py`.

**The judgement call worth challenging** is trusting a client-supplied `value` at all, even bounded. The alternative is persisting a member-value breakdown on `ProgramEligibilitySnapshot` and reproducing `programValue()` server-side — correct, but a migration plus a change to the eligibility write path. I have a written plan and would rather do it as its own PR. Happy to be argued into doing it now.

**Known limitations:**

- The server-side insurance guard hides a program when *any* member is covered, where the results page shows it with a reduced value. A partially-enrolled household therefore loses a program it should have seen — but only on clients that don't send `visible_programs`, which today is nothing (web always sends it). It matters when a non-web channel ships (ADR-002).
- `visible_programs` is fixed when the conversation opens; a mid-session filter change applies on the next page load, where ai-service's context refresh picks it up.
- One pre-existing test, `test_has_benefits_programs.py::test_response_shape`, fails in my local sqlite-based sandbox (translation serialization). It fails in isolation on `main` too and is unrelated to this change — please confirm it passes in CI against postgres.
- The throttle rates are a first guess. 30 conversation-opens/hour per IP is generous for a person and tight for a script, but shared-NAT environments (libraries, shelters) are exactly our user base — worth a second opinion, and easy to raise.

**Future considerations:**

- Unifying `Insurance.insurance_map()` and `CurrentBenefit`. Two enrollment systems is the root cause of why `has_benefit()` silently misses Medicaid and why this bug was invisible. ~30 calculators read `has_benefit`/`has_base_benefit`, so it's a separate project.
- The results page's legal-status and `excludes_programs` filters are still unmatched server-side. `visible_programs` covers them for web; the fallback doesn't.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards to limit assistant session starts and messages.
  * Improved assistant responses with translation fallbacks, validated links, relevant program context, and personalized enrollment information.
  * Added filtering to show eligible, relevant programs while preserving current benefits and displayed values.
  * Improved handling of incomplete or malformed requests.

* **Bug Fixes**
  * Improved results links, insurance-based eligibility detection, and program ordering.
  * Reduced unnecessary data loading for assistant requests.


elliottwaissbluth:
## Review: MFB-1427 — already-received programs + visible_programs

Reviewed at high recall. The shape of the fix is right: `eligible_programs` has to equal what
the page renders, already-received programs have to survive into a separate list, and
intersecting a client-supplied list beats re-deriving six client-side filters in Python. The
`min(client, snapshot)` bound is sound — `programValue()` is `household_value + Σ member.value`
over uncovered members, and `eligibility.value` is the same sum over *all* members
(`programs/programs/calc.py:65`), so the client figure genuinely can only reduce.

Three things I'd fix before merge, then a set of smaller ones.

---

### Blocking

#### 1. `MAX_PROMPT_FIELD_LEN` truncates real apply URLs — `screener/assistant.py:108`

`_translated` is shared between program names and `apply_button_link`, so the 160-char prompt
cap lands on URLs too. Two links in the shipped config already exceed it:

| program | field | length |
|---|---|---|
| `tx_wic` | `apply_button_link` | 198 |
| `il_ibccp` | `apply_button_link` | 174 |

`tx_wic`'s link is `https://mywic.us/participantreferral?program=TX&_gl=1*...` — truncation cuts
mid-query-string and `apply_url` becomes a dead link. The prompt forbids the assistant from
offering any URL it wasn't given, so there's no fallback: the user gets a 404.

The cap is defensible for names. For a URL, drop it rather than truncate it — a link that's too
long to send is a link we don't send:

```python
link = _translated(program.apply_button_link, max_len=None)
if link and len(link) <= MAX_URL_LEN:
    urls[program.name_abbreviated] = link
```

#### 2. `insurance_map()` doesn't cover the state Medicaid names — `screener/assistant.py:167`

Both new helpers derive their candidate set from `Insurance.insurance_map()`'s keys. Those keys
are `medicaid`, `nc_medicaid`, `co_medicaid`, `wa_apple_health_medicaid`,
`wa_apple_health_for_kids`, `ks_chip`, `ma_mass_health`, `medicare`, `emergency_medicaid`,
`family_planning`, `va`.

Programs that exist and are **not** in that list:

- `tx_medicaid`, `tx_chip`, `tx_medicaid_for_parents_and_caretakers`, `tx_medicaid_for_pregnant_women`
- `ks_medicaid` (note `ks_chip` **is** in the map — this reads as an oversight, not a decision)
- `il_medicaid`, per `programs/programs/helpers.py:3`:
  `STATE_MEDICAID_OPTIONS = ("co_medicaid", "nc_medicaid", "il_medicaid", "ks_medicaid")`

So for a Texas or Kansas household already on Medicaid,
`_has_member_insurance(screen, "tx_medicaid")` hits `strict=False` and returns `False`.
`tx_medicaid` stays in `eligible_programs` **with an apply URL** — the original MFB-1427 bug,
unfixed — and it's absent from `current_programs`, so the closed-world rule also forbids the
assistant from discussing it.

The docstring's claim that deriving from `insurance_map` means it "can't go stale" is already
false at merge time. At minimum, cross-check the candidate set against `STATE_MEDICAID_OPTIONS`
and fail loudly on a program that looks like Medicaid but isn't mapped.

#### 3. The two lists are not disjoint — `screener/assistant.py:324` / `:204`

The insurance gate is skipped whenever the client sends a list (line 324, correctly — it's
coarser than the page). But `_current_programs` unions the insurance-derived names
*unconditionally* (line 204). For a partially-enrolled household:

- CO household, 3 members, one already on Medicaid
- page renders `co_medicaid` at a reduced value → posts it in `visible_programs`
- `eligible_programs` contains `co_medicaid` **with `apply_url`**
- `current_programs` contains `co_medicaid` via the insurance union

ai-service is told "you may recommend co_medicaid, apply here" and "they already receive
co_medicaid" in the same payload. The PR description states the lists are "disjoint by
construction"; they aren't on the path that every web client takes.

`test_member_insurance_gate_does_not_override_a_client_list` builds exactly this state and
asserts only the eligible side. Either subtract the eligible names from `current_programs`, or
withhold `apply_url` for insurance-held programs — and assert the disjointness in that test.

---

### Correctness, non-blocking

#### 4. `except AttributeError: return ""` defeats the fallback it documents — `:102`

`PARLER_LANGUAGES['default']['hide_untranslated'] = True` (`benefits/settings.py:285`), so
parler does **not** fall back on attribute access — `get_active_choices()` returns only the
current language. Reading `.text` for a language with no row raises `TranslationDoesNotExist`,
which subclasses `AttributeError` (`parler/models.py:117`).

So the `except` fires on the *first* loop iteration and returns `""` without ever trying
`settings.LANGUAGE_CODE` — the exact fallback the docstring calls necessary. It only bites when
a row is genuinely absent (a `Translation` created before a language was added to
`PARLER_LANGUAGES`, or created outside `add_translation`), but then every name degrades to its
abbreviation and every `apply_url` is dropped for that whole conversation.

Only the `None`-FK case warrants `return ""`. Split them:

```python
if translation is None:
    return ""
...
except TranslationDoesNotExist:
    continue
```

#### 5. Language comes from `Accept-Language`, not from the screen — `:96`

`get_language()` under `LocaleMiddleware` (`benefits/settings.py:156`) reflects the browser's
`Accept-Language` header, not the language the user picked in the app.

A user on English in MFB with a Spanish browser gets Spanish names and Spanish `/es` apply
links, while `payload["locale"]` is `"en-US"` — the frontend passes `locale: undefined`
(`apiCalls.ts:387`). BenBot then answers in English and quotes Spanish program names. The
reverse case serves English names into a Spanish conversation.

`Screen.request_language_code` / `Screen.get_language_code()` (`screener/models.py:383`) already
holds the right answer and is what `MessageUser` uses for the results email. Thread it through
as `_translated(..., language_code=screen.get_language_code())`.

#### 6. Fallback discards a valid client list once every row gates out — `:333`

The guard tests `not rows` — the list *after* all server gates — rather than "no client name
matched an eligible row".

Two tabs open. Tab A's results page is showing exactly one program, `snap`. In tab B the
household toggles "I already receive SNAP" (writes `CurrentBenefit`; the snapshot rows still
have `eligible=True`, since `already_has` is a display flag). Tab A's chatbot opens:

- `visible = {"snap"}` — matches an eligible row, so the list is *not* junk
- `not screen.has_benefit("snap")` drops it → `rows == []`
- `any(p.eligible ...)` is True → fallback replaces the list with every server-filtered eligible
  row, **including programs the page hid for legal status and `excludes_programs`**

That's the MFB-1427 failure mode the client list exists to prevent. Test the intersection, not
the outcome:

```python
matched = {p.name_abbreviated for p in all_rows if p.eligible} & set(visible)
if visible and not matched:
    ...  # genuinely malformed input
```

#### 7. The bounds stop inflation but not narrowing — `:275`

`min(client, snapshot)` and the name-intersection stop a third party from making a victim's
assistant quote a *larger* number. Nothing stops them shrinking it.

Anyone who has seen a results link can POST
`{"visible_programs": [{"name_abbreviated": "co_snap", "value": 1}]}` — both fields "legal" per
the bounds. Per the PR description, ai-service resumes by `screen_uuid` and overwrites the
stored context, so the victim's live conversation now believes they qualify for one program
worth $1/yr. The 30/hr throttle bounds volume, not the single write.

I don't think this blocks — but it's the half of the AllowAny exposure the bounds *don't* cover,
and it deserves an explicit decision (ownership check on start, or write-once context per
conversation) rather than being folded into the deferred value-breakdown work.

---

### Cleanup

| # | Where | Issue |
|---|---|---|
| 8 | `:340` | The fallback comprehension re-states the primary one's predicates by copy-paste, and they already differ (`visible is not None or not _has_member_insurance(...)` vs bare `not _has_member_insurance(...)`). Two places, 20 lines apart, must stay in sync by hand, and no test compares them. Extract `_passes_server_gates(p, values, screen, *, apply_insurance_gate)`. |
| 9 | `:250` | `_has_member_insurance` re-walks every member and rebuilds `insurance_map()`'s 16-key dict per program, duplicating what `_insurance_program_names` just did. 80 programs × 5 members = 400 dict builds per open, on top of 12 × 5. Compute the held set once and test membership. Same fix removes the drift risk in #2. |
| 10 | `:234` | `_has_member_insurance` / `_insurance_program_names` are Screen-level enrollment predicates living in a views module, next to a model that already has `has_benefit`, `has_base_benefit`, `has_insurance_types` and `HouseholdMember.has_insurance` — the same one-liner at member level. CLAUDE.md: *"Fat models, skinny views — Business logic belongs in models and managers, not views."* The next caller is the results serializer at `screener/views.py:547`, computing per-member `already_has` for the same purpose. |
| 11 | `:253` | CLAUDE.md: *"Type hints everywhere — All functions and methods should have type annotations."* `_displayed_value(row, ...)`, `_translated(translation, ...)` (`:76`), `_clean_value(value)` (`:451`), `_body(request)` (`:498`) annotate returns only. |
| 12 | `benefits/settings.py:91` | Two new throttles on user-facing endpoints, no test that either engages or returns 429 — and the rates are self-described as a first guess. Note throttle history is cache-backed and shared across the test session (LocMemCache, `settings.py:334`), so such a test needs an explicit `cache.clear()`; `AssistantStartViewTests` already spends 6 of the 30/hour budget, so it'll start failing on its own as cases are added. |
| 13 | `:520` | `white_label` isn't `select_related` but is dereferenced twice (feature flag, `results_url`) — one avoidable query per open. Adding it would break `assertNumQueries(9)` (`test_assistant_context.py:700`); a hardcoded absolute count makes an unambiguous improvement look like a regression. Prefer `<=` or keep to the flatness assertions. |
| 14 | `:164` | `members = list(screen.household_members.all())` is materialized only for an early return that `screen.has_insurance_types` already handles — it's `any(...)` over the same relation (`screener/models.py:289`). Both lines can go. |
| 15 | `:34` | `from .views import HashedIPAnonRateThrottle` pulls the PolicyEngine registry, urgent-need functions, serializers and webhooks into `assistant` for one 6-line class, and makes any future `views → assistant` reference a startup import cycle. There are now four subclasses across two modules; `screener/throttles.py` holding the base plus all four scopes removes the coupling. |

---

### On the judgement call you flagged

Trusting a bounded client `value` reads as the right trade to me, and the bound is provably safe
in the direction you claim. The gap isn't the value — it's #7 (narrowing isn't bounded at all)
and #3/#6 (the client list and the server gates interact in two places where the result isn't
what the docstrings say). Those are cheap to close now, and closing them makes the deferred
member-breakdown migration a genuine follow-up rather than a prerequisite.

Not re-litigating the deploy order — ai-service#9 first is right for the reason you give.

PR #1673, opened 2026-08-04
MFB-1581: Add PolicyEngine VCR spec-scenario test harness

david-j-stoll:
## Context & Motivation

Program calculators are getting tests that are 1:1 with their `spec.md` Test Scenarios — one test per scenario, asserting both eligibility and benefit value. Custom (MFB) calculators can be tested directly, because the rules and dollar amounts are our code. PolicyEngine calculators can't: the answer comes from PolicyEngine over HTTP, so asserting it either means hitting the network on every test run or not testing it at all.

This generalizes the existing VCR harness (built for the HUD client) to PolicyEngine. PE program tests record a cassette live once when the program is implemented, then replay it in CI — fast, deterministic, no network — so CI protects our half of the contract (request assembly, wiring, overrides) without being at the mercy of PolicyEngine uptime.

- Linear: [MFB-1581](https://linear.app/myfriendben/issue/MFB-1581/policyengine-vcr-spec-scenario-test-harness)
- Delivers the **record** and **store** halves of [MFB-1565](https://linear.app/myfriendben/issue/MFB-1565/test-machinery-pe-cassette-refresh-policy-ci-vcr-mode-enforcement), which has been rescoped to the refresh policy that remains (see Notes for Reviewers)
- Partially advances [MFB-1564](https://linear.app/myfriendben/issue/MFB-1564/test-machinery-unified-pytest-suite-custom-unit-pe-replay-per-program) — the PE replay half only
- Related: the `add-pe-program` skill in `team-claude-config` was updated to use this harness ([team-claude-config#24](https://github.com/MyFriendBen/team-claude-config/pull/24), separate repo)

## Changes Made

**`conftest.py`** — six changes to the shared VCR harness:

- **Scoped the `HUD_API_TOKEN` collection skip to HUD tests.** It previously skip-marked *every* `@pytest.mark.integration` test, so PolicyEngine tests would have been silently disabled locally and on fork PRs while still reporting green. Now keyed on test location / use of the `integration_requires_token` fixture.
- **Added a `policy_engine_body` matcher.** Every PE call goes to the same URL (`POST household.api.policyengine.org/us/calculate`) with the household and pinned version in the body. Matching only on method/path would make a cassette replay its recorded response for *any* payload, so a regression in how we assemble the request would replay the old answer and still pass. The matcher is a no-op for other hosts, so existing HUD cassettes keep matching exactly as before.
- **Added the auth0 host to `ignore_hosts`.** The token exchange response body is a live bearer token and must never be written to a cassette. Recording fetches a token live (passing straight through VCR); every other run pre-seeds a placeholder instead.
- **Discard whatever a failing test recorded.** A bad recording (wrong credentials, a 4xx, an unserved PE version) would otherwise leave a cassette holding the error response, which then replays forever and reads as a code failure. `auto_vcr` snapshots the cassette, and on teardown restores it — or deletes it, if the test created it — when the test body failed. VCR's own `record_on_exception=False` is set too but cannot do this job: it fires from `Cassette.__exit__` keyed on an exception propagating out of the `with` block, and pytest never throws a test failure into a fixture generator, so for a cassette opened inside a fixture it never triggers. A new `pytest_runtest_makereport` hookwrapper stashes the phase reports the fixture needs to tell a pass from a failure.
- **Class-qualified cassette filenames** (`<TestClass>.<test_name>.yaml`). The bare test name meant two identically-named methods in different classes in the same module would share one cassette and silently replay each other's recording.
- **One reader for `VCR_MODE`: `vcr_record_mode()`.** `auto_vcr` and the PolicyEngine helpers each parsed the variable themselves, which let them disagree about what mode a run is in. It now also **rejects** an unrecognized value instead of silently falling back to `once`: the modes differ in whether they touch the network, so substituting one for another turns a typo into a behavior change — mistyping `none` is the case that matters, since the run was meant to be strict offline replay. Validated in `pytest_configure` and raised as a `UsageError`, so a bad value fails the session up front with `ERROR: Unrecognized VCR_MODE 'nonw'. Expected one of: none, new_episodes, all, once.` rather than erroring per-test. Unset or empty still means `once`.

**`programs/programs/policyengine/tests/integration_test_helpers.py`** (new) — helpers for running a PE calculator end to end in a test:

- `make_screen` / `add_member` / `add_income` — build a household with **explicit primary keys**
- `pin_pe_version` / `seed_pe_token` — keep the recorded request reproducible and the auth call out of the test. `seed_pe_token` writes the Django cache key `engines` reads (`_PE_TOKEN_CACHE_KEY`); the autouse `clear_cache` fixture empties it between tests, so the base class re-seeds per test
- `make_program` — the `Program` + FPL year row a PE calculator needs for `program.year.period`
- `calc_pe_program` — runs one calculator end to end with exactly one POST. Deliberately bypasses `calc_pe_eligibility`, which catches every exception and returns an empty result — that would turn a cassette miss into a confusing empty-dict failure instead of VCR's own error. It does keep that function's `can_calc()` gate, raising if the calculator can't calc: production drops such a calculator *before* building a payload, so a scenario asserting on `calc()` output there would be asserting against a path the screener never takes.
- `screener_value` — truncates to whole dollars the way the API does
- `PeIntegrationTestCase` — base class that pins the version, requires `pe_version` be declared, and carries `pytestmark = pytest.mark.integration` so a subclass that forgets the decorator still gets VCR rather than silently calling PolicyEngine live on every run
- **Recording is opt-in via `PE_RECORD=1`.** Inferring it from "credentials present and the mode isn't `none`" meant an ordinary local run — default mode `once`, cassette already present, nothing to record — still fired a live auth0 token exchange for anyone with PE credentials in `.env`, since `ignore_hosts` passes that request through to the network. Now nothing authenticates unless the run asks to record.
- **These tests skip under `VCR_MODE=all`.** That mode never replays, and these cassettes pin an exact model version that PolicyEngine stops serving the moment it promotes past it (`422 unsupported_version`) — so re-recording them on a release could only ever fail. Confirmed against live PE, which had already moved past the original `1.779.3` pin two days after it was recorded. Live re-runs belong in the scheduled drift job, which bumps the pin deliberately.

**`programs/programs/policyengine/tests/test_integration_helpers.py` + cassette** (new) — covers the harness itself with a member-level program, which is what exercises the id-keyed response path. A program's own PolicyEngine tests live next to the program, not here. Pinned at `1.784.3`, a version PolicyEngine currently serves, so the committed cassette can actually be refreshed; the asserted value is unchanged from `1.779.3`.

The helpers are deliberately named for what they do rather than for `spec.md`: they build a household, run one calculator, and hand back an `Eligibility`. Mirroring a program's Test Scenarios 1:1 is a convention of the callers (and of the `add-pe-program` skill), not something the helpers encode — so the naming doesn't have to change if that convention does.

**Renamed the 17 existing HUD cassettes** to the class-qualified scheme. Pure renames — no content change (git reports them as renames), and the suite passing under `VCR_MODE=none` proves each one still maps to its test, since that mode cannot record.

**`docs/TESTING.md`**:

- New "PolicyEngine Spec-Scenario Tests" section: test skeleton, the three replayability rules, record/verify commands, and how to refresh cassettes when PolicyEngine moves — including that refreshing a cassette and adopting a new PE version are always the same act.
- Corrected the CI mode table. It claimed push-to-main runs `VCR_MODE=all`; `deploy-staging` actually runs `new_episodes`, and only `deploy-production` (on release) uses `all`. Added that no workflow commits cassettes, so a CI run in a recording mode does not refresh the repo — it only changes what that run tests against.

## Testing

- Migrations to run: none
- Configuration updates needed: none. `PolicyEngineConfig` is pinned per-test by the helper, not globally
- Environment variables/settings to add: none to replay. **Recording** needs `PE_RECORD=1` plus `POLICY_ENGINE_CLIENT_ID` / `POLICY_ENGINE_CLIENT_SECRET` in `.env` (already required for local PE work)
- Manual testing steps:

```bash
# Full suite, strict replay — no network, no credentials
VCR_MODE=none venv/bin/pytest -q

# The PolicyEngine harness test on its own
VCR_MODE=none venv/bin/pytest programs/programs/policyengine/tests/ -v

# Existing HUD cassettes still replay under the new matcher and new filenames
VCR_MODE=none venv/bin/pytest integrations/clients/hud_income_limits/tests/ -q

# Optional: re-record the harness cassette live (needs PE credentials)
rm programs/programs/policyengine/tests/cassettes/*.yaml
PE_RECORD=1 VCR_MODE=once venv/bin/pytest programs/programs/policyengine/tests/ -v
```

Results on this branch (rebased on main): full suite **3201 passed** in strict replay mode; `black --check` clean. Recorded live at PE `1.784.3`, then verified replaying with `VCR_MODE=none`. Every valid `VCR_MODE` was also run against the harness directory — `none`/`once`/`new_episodes` all green, `all` skips the PE test as designed.

Each guard was verified by reproducing the failure it prevents:

| Scenario | Behavior |
|---|---|
| Household perturbed by one year of age | clean body-mismatch error naming the matcher, not a silent stale replay |
| Cassette deleted, recording fails (no credentials) | test fails, **no cassette written** |
| Cassette present, `VCR_MODE=all` | test skips, **cassette byte-identical** |
| `PE_RECORD=1` record against live PE | cassette written and retained, replays with no network, contains no token |
| `integration` decorator removed from the subclass | still collected by `-m integration` via the base class |
| Default local run with PE credentials in `.env` | `_can_record()` is `False` — no auth0 call |
| `VCR_MODE=nonw` (typo of `none`) | session fails immediately: `ERROR: Unrecognized VCR_MODE 'nonw'` |

## Deployment

- Admin updates needed: none
- Notify team/users of: two workflow rules for anyone writing PolicyEngine tests — run them with `pytest`, **never** `venv/bin/python manage.py test` (VCR is a pytest fixture, so under the Django runner the tests call PolicyEngine live on every run), and commit the cassette alongside the test. Recording now also needs `PE_RECORD=1`. All three are documented in `docs/TESTING.md`. Also note cassette filenames now carry a class prefix, so any local un-committed cassettes need renaming — and that a misspelled `VCR_MODE` is now a hard error instead of a silent fall back to `once`.

### Addressed from the first review round

- **`PrivateApiSim.token` no longer exists** — #1663 moved the bearer token to `django.core.cache`. This wasn't cosmetic: rebasing on main made the harness test fail outright with `AttributeError`. `seed_pe_token` now seeds `_PE_TOKEN_CACHE_KEY`.
- **Renamed the helper module** away from "spec": `spec_scenarios.py` → `integration_test_helpers.py`, `test_spec_scenarios.py` → `test_integration_helpers.py`, `PeSpecScenarioTestCase` → `PeIntegrationTestCase`, and the harness cassette renamed to match its class. Prose reworked so the helpers read as generally useful rather than tied to `spec.md`. `team-claude-config#24` is updated in step.
- **`VCR_MODE` is read in one place** (`vcr_record_mode()`), now rejecting unrecognized values. Worth noting the divergence as originally described wouldn't have been fixed by sharing the helper alone: coercion sends an unknown mode to `once`, which is itself recording-capable, so `_can_record()` stayed `True` either way — only refusing the value closes it.
- **`eligibility.value` is always numeric** — it's a computed property summing `household_value` and each member's value, and `screener/views.py` calls `math.trunc` on it unconditionally. No string reaches it; a "Varies"-style override would be applied downstream of this repo.

Worth focusing on:

- **The body matcher's failure mode.** Person keys are compared as recorded, deliberately un-normalized. Normalizing ids would let a cassette recorded under different pks "match" and then blow up with a `KeyError` when the response is read back — a clean "no matching request" error is the more honest signal. This is why the helpers require explicit primary keys.
- **The failed-recording discard.** It lives in `auto_vcr`'s teardown rather than in VCR's config, for the reason above — worth a look at whether tracking the outcome through `pytest_runtest_makereport` is the shape we want, since it's now load-bearing for both integrations rather than PE-specific.
- **`VCR_MODE=all` skipping PE tests.** The alternative was leaving `deploy-production` to re-record pinned cassettes live, which fails by construction rather than reporting drift. Skipping keeps the release signal honest but means nothing checks PE drift until the scheduled job exists (MFB-1565).
- **The HUD cassette renames** are the largest part of the diff by file count and the least interesting: no content changed. `git diff --stat -M` shows them as renames.
- **`ignore_hosts` for auth0.** Please sanity-check that the committed PE cassette contains exactly one interaction and no token: `grep -iE "authorization|access_token|Bearer" programs/programs/policyengine/tests/cassettes/*.yaml` should return nothing.
- **The truncation choice.** `screener_value()` applies `math.trunc` because PE returns fractional dollars (`12076.413`) and `clean_program` in `screener/views.py` truncates before serving. This keeps `spec.md` expectations in whole dollars — the number a QA run or a user sees — at the cost of not catching sub-dollar drift.

Known limitations:

- **CI does not block on a missing cassette.** PR validation and `deploy-staging` run `VCR_MODE=new_episodes`, so an un-recorded scenario is recorded live in CI rather than failing — and with no PE credentials in CI secrets that recording gets a 401, so it fails for the wrong reason (though the discard now keeps the bad recording out of the way). Flipping the PR job to `VCR_MODE=none` would make cassettes a genuine merge gate. Deliberately not changed here; it's the main remaining scope of [MFB-1565](https://linear.app/myfriendben/issue/MFB-1565/test-machinery-pe-cassette-refresh-policy-ci-vcr-mode-enforcement).
- **No drift detection yet.** With PE tests skipping under `all`, nothing re-runs them against live PolicyEngine until the scheduled job lands. That job is where a changed PE answer should surface as a triage signal rather than a failed deploy.
- The pinned `pe_version` must be a version PolicyEngine currently serves — it rejects exact versions other than what `current` / `frontier` resolve to. So a cassette can't be re-recorded at its original pin once PE promotes past it: refreshing a cassette and adopting a new PE version are always the same act.

Future considerations:

- **The scheduled live drift job isn't here yet.** The batched non-blocking job that re-records every cassette and surfaces both "a value changed on the same pin" and "the pinned version is behind" still needs standing up (M6).
- Each program's spec-scenario suite lands with its own program PR, contributing its cassettes to that eventual batch.
- MFB-1564 still needs the unified custom+PE suite, the `@pytest.mark.pe` marker naming (this PR reuses `integration`), and the `--live [--program X]` surface.


programs/programs/policyengine/tests/integration_test_helpers.py
```diff
@@ -0,0 +1,286 @@
...
+
+def screener_value(eligibility: Eligibility) -> int:
+    """The whole-dollar value the screener reports for this eligibility.
+
+    PolicyEngine returns fractional dollars (a member value can be 12076.413) and the API
+    truncates before serving it (``screener/views.py`` ``clean_program``). Asserting the
+    truncated value keeps spec.md expectations in whole dollars — the number a QA run or a
+    user actually sees — instead of encoding sub-cent noise into every test.
+    """
+    return math.trunc(eligibility.value)
```
catonph, line 253: Just want to confirm eligibility.value is always a number - a string override comes in later (e.g. "Varies")?

programs/programs/policyengine/tests/spec_scenarios.py
```diff
@@ -0,0 +1,286 @@
+"""Support for PolicyEngine spec-scenario tests.
```
catonph: Kind of a nit but can we rename this file? The name makes me expect to see spec scenarios in here - probably something with integration_test_helpers. Because also wary to couple this too tightly to the term "spec" in case that changes down the line

catonph: These helpers should help build tests regardless of where they come from  (probably should update the comments too to be a bit more neutral?)

programs/programs/policyengine/tests/spec_scenarios.py
```diff
@@ -0,0 +1,286 @@
...
+    fully-recorded suite fire a live auth call for no benefit.
+    """
+    if os.getenv(PE_RECORD_ENV_VAR, "").lower() not in _TRUTHY:
+        return False
+
+    has_credentials = bool(config("POLICY_ENGINE_CLIENT_ID", default="")) and bool(
+        config("POLICY_ENGINE_CLIENT_SECRET", default="")
+    )
+
+    return has_credentials and os.getenv("VCR_MODE", "once").lower() != "none"
```
catonph: `_can_record()` reads `VCR_MODE` directly while `auto_vcr` coerces invalid values to once. So `VCR_MODE=typo` + `PE_RECORD=1` authenticates live while the cassette replays. A shared `_vcr_mode()` helper would help ensure these don't diverge

programs/programs/policyengine/tests/spec_scenarios.py
```diff
@@ -0,0 +1,286 @@
...
+    never written to a cassette (its response body is a live token).
+
+    Otherwise seed a placeholder so nothing tries to authenticate at all. The cache is an
+    in-process class attribute on ``PrivateApiSim`` (not the Django cache), so this holds for
+    the rest of the test process.
+    """
+    if _can_record():
+        return
+
+    token_cache = PrivateApiSim.token
```
catonph: This got deleted in the redis PR

catonph (changes requested):
Biggest this in the deletion of PrivateApiSim

david-j-stoll:
> Biggest this in the deletion of PrivateApiSim

Addressed all of the above

catonph (approved):

PR #1674, opened 2026-08-04
MFB-1571: fix WIC income-blindness (PE income sources + reported SNAP/TANF receipt) and the dead has_benefit() base-name reads

david-j-stoll:
## Context & Motivation

- Linear: **MFB-1571** — WIC calculators are income-blind: the federal `Wic` sends none of PolicyEngine's WIC income sources
- **Sequence: #1659 (MFB-1382) → this PR → MFB-1312.** Merge after #1659.
- **This PR is now also the `has_benefit()` cleanup vehicle.** It was stacked on #1659's original branch, which carried both income capture and a set of dead-read fixes found while verifying the SNAP receipt path. #1659 has since been narrowed to income capture alone, so that cleanup now lives here permanently rather than dropping out on merge — see "Cleanup work carried here" below. MFB-1571 was retitled to match.
- `main` is already merged in here, because the MO programs (#1666/#1668/#1669/#1671) landed after this branch was cut and this change edits `MoWic`.
- Related: **MFB-1213** (MO WIC), which is where the bug surfaced

PolicyEngine decides WIC with

```
is_wic_eligible = demographic_eligible & (meets_income_test | meets_categorical_test) & nutritional_risk
```

and neither inner branch was fed.

**Income.** `Wic`'s only income input was `school_meal_countable_income`, which WIC's tree never reads — `wic_countable_income` sums its own parameter list, `gov.usda.wic.income.sources` (24 variables), and the school-meals variable is not among them. Given none of its sources, PE substitutes an imputation and WIC returns eligible at *any* reported income; verified live at $108k/yr. `mo_wic` shipped a partial fix (`irs_gross_income` on the subclass only); `co_wic` / `nc_wic` / `ma_wic` / `tx_wic` / `il_wic` were all still affected.

**Categorical.** `meets_wic_categorical_eligibility` reads `receives_snap` / `receives_tanf` for *reported* enrollment, and nothing in our codebase populated either. So a household that told us it receives SNAP could still fail the categorical test whenever PE's own SNAP determination disagreed — state rules PE doesn't model, or income risen since enrollment.

### Two corrections to the ticket's analysis

Both were checked against the live private API (pregnant adult + 3yo + infant, MO, 2026-08-04) rather than by reading names.

1. **The ticket says `irs_gross_income` covers "5 of 24" sources and that three of its fields are inert for WIC. Neither holds.** PE resolves several sources through `adds` chains, so the bundle already reached **8**. Each of these moves `wic_countable_income` by the amount sent:

   | Sent | Reaches | Screener types |
   | -- | -- | -- |
   | `taxable_pension_income` | → `pension_income` | pension, veteran |
   | `taxable_ira_distributions` | → `taxable_retirement_distributions` → `retirement_distributions` | deferredComp |
   | `long_term_capital_gains` | → `capital_gains`, which `wic_countable_income` adds as its own term | investment |

2. **The reported-receipt gap is better closed with `receives_snap` / `receives_tanf` than with the `snap` amount input.** After #1659 removed the `1` sentinel, `spm.Snap` sends a dollar amount or nothing — and most households can't give one, since SNAP lands on EBT monthly. The two `receives_*` variables are plain booleans with no formula, read only by adjunctive/categorical tests, so they carry the receipt signal without pinning any computed benefit. Measured: at $108k/yr, WIC flips from $0 back to eligible with either set to True.

Also measured: the ~$3,505 phantom income PE imputes for a $0-income MO household is **TANF**, not SSI — sending `ssi: 0` changes nothing, `tanf: 0` zeroes it.

## Changes Made

- **New `wic_income` bundle** (`dependencies/__init__.py`) — every source in `gov.usda.wic.income.sources` the screener actually collects: `irs_gross_income` plus `workers_compensation`, `alimony_income`, `miscellaneous_income` (gifts), `child_support_received`, reported `ssi`, and reported `tanf`. The comment there records the `adds` chains, so the next reader doesn't re-derive them.
- **`ChildSupportReceivedDependency`** (`dependencies/member.py`) — child support *received*, as income. Distinct from the existing `SnapChildSupportDependency`, which sends child support *paid* as an expense; both are asserted side by side in tests so an edit can't collapse them.
- **`ReceivesSnapDependency` / `ReceivesTanfDependency`** (`dependencies/spm.py`) — reported receipt as a boolean, resolved via `has_base_benefit` so the prefixed names white labels ship (`co_snap`, `ma_tafdc`, …) count. Complements `Snap`/`Tanf` rather than replacing them: those need an amount, these fire on the receipt signal alone.
- **Federal `Wic`** now sends `wic_income` + the two receipt inputs, and no longer sends `school_meal_countable_income`. All six state WIC calculators inherit the fix. The class docstring explains both branches and warns that the practical eligibility boundary is the state's Medicaid limit, not 185% FPL.
- **`MoWic`** trimmed to just `MoStateCodeDependency` — its partial fix and ~40-line gap note are superseded.
- **`MaWic`** now sends `MaStateCodeDependency`. It was the only WIC subclass sending no state code, and worked only because a sibling MA program put the state in the shared payload. WIC's FPG table branches on AK/HI vs. contiguous US, so it shouldn't be inherited by luck.
- **Migration** `0165_backfill_wa_snap_base_program` (renumbered from `0164`, re-pointed at `0164_seed_mo_generic_referrers` after merging `main` produced two leaf nodes, which makes `manage.py migrate` refuse to run at all). Sets `base_program="snap"` on `wa_snap`, the only SNAP variant with a NULL `base_program` — **confirmed against staging (43 rows) and prod (41 rows): it's the only NULL in the `snap`/`tanf`/`wic`/`ssi`/`ssdi`/`section_8` group in both.** This migration now belongs solely to this PR; #1659 no longer carries one.
- **Tests** — new `federal/pe/tests/test_wic.py` (15 tests) is the authoritative coverage: it pins our mapping against PE's full 24-source list, accounting for the `adds` chains, and asserts *every* state WIC calculator inherits the bundle and sends a state code, enumerated from the registry so a new state program is covered the day it's added. Plus dependency tests for the three new classes, and a WIC-only payload test (see reviewer notes).
- Updated the MO and TX tests that encoded the old behavior, including MO's `test_wic_income_source_coverage_is_partial`, whose "5 of 24" assertion was never right.

### Cleanup work carried here (moved out of #1659)

`Screen.has_benefit()` is an exact `name_abbreviated` match, and no white label ships a program literally named `snap` / `tanf` / `section_8` — they ship `co_snap`, `ks_snap`, `wa_snap`, `nc_tanf`, … So every read of a bare base name was always False and the branch it guarded never ran. This PR is where that gets fixed, because its own `ReceivesSnapDependency` resolves receipt through `has_base_benefit("snap")` and would otherwise send a boolean that is False for every WA SNAP household.

- `has_base_benefit()` in place of the dead reads across ~14 calculators (CO Denver sidewalk/trash rebates, CO WAP, CESN vehicle exchange, CESN home efficiency, NC Head Start, NC SUN Bucks, TX CCAD, TX WAP, WA Head Start, WA WIC, KS SSDI, WA SSDI, IL + CO Nurse-Family Partnership) and in the PolicyEngine `Snap` input. This also closes the `has_benefit("wic")` gap MFB-1571 flagged as a related observation.
- New `Screen.has_benefit_or_variant()`, with `has_benefit_from_list()` routed through it — `presumptive_eligibility` lists legitimately mix exact names and base programs, so a blanket `has_base_benefit` swap would have broken the exact-name entries.
- Income-derived receipt in `_derived_current_benefit_names()`, resolving through `base_program` rather than a hardcoded name list (the old list missed `ks_ssi`), extended to SNAP/WIC so a reported amount and the receipt signal can't disagree.
- `Snap.value()` sending a reported dollar amount, output-slot reservation, and the two-pass `pe_input()` write so an output's `None` never displaces a reported input value.
- Guard tests: `test_has_benefit_literals` (AST-parses the calculator tree, fails on a base program name **or** a non-literal argument to `has_benefit()`) and `test_presumptive_eligibility_resolution`.

**Behavior flips this introduces, beyond the WIC fix:** NC SUN Bucks now excludes SNAP/TANF households; WA WIC and WA Head Start now grant adjunctive/categorical eligibility above the income limit; Head Start / Early Head Start gain PE categorical eligibility for SNAP-reporting households across the SNAP states. All three change live results and need QA alongside the WIC states.

## Testing

- **Migrations to run:** `python manage.py migrate`. This branch renumbers #1659's migration to `0165_backfill_wa_snap_base_program`. If you already applied it under the old number, delete the stale row first or `0165` looks unapplied while its work is done:
  ```sql
  DELETE FROM django_migrations WHERE app='programs' AND name='0164_backfill_wa_snap_base_program';
  ```
- **Configuration updates needed:** none for this PR. (#1659 still needs its `add_config --all` and translation steps.)
- **Automated tests:** `python manage.py test screener programs configuration` → **2536 pass, 0 failures** locally.
- **Manual testing steps:**
  1. **Income now binds.** A WIC-eligible household (pregnant adult, or a child under 5) at ~$108k/yr should now be **not eligible** for WIC in CO / NC / MA / TX / IL. Before this PR it returned eligible at every income.
  2. **The adjunctive band still holds, and this is the step most likely to be misread as a bug.** A pregnant applicant *above* 185% FPL but under the state's Medicaid limit should still be **eligible**, carried by the Medicaid branch. In MO that band runs to ~201% FPL (MO HealthNet for Pregnant Women, 196% + 5% disregard). Only above *both* thresholds does WIC drop out. Asserting a hard 185% cutoff will flag correct behavior as a bug.
  3. **Reported receipt.** Tick the SNAP or TANF benefit tile without entering an amount, at an income above the WIC limit: WIC should be eligible via the categorical branch. This path could never fire before.
  4. **Newly counted income types.** Enter child support received, workers' comp, alimony, or gifts — these now reach PE and reduce WIC eligibility at the margin. See the reviewer note below, because they reach more than WIC.

## Deployment

- **Post-deployment scripts needed:** `heroku run python manage.py migrate -a <app>` (no release phase in the `Procfile`, so migrations are manual). Nothing else specific to this PR.
- **Production config updates:** none.
- **Admin updates needed:** none.
- **Notify team/users of:** eligibility changes in **six** states, before this reaches prod.
  - WIC stops showing for high-income households in CO, NC, MA, TX, IL (and stays fixed in MO). This is the intended correction, but it removes a program from results for real users.
  - Households reporting child support, workers' comp, alimony or gifts may see **SNAP, SSI, state TANF and childcare-subsidy** values fall — see the reviewer note. That is a wider surface than WIC and deserves its own QA pass.
  - QA should cover CO, IL, MA, NC, TX and MO.

## Notes for Reviewers

- **The blast radius is wider than WIC, and that's the thing to scrutinize.** PolicyEngine gets one merged household per request, not one per program, so the four newly-sent income fields are visible to every calculator in the request. I checked PE's parameter files: `child_support_received`, `workers_compensation` and `alimony_income` are in `gov/usda/snap/income/sources/unearned.yaml`, `gov/ssa/ssi/income/sources/unearned.yaml`, the CO/NC/IL/TX state TANF unearned lists, MA CCFA, NC SCCA, IL AABD/HBWD and Denver property tax relief; `miscellaneous_income` is in the SNAP, school-meals and several state TANF lists. So a household reporting any of them will see those benefits fall. That is more correct — the income is legally countable and we were hiding it — but it changes shipped values well beyond this ticket. Say the word if you'd rather the four non-WIC-visible fields moved to their own ticket, leaving this PR to `irs_gross_income` + the receipt inputs.
- **Why the WIC-only payload test matters.** `TestMoWicPeInput.test_wic_alone_carries_every_income_source` builds the payload from the WIC calculator *alone*. With siblings present it passes either way — which is precisely why this bug survived: CO/TX/MA/NC screens run Medicaid and Head Start, both of which send `irs_gross_income`, so WIC borrowed their inputs. MO shipped WIC as its only program and the omission became visible immediately. A WIC-only payload is the only assertion that catches a regression here.
- **`receives_snap` / `receives_tanf` are not version-gated, and probably should be.** Both are policyengine-us 1.779.3 fields. PE's `current` is exactly 1.779.3 as of 2026-08-05, so this works against current — but if the DB pin is an older exact version, an unknown input 400s the *entire* request, taking down every PE program in it, not just WIC. Either add `min_pe_version = (1, 779, 3)` to both (the pattern the other gated dependencies use) or confirm the staging and prod pins are ≥ 1.779.3 before merge.
- **This carries a slice of MFB-1312.** `receives_snap` / `receives_tanf` are listed under that ticket's sub-issue 1. They're the safe half — plain booleans that confer categorical eligibility without pinning any computed amount — but MFB-1312's scope shrinks accordingly, and its sub-issue 1 should not re-add them.
- **Rebase needed once #1659 merges.** This branch still contains #1659's pre-split commits, and that PR was rewritten to a single slim commit. Expect conflicts in `configuration/white_labels/base.py`, `screener/models.py`, `screener/tests/test_models.py` and `dependencies/tests/test_member.py` — take main's side for the income-capture hunks; the cleanup hunks in those files stay.
- **Known limitations:**
  - Seven WIC sources have no screener field and stay unreachable: `military_service_income`, `gi_cash_assistance`, `financial_assistance`, `strike_benefits`, `disability_benefits`, `educational_assistance`, `railroad_benefits`. `UNREACHABLE_SOURCES` in the new test file pins the list, so adding a screener field for one of them fails the test until the mapping is added.
  - **`boarder` income is still unmapped.** It arguably belongs in `rental_income`, but that dependency is in `irs_gross_income` and shared by every consumer, so changing it is a much wider blast radius than this ticket. Left alone deliberately.
  - **`veteran` income maps to `taxable_pension_income`.** That makes it count for WIC (correct — `veterans_benefits` is a WIC source), but VA benefits are not taxable, so the mapping is questionable for the tax programs that read the same field. Pre-existing; not touched here.
  - PE's imputed TANF adds ~$3,505/yr of phantom income to a $0-income MO household's WIC countable income, because `spm.Tanf` sends `None` when nothing is reported and PE computes it. It never changes an outcome — such households pass categorically — so I left the `Tanf` contract alone; MFB-1312 owns it.
  - Nutritional risk defaults to `True` in PE and the screener doesn't ask, so `is_wic_eligible`'s third term is always satisfied. Disclosed in the program description; unchanged here.
- **Future considerations:**
  - `receives_snap` / `receives_tanf` are read by Head Start's and SNAP's own categorical tests too, and by Medicaid community-engagement. Wiring them into those calculators would close the same reported-vs-computed gap the KS Head Start spec documents as a known limitation — deliberately out of scope here, worth its own ticket.
  - The federal `Wic` has no `spec.md` ("Fed (as-is)" tier). The eligibility logic is now documented in the class docstring instead; if WIC ever gets a spec, the 185%-vs-Medicaid boundary is the thing that most needs writing down.

## Summary by CodeRabbit

* **New Features**
  * Added SNAP and WIC as recognized government income options.
  * Improved support for state-specific benefit variants across eligibility screenings.
  * Added more complete WIC income, household, and benefit-receipt information for calculations.

* **Bug Fixes**
  * Corrected eligibility checks for programs using SNAP, WIC, TANF, SSI, and SSDI.
  * Improved benefit-derived income handling while excluding SNAP and WIC from countable income totals.
  * Preserved reported benefit amounts and improved calculated benefit results when information is incomplete.


david-j-stoll:
Closing, as this PR contained unapplicable changes.

PR #1681, opened 2026-08-06
MFB-1604: resolve dead has_benefit() base-name reads through base_program

david-j-stoll:
## Context & Motivation

- Linear: **MFB-1604** — dead `has_benefit()` base-name reads: categorical / presumptive eligibility branches that never fire
- **Split out of #1674 (MFB-1571).** That PR had absorbed this cleanup because it was stacked on MFB-1382's branch and inherited it. This is the independent half: it touches no PolicyEngine payload and merges on its own. #1674 is now WIC-only, and MFB-1571 has been re-narrowed to match.
- Branched fresh from `main` rather than rebased out of #1674, so the diff is a clean extraction rather than a replay through 8 conflicting files.
- Unblocks **MFB-1312** (actual-receipt contract), which sends `takes_up_X_if_eligible: false` for households *not* receiving X and would classify every real recipient as a non-recipient while these reads stay dead.

`Screen.has_benefit()` is an exact `name_abbreviated` match. Calculators across six white labels ask it about **base program names**, and for `snap` / `tanf` / `wic` / `section_8` no white label ships a program under the bare name — they ship `co_snap`, `ks_snap`, `wa_snap`, `nc_tanf`, `ma_tafdc`, `wa_hcv`, … So the read is always False and the categorical / adjunctive / presumptive-eligibility branch it guards silently never fires. 25 such reads across 14 calculator files, found by AST-scanning the calculator tree against `BaseProgram.choices`.

### `ssi` / `ssdi` are the partial case

CO, IL, MA and NC **do** ship programs literally named `ssi` and `ssdi`, with `base_program` set to the same string and `show_in_has_benefits_step=true`. So the exact-match read worked in those four white labels and was dead only in KS, WA, TX and MO, which ship `ks_ssdi` / `wa_ssi` / `tx_ssi` / `mo_ssi`.

This is why the guard test is a blanket ban rather than a per-name allowlist: `has_base_benefit("ssi")` is a superset of the exact read and correct in every white label, so there's no reason to keep the fragile form anywhere. The fix is uniform even though the bug wasn't.

### Why now, and not with #1674

`main` already reads `has_base_benefit("snap")` in the PolicyEngine `Snap` and `Tanf` inputs (landed with #1678), but **`wa_snap.base_program` is NULL in staging and prod**. So reported SNAP receipt currently confers no categorical eligibility for any WA household — a live bug on `main` today, not a latent one. The migration here fixes it, which is the main argument for landing this ahead of #1674.

## Changes Made

- **`Screen.has_benefit_or_variant(name)`** (`screener/models.py`) — matches either an exact `name_abbreviated` or a `base_program` group. `has_benefit_from_list()` now routes through it, because `presumptive_eligibility` lists legitimately mix both kinds of name — CO WAP's is `("andcs", "ssi", "snap", "leap", "tanf")`, CO-only programs next to cross-state base programs. A blanket `has_base_benefit` swap would have broken the exact-name entries.
- **`has_base_benefit()` in place of the dead base-name reads** across 14 calculators: CO Denver property-tax relief, CO Denver sidewalk rebate, CO Denver trash rebate, CO WAP, CO Low-Wage COVID Relief, CO Nurse-Family Partnership, CESN home efficiency, CESN vehicle exchange, IL Nurse-Family Partnership, KS SSDI, NC Medicare Savings, NC SUN Bucks, TX CCAD, TX WAP, WA Head Start, WA NSLP, WA SSDI, WA WIC.
- **`wa_snap.base_program` must be set to `"snap"` via Django admin** on staging and prod. This started as migration `0165_backfill_wa_snap_base_program`; per review it's an admin edit instead. `wa_snap` is the only SNAP variant with a NULL `base_program` — confirmed against **staging (43 rows) and prod (41 rows)**, where it is the only NULL in the `snap`/`tanf`/`wic`/`ssi`/`ssdi`/`section_8` group. See Deployment below for the steps.
- **`seed_program(..., base_program=)`** (`screener/tests/helpers.py`) — `Program.objects.new_program` leaves the field unset, so tests exercising the structural resolution need to pass it.
- **Guard test `test_has_benefit_literals`** — AST-parses the calculator tree and fails on a base program name **or** a non-literal argument to `has_benefit()`. The non-literal rule is the load-bearing half: every one of these reached the bad name by looping a class attribute (`presumptive_eligibility`, `categorically_eligible`, `auto_eligible_benefits`), so a literal-only check would have passed while production was broken. Includes self-tests that the walker actually finds both shapes, so it can't silently stop working if the `ast` assumptions drift.
- **Guard test `test_presumptive_eligibility_resolution`** — pins the resolution seam with real `Program` / `CurrentBenefit` rows for the calculators that have no suite of their own, plus a real `SunBucks.household_eligible()` run with only the FPL lookup and income stubbed.
- **Test updates** for the calculators whose mocks encoded the old behavior (KS/WA SSDI, TX CCAD, TX WAP, WA Head Start, WA WIC, IL NFP). In each, `has_benefit` is now stubbed to return False explicitly and `has_base_benefit` is stubbed separately — a single `has_benefit` mock would let a base-name read pass on a `Mock`'s truthy default and hide exactly this bug.

### Three additional fixes, found while verifying the above against the program table

These were caught by checking `show_in_has_benefits_step` and `base_program` for every name involved, rather than by reading names:

1. **WA WIC's Medicaid adjunctive leg was still dead after the base-name swap.** WA captures Apple Health as *insurance* (`health_insurance_options` → `member.insurance.medicaid`), not as a has-benefits program: `wa_apple_health_medicaid` and `wa_apple_health_for_kids` both have `show_in_has_benefits_step=false`, so nothing ever writes a `base_program="medicaid"` row on a WA screen and `has_base_benefit("medicaid")` is False for every household. Now reads `has_insurance_types(("wa_apple_health_medicaid",))`, the way `wa/wap` already does. Only the Medicaid leg is counted — whether Apple Health for Kids (CHIP) confers WIC adjunctive eligibility is a policy question, so it stays out.
2. **`LowWageCovidRelief` used base names as `self.data` keys too.** `self.data` is keyed by `name_abbreviated` (see `eligibility_results()`), so `self.data.get("snap")` was also always None — the calculated-eligibility leg was dead alongside the receipt leg. Switched to CO's real names (`co_tanf`, `co_snap`, `co_wic`, `leap`), which resolve for both uses. The program is currently inactive, so this is low stakes.
3. **The CESN energy calculators carried CO names.** `leap`, `cowap`, `rtdlive` and `co_care` match nothing on a CESN screen. Vehicle exchange now lists `cesn_care` / `cesn_rtdlive` / `cesn_section_8` alongside the `liheap` / `snap` / `wic` / `ssi` / `ssdi` base groups; home efficiency reads the `liheap` group in place of `leap`. `liheap` resolves to `cesn_leap`, which the Deployment step makes reportable.

   **Base group rather than the exact `cesn_leap`,** because `has_benefit_from_list` resolves either and the group is a strict superset — it also survives a rename of the row, which has happened here before (`0140_rename_cesn_andso_care`). Every other entry in these lists is a base group, so the exact name would have been the odd one out.

   **WAP is deliberately absent from the receipt list.** `cesn_cowap` stays unreportable and is already read as *calculated* eligibility via `calculated_presumptive_eligibility`, so a receipt entry would have no signal to pick up.

   *Changed during review:* an earlier revision dropped LEAP and WAP entirely, on the reasoning that `cesn_leap` / `cesn_cowap` are `has_calculator=true` programs and therefore never reportable. The first half of that is a real distinction; the conclusion was wrong — **47 of 63** active reportable programs across all white labels are `has_calculator=true`, CO's own `leap` included. Deleting LEAP removed a leg of the program rule rather than renaming it, so it's back.

## Testing

- **Migrations to run:** none. (`makemigrations --check` reports no changes; leaf stays `0164_seed_mo_generic_referrers`.)
- **Configuration updates needed:** none.
- **Automated tests:** `python manage.py test screener programs configuration` → **2543 pass, 0 failures** (main is 2529; +14 from the new tests). `black --check --line-length 120` clean.
- **Manual testing steps:**
  1. **The `wa_snap` admin edit.** After setting it, confirm `SELECT name_abbreviated, base_program FROM programs_program WHERE name_abbreviated = 'wa_snap';` returns `snap`.
  2. **NC SUN Bucks now excludes SNAP/TANF households.** A NC screen reporting SNAP should no longer be shown SUN Bucks — they're auto-enrolled, so showing it was the bug.
  3. **WA WIC adjunctive eligibility.** A WA household above 185% FPL with a WIC-category member: reporting SNAP or TANF should now return eligible. Separately, selecting **Apple Health (Medicaid)** as insurance should also return eligible — that's the leg fixed in this PR, and it's the one most likely to be missed in QA because it's an insurance question, not a benefit tile.
  4. **WA Head Start** categorical eligibility above the income limit for a household reporting SNAP / TANF / SSI.
  5. **KS and WA SSDI** should now be correctly excluded for a household already reporting SSDI (previously shown to existing recipients).
  6. **CO / IL Nurse-Family Partnership** — reporting WIC should bypass the income test.
  7. **Regression check on the four white labels where the read already worked.** CO, IL, MA and NC ship real `ssi` / `ssdi` rows, so `has_benefit("ssi")` was live there. Confirm CO Denver property-tax relief and NC Medicare Savings behave identically before and after for an SSI-reporting household — the swap should be a no-op there, since each of those white labels has exactly one program in the `ssi` base group.

## Deployment

- **Post-deployment scripts needed:** none. This PR adds no migrations — `makemigrations --check` reports no changes and the leaf stays `0164_seed_mo_generic_referrers`.
- **Production config updates:** none.
- **Admin updates needed: YES — one edit, required on both staging and prod.**

  **Set `base_program = snap` on WA's SNAP program (`wa_snap`).**

  This began as migration `0165_backfill_wa_snap_base_program`; per review it's a Django admin edit instead. **It is not optional and it is not covered by the code in this PR.** Until it's done, WA is the one state where reported SNAP receipt still confers no categorical/adjunctive eligibility (WA Head Start, WA WIC) and no PolicyEngine categorical eligibility — a live bug on `main` today, not a latent one. Everything else in this PR ships correctly without it; only the WA SNAP paths depend on it.

  **Steps** (repeat on staging, then prod):

  1. Sign in to Django admin. You need a **superuser** or an account scoped to the **WA** white label — `ProgramAdmin` filters the queryset by the signed-in user's white labels.
  2. Go to **Programs → Programs**.
  3. Find the row. There is no white-label filter on this list, so either:
     - use the right-hand sidebar **By base program → `-`** (the empty bucket), which is a short list; or
     - search `SNAP` — note the search box matches the *translated program name*, not `name_abbreviated`.

     Either way, confirm you have the right row by the **Name abbreviated** column: it must read exactly **`wa_snap`**. Do not touch `co_snap`, `ks_snap`, `nc_snap`, `tx_snap`, `il_snap`, `ma_snap` or `cesn_snap` — those already have `base_program` set.
  4. Open `wa_snap` and set **Base program** to **SNAP**.
  5. Leave every other field alone — in particular `active`, `show_in_has_benefits_step` and `has_calculator`.
  6. **Save.**

  **Verify:**

  ```sql
  SELECT p.name_abbreviated, p.base_program
  FROM programs_program p
  JOIN screener_whitelabel w ON w.id = p.white_label_id
  WHERE w.code = 'wa' AND p.name_abbreviated = 'wa_snap';
  ```

  Expected: `wa_snap | snap`.

  **Then smoke-test it** on the environment you just changed: a WA screen reporting SNAP, above the income limit, with a WIC-category member should now return WA WIC as eligible. If it doesn't, the edit didn't land on that environment.

  **Rollback:** clear the field back to empty. It's a single nullable column on one row, with no dependent data.

  **Known follow-up:** an admin edit doesn't reach databases created later (local, CI, a fresh white-label clone). `import_program_config` writes the same field, so it's recoverable, but `wa_snap.base_program` is no longer guaranteed by code. Worth a config-level fix if WA is ever re-seeded.

  ---

  **Set `show_in_has_benefits_step = True` on CESN's LEAP program (`cesn_leap`).**

  Makes LEAP selectable on the has-benefits step for CESN, which brings the `cesn_leap` presumptive read in this PR to life. Without it that read is correctly named but permanently False.

  **Why it's safe to do here:** it's a data change with no code diff, in the same class as the `wa_snap` edit, and the read it activates is already part of this PR. It is *not* the convention break it first appears to be — **47 of 63** active reportable programs across all white labels are `has_calculator=True`, including **CO's own `leap`**, which is both calculated and reportable. CESN diverges from its parent white label here for no recorded reason.

  **Steps** (staging, then prod): Django admin → **Programs → Programs** → open **`cesn_leap`** (CESN white label) → tick **Show in has benefits step** → Save. Leave `active`, `has_calculator` and `show_on_current_benefits` alone.

  **Verify:**

  ```sql
  SELECT p.name_abbreviated, p.show_in_has_benefits_step, p.has_calculator, p.active
  FROM programs_program p
  JOIN screener_whitelabel w ON w.id = p.white_label_id
  WHERE w.code = 'cesn' AND p.name_abbreviated = 'cesn_leap';
  ```

  Expected: `cesn_leap | t | t | t`. Then confirm LEAP appears as a tile on the CESN has-benefits step, under **Assistance Programs**.

  **No new translations needed** — `cesn_leap` already has a name ("Colorado Low-Income Energy Assistance Program (LEAP)"), a `website_description` ("Help with winter heating costs (seasonal: November through April)") and a category, so the tile renders fully. Compare `cesn_rtdlive`, which has no category and would fall into the frontend's "Other" bucket.

  **Blast radius — one active calculator:**

  - **CESN Home Efficiency Assistance (`cesn_heap`, active)** — the presumptive branch starts firing. LEAP recipients bypass *all three* of its other tests (Colorado Springs Utilities customer, homeowner, income ≤60% SMI). This **adds** households to results and is the intended effect, but it is a live result change and needs the same product notice as the rest of this PR.
  - **`cesn_leap` itself** — a household that ticks LEAP gets `already_has: true` on the LEAP result (`screener/views.py:573`). Correct and expected.
  - **CESN Energy EBT (`cesn_energy_ebt`)** — reads `has_benefit("cesn_leap")` at `energy_ebt/calculator.py:17` to *exclude* LEAP households. That read is dead today for the same reason, so flipping the flag would revive it — **but the program is inactive**, so nothing changes in prod. Worth knowing before it's ever reactivated.
  - **Nothing else.** The affordability calculators (`cesn_xcelgap`, `cesn_xceleap`, `cesn_bhgap`, `cesn_bheap`, `cesn_cngba`, `cesn_poipp`) and `cesn_eoccip` read LEAP as *calculated* eligibility via `self.data`, not as receipt, so they're untouched. After this PR no calculator reads the `liheap` base group at all, so only the exact-name read is affected.

  **Seasonality — worth a content look, not a blocker.** LEAP runs November–April. A "do you currently receive this?" tile for a seasonal benefit may want different wording; the description already says so, and today's date means the presumptive path stays quiet until the season opens either way.

  **Deliberately not doing the same for `cesn_cowap`.** It's the symmetric case (`has_calculator=True`, `show_in_has_benefits_step=False`, while CO's `cowap` is reportable), so consistency would argue for it — but no *active* CESN calculator reads WAP receipt. Vehicle exchange was the only one, it's inactive, and its WAP leg is covered by `calculated_presumptive_eligibility` anyway. Flipping it would add a tile with no consumer. Left for the CESN config audit.

  **Rollback:** untick the box. Nothing persists that depends on it — existing `CurrentBenefit` rows for `cesn_leap` would simply stop being written for new screens.
- **Notify team/users of:** eligibility changes in **six states**, before this reaches prod. These are live result changes, not refactors — several *remove* a program from results, and one *adds* programs for households that previously fell through to an income test.
  - **NC SUN Bucks** stops showing for SNAP/TANF households. Correct (they're auto-enrolled), but it removes a program from results for real users.
  - **WA WIC** and **WA Head Start** start showing for households above the income limit that report SNAP/TANF/SSI or have Apple Health.
  - **WA WAP, TX CCAD, TX WAP, CO WAP, CO Denver sidewalk/trash rebates, NC Head Start** presumptive/categorical branches begin firing, so more households qualify without meeting the income test.
  - **KS SSDI** and **WA SSDI** stop showing for households already receiving SSDI.
  - **CO / IL Nurse-Family Partnership** gain the WIC income-test bypass.
  - **CESN Home Efficiency Assistance** starts granting presumptive eligibility to LEAP recipients once `cesn_leap` is made reportable — bypassing the utility-provider, homeowner and 60%-SMI tests. Lands with the admin edit, not the code.
  - **WA SNAP recipients** gain PolicyEngine categorical eligibility they should already have had — the `Snap`/`Tanf` inputs on `main` read `has_base_benefit("snap")` but WA had no `base_program` to match. This one lands with the admin edit above, not with the code.
  - QA should cover **CO, IL, KS, NC, TX, WA and CESN**.

## Notes for Reviewers

- **The thing most worth scrutinising is the `ssi` / `ssdi` claim, because the obvious version of it is wrong.** It's tempting to describe this whole PR as "these reads were always False." That's true for `snap` / `tanf` / `wic` / `section_8`, and **false** for `ssi` / `ssdi`, which CO, IL, MA and NC ship under the bare name. The swap is still safe — it's a superset, and each of those white labels has exactly one program in the base group, so behaviour is unchanged there — but if you're reviewing on the assumption that every changed line was dead code, the `ssi`/`ssdi` lines are a behaviour-preserving change sitting in a diff full of behaviour-changing ones. The guard test's docstring spells this out; please check I've got it right.
- **One review comment was partly reversed — flagging it rather than doing it quietly.** @catonph asked whether all the CESN entries exist and noted they "should be programs where `has_calculator` is false." Checking that surfaced a real bug (`co_care` → `cesn_care`, and LEAP/WAP genuinely unreachable), so the question did its job. But the `has_calculator=false` rule itself doesn't hold: 47 of 63 active reportable programs are `has_calculator=true`, and CO ships `leap` as both calculated and reportable while CESN doesn't. So rather than dropping LEAP, this PR makes `cesn_leap` reportable (Deployment) and keeps the entry. WAP stays dropped, for the separate reason that it's already read as calculated eligibility.
- **Why `has_benefit_or_variant` exists rather than a blanket `has_base_benefit` swap.** `presumptive_eligibility` lists mix exact names and base programs. `andcs` and `leap` are CO program names with no matching `base_program` value; `snap` and `tanf` are base programs with no matching exact name. A single-method swap in either direction breaks half of each list.
- **How the naming claims were verified.** Not by reading code: by querying `programs_program` joined to `screener_whitelabel` for `name_abbreviated`, `base_program`, `active`, `has_calculator` and `show_in_has_benefits_step` across all white labels, resolving each list entry the way `has_benefit_or_variant()` does (exact `name_abbreviated` first, then `base_program` group). That's what surfaced the three extra fixes, and it's how I'd suggest checking any claim in this PR you doubt.

  **What makes a name reachable.** A `presumptive_eligibility` entry can only ever be True if the household can report it, and the has-benefits step is served by `HasBenefitsProgramsView` (`screener/views.py:820`) — `Program.objects.filter(active=True, show_in_has_benefits_step=True, white_label__code=...)`. The frontend submits the `name_abbreviated` from that response, and `_write_current_benefits()` resolves it against the same column. So the gate is two booleans on the Program row, nothing else.

  Worth stating explicitly because it's easy to get wrong: **`category_benefits` does not drive the has-benefits step.** Its keys (`snap`, `wic`, `rtdlive`, `section_8`, …) are mostly bare names that match no `name_abbreviated`, which looks alarming until you trace `AlreadyHasBenefits.tsx` → `getHasBenefitsPrograms()` → the view above and find that config isn't in the path. An earlier draft of this PR cited a "`category_benefits` config gap" as the reason the CESN LEAP branch stays inert; that was the wrong mechanism and has been corrected in the code comment.

- **Known limitations:**
  - **CESN LEAP receipt — resolved here, via the admin edit in Deployment.** `cesn_leap` had `show_in_has_benefits_step=false`, so it never appeared on the has-benefits step and home efficiency's presumptive read could never fire. Flipping that flag (see Deployment) fixes it. My first pass called this a convention break, because every *CESN* reportable program is a tracking row — that reasoning was wrong: 47 of 63 active reportable programs across all white labels are `has_calculator=true`, CO's own `leap` included. CESN was the outlier. Vehicle exchange no longer lists LEAP or WAP at all — WAP is covered by its `calculated_presumptive_eligibility` leg. (`cesn_energy_vec` is also inactive since 2026-01-21, so vehicle exchange is moot in prod either way.)
  - **`cesn_cowap` is left non-reportable**, though it's the symmetric case to `cesn_leap`. WAP is already covered by vehicle exchange's `calculated_presumptive_eligibility`, and no active CESN calculator reads WAP *receipt*, so the tile would have no consumer. Folded into the CESN config audit rather than fixed here.
  - **`energy_ebt` still reads `has_benefit("cesn_leap")` by exact name** (`energy_ebt/calculator.py:17`) rather than the `liheap` group — same rename fragility as the entries fixed here. `cesn_energy_ebt` is inactive and the line is outside this PR's diff, so it's left alone; worth folding into whichever ticket picks up the overloaded-attribute cleanup below.
  - **`presumptive_eligibility` is an overloaded attribute name in `co/energy_calculator`.** In the gas/electric affordability calculators it's a `self.data` calculated-eligibility list, where `cesn_leap` is the *correct* key. In home-efficiency and vehicle-exchange it's a reported-receipt list read via `has_benefit`. `energy_ebt` checks both legs for the same program. I kept both as receipt reads rather than silently converting home-efficiency's LEAP branch to a calculated read, because "receives LEAP" vs "would be eligible for LEAP" is a product decision, not a naming one. Worth deciding and then naming the two attributes apart.
  - **Nothing checks that a `presumptive_eligibility` entry is a program its white label actually offers.** The AST guard can't see this — it needs the program table. That's precisely how the CESN entries survived the first pass. A DB-backed test over each white label's `show_in_has_benefits_step` programs would close it.
  - **`category_benefits` has drifted from the program table** — on CESN only 3 of 18 keys (`cesn_andso`, `cesn_care`, `cesn_heap`) match a `name_abbreviated`; the rest are bare names like `snap` and `rtdlive`. WA lists `medicaid` and `chp`, CO lists `ccap` where the shipped row is `cccap`. **This is not a screener bug** — the has-benefits step reads the Program table, not this config, so no tick is being dropped. But something still consumes `category_benefits` (`add_config` writes it to the `Configuration` table), and a config whose keys mostly resolve to nothing is worth an audit on its own terms. Pre-existing and out of scope here.
  - The guard test only scans `programs/programs/**`. `has_benefit()` calls elsewhere (e.g. `screener/views.py`) aren't covered; that one passes `program.name_abbreviated`, so it's fine, but the scope is worth knowing.

- **Future considerations:**
  - **MFB-1312** should be re-checked against this branch. Its premise — that SNAP/WIC receipt can never read True — no longer holds once this lands, which is the point, but the sub-issue text still assumes the broken behaviour.
  - The **`_derived_current_benefit_names()`** rework (resolving derived receipt through `base_program` rather than a hardcoded name list, which currently misses `ks_ssi`) is a natural follow-on but lives in #1659/#1674's territory and isn't here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved eligibility screening across multiple assistance programs and states.
  * State-specific variants of SNAP, TANF, WIC, SSI, SSDI, Medicaid, and related benefits are now recognized more consistently.
  * Presumptive eligibility checks now correctly identify qualifying benefits without requiring exact program-name matches.
  * Updated energy, vehicle exchange, housing, food, childcare, utility, and tax-relief eligibility rules.
  * Washington WIC screening now distinguishes Apple Health Medicaid from CHIP coverage.


programs/migrations/0165_backfill_wa_snap_base_program.py
```diff
@@ -0,0 +1,42 @@
...
+    dependencies = [
+        ("programs", "0164_seed_mo_generic_referrers"),
+    ]
+
+    operations = [
+        # No-op reverse: forward fills NULLs only and doesn't record which rows it filled,
+        # so clearing every wa_snap row on the way back would also discard a base_program
+        # set independently (import_program_config writes the same field).
+        migrations.RunPython(forward, migrations.RunPython.noop),
+    ]
```
catonph: Don't need a migration - can set this in prod admin

programs/programs/co/energy_calculator/home_efficiency_assistance_program/calculator.py
```diff
@@ -28,8 +28,18 @@ class EnergyCalculatorHomeEfficiencyAssistance(ProgramCalculator):
...
+    #
+    # Caveat worth knowing before trusting this branch: cesn_leap has
+    # show_in_has_benefits_step=False, and CESN's category_benefits still offers the key
+    # "leap", which matches no CESN program and is dropped on write. So LEAP receipt is not
+    # currently capturable on a CESN screen and this stays False in practice. That is a
+    # configuration gap, not a naming one, and is tracked separately — the read is written
+    # correctly here so it starts working the moment the config is fixed.
     presumptive_eligibility = [
-        "leap",
+        "liheap",
```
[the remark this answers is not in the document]
catonph: check this out?

programs/programs/co/energy_calculator/vehicle_exchange/calculator.py
```diff
@@ -11,16 +11,33 @@ class EnergyCalculatorVehicleExchange(ProgramCalculator):
     amount = 4_000
     min_age = 18
     ami_percent = "80%"
-    presumptive_eligibility = ["co_care", "cowap", "rtdlive", "cesn_section_8", "ssdi", "wic", "leap", "snap", "ssi"]
+    # CESN program (see cesn_calculators), so CESN's own names are what the join table can
+    # hold — the previous CO names (co_care, cowap, rtdlive, leap) match nothing on a CESN
+    # screen. has_benefit_from_list takes each entry as either an exact name or a
```
catonph: Remove this "the previous" change log vibe comment

programs/programs/co/energy_calculator/vehicle_exchange/calculator.py
```diff
@@ -11,16 +11,33 @@ class EnergyCalculatorVehicleExchange(ProgramCalculator):
...
+    presumptive_eligibility = [
+        "cesn_care",
+        "cesn_rtdlive",
+        "cesn_section_8",
+        "liheap",
+        "wap",
+        "ssdi",
+        "wic",
+        "snap",
+        "ssi",
```
catonph, line 24: Double check we have all 5 of these programs in CESN? (the should be programs where has_calcuclator is false)

programs/programs/co/low_wage_covid_relief/calculator.py
```diff
@@ -7,7 +7,12 @@
 
 class LowWageCovidRelief(ProgramCalculator):
     amount = 1_500
-    auto_eligible_benefits = ("tanf", "snap", "wic", "leap")
+    # CO's own program names, not base program names. Each entry is used twice — as a
+    # reported-receipt read AND as a `self.data` key — and `self.data` is keyed by
+    # `name_abbreviated` (see eligibility_results()), so a base name like "snap" resolves
+    # neither: it matched no CO program for the receipt read, and no `self.data` entry for
+    # the calculated-eligibility read.
+    auto_eligible_benefits = ("co_tanf", "co_snap", "co_wic", "leap")
```
catonph, line 12: is this comment necessary?

programs/programs/co/low_wage_covid_relief/calculator.py
```diff
@@ -32,9 +37,11 @@ def household_eligible(self, e: Eligibility):
         # other benefits
         has_benefit = False
 
+        # has_benefit_or_variant rather than has_benefit_from_list because the per-entry
+        # result is needed for the self.data lookup on the same name.
```
catonph: This is a misleading explanation since has_benefit_from_list calls has_benefit_or_variant right?

programs/programs/co/nurse_family_partnership/calculator.py
```diff
@@ -61,6 +61,8 @@ def member_eligible(self, e: MemberEligibility):
 
         insurance: Insurance = member.insurance
         has_medicaid = insurance.medicaid or insurance.emergency_medicaid
-        has_wic = self.screen.has_benefit("wic")
+        # has_base_benefit resolves co_wic, the live CO program. The bare name matched only
+        # CO's retired "wic" row, which the has-benefits step no longer offers.
```
catonph: change log comment

programs/programs/il/nurse_family_partnership/calculator.py
```diff
@@ -51,7 +51,8 @@ def household_eligible(self, e: Eligibility):
         gross_income = int(self.screen.calc_gross_income("yearly", ["all"]))
         is_income_eligible = gross_income <= income_limit
 
-        has_wic = self.screen.has_benefit("wic")
+        # has_base_benefit resolves il_wic; the bare name matches nothing in this WL.
```
catonph: "the bare name matches nothing in this WL." in this comment may be more confusing than helpful?

programs/programs/ks/ssdi/calculator.py
```diff
@@ -92,8 +92,9 @@ def member_eligible(self, e: MemberEligibility):
         e.condition(not already_receiving_ss_retirement)
 
     def household_eligible(self, e: Eligibility):
+        # has_base_benefit resolves ks_ssdi; the bare name matches nothing in this WL.
```
catonph: "the bare name matches nothing in this WL." may be more confusing than helpful here

programs/programs/nc/sun_bucks/calculator.py
```diff
@@ -15,9 +15,12 @@ def household_eligible(self, e: Eligibility):
         income_limit = int(self.fpl_percent * fpl.get_limit(self.screen.household_size))
         gross_income = int(self.screen.calc_gross_income("yearly", ["all"]))
 
-        # Must not have the following benefits
-        e.condition(not self.screen.has_benefit("snap"), messages.must_not_have_benefit("snap"))
-        e.condition(not self.screen.has_benefit("tanf"), messages.must_not_have_benefit("tanf"))
+        # Must not have the following benefits. has_base_benefit matches this white label's
+        # variants (nc_snap / nc_tanf); the exact-name reads never matched anything, so the
+        # conditions always passed and SNAP/TANF households — who are auto-enrolled in SUN
+        # Bucks — were shown the program anyway.
```
catonph: change log comment

programs/programs/tests/test_has_benefit_literals.py
```diff
@@ -0,0 +1,141 @@
...
+        if not isinstance(node, ast.Call) or not isinstance(node.func, ast.Attribute):
+            continue
+        if node.func.attr != "has_benefit" or not node.args:
+            continue
+
+        first = node.args[0]
+        if not isinstance(first, ast.Constant) or not isinstance(first.value, str):
+            rendered = ast.unparse(first)
+            offenders.append((node.lineno, f"has_benefit({rendered}) — non-literal argument"))
+        elif first.value in BASE_PROGRAM_NAMES:
```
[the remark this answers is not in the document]
catonph: Worth a test?

programs/programs/wa/head_start/calculator.py
```diff
@@ -57,9 +57,13 @@ def household_eligible(self, e: Eligibility):
         # Foster care categorical — any age-eligible member regardless of income (§ 1302.12(c)(1)(iv))
         has_foster_child = any(me.eligible and me.member.relationship == "fosterChild" for me in e.eligible_members)
 
-        # TANF, SSI, SNAP categorical eligibility per OHS interpretation (ACF-IM-HS-22-03)
+        # TANF, SSI, SNAP categorical eligibility per OHS interpretation (ACF-IM-HS-22-03).
```
catonph: probably dont' want this citation here

programs/programs/wa/head_start/calculator.py
```diff
@@ -57,9 +57,13 @@ def household_eligible(self, e: Eligibility):
...
+        # has_base_benefit resolves WA's wa_tanf / wa_ssi / wa_snap variants — the bare
+        # names are exact-match reads that never hit, so this never fired.
```
catonph: change log comment

programs/programs/wa/wic/calculator.py
```diff
@@ -36,8 +36,23 @@ def member_eligible(self, e: MemberEligibility):
...
+        # through to the income test.
+        #
+        # Medicaid is NOT a has-benefits program in WA — Apple Health is collected as
+        # insurance, so wa_apple_health_medicaid never enters current_benefits and
+        # has_base_benefit("medicaid") would be False for every WA screen. Read it off
+        # insurance instead, the way wa/wap does.
+        #
+        # Only the Medicaid leg is counted. Apple Health for Kids maps to the CHP+ field
+        # and mixes Medicaid and CHIP coverage; whether CHIP confers WIC adjunctive
+        # eligibility in WA is a policy question, not a naming one, so it stays out.
```
catonph: This comment seems like overkill

catonph (changes requested):
Mostly comments about comments

programs/programs/co/energy_calculator/vehicle_exchange/calculator.py
david-j-stoll: Good call, some of those should have been removed.

catonph (approved):

PR #1684, opened 2026-08-10
MFB-1571: fix WIC income-blindness — send PolicyEngine's WIC income sources

david-j-stoll:
## Context & Motivation

- Fixes **MFB-1571** — WIC calculators are income-blind: the federal `Wic` sends none of PolicyEngine's WIC income sources
- Related PR: #1674 — the previous attempt at this ticket. It accumulated work from two other investigations (income capture from #1659, and the dead-`has_benefit()` cleanup that has since landed on `main` as #1681 / MFB-1604). **This branch replaces it and is WIC-only.** #1674 should be closed.
- Related: **MFB-1213** (MO WIC), where the bug surfaced

PolicyEngine decides WIC with

```
is_wic_eligible = demographic_eligible & (meets_income_test | meets_categorical_test) & nutritional_risk
```

and the income term was never fed. The federal `Wic`'s only income input was `school_meal_countable_income`, which WIC's tree never reads — `wic_countable_income` sums its own parameter list, `gov.usda.wic.income.sources` (24 variables), and the school-meals variable is not among them.

Given none of its sources, PE substitutes an imputation. Two things follow: the imputation itself is nonzero (a $0-income MO household shows ~$3,505 of countable income, all of it PE's own computed TANF), and seeing no earnings PE also satisfies the categorical branch via its computed SNAP/TANF and Medicaid eligibility. Since the two branches are OR'd, the categorical branch carried eligibility on its own and the income test never bound. **WIC returned eligible at any reported income** in all six states.

`mo_wic` shipped a partial fix (`irs_gross_income` on the subclass only), so wage-type income bound there; `co_wic` / `il_wic` / `ma_wic` / `nc_wic` / `tx_wic` were fully affected.

## Changes Made

- **New `wic_income` bundle** (`dependencies/__init__.py`) — every source in `gov.usda.wic.income.sources` the screener actually collects: `irs_gross_income` plus `workers_compensation`, `alimony_income`, `miscellaneous_income` (gifts), `child_support_received`, and reported `ssi` / `tanf`. The comment records the `adds` chains so the next reader doesn't have to re-derive them.
- **`ChildSupportReceivedDependency`** (`dependencies/member.py`) — child support *received*, as income. Distinct from the existing `SnapChildSupportDependency`, which sends child support *paid* as an expense (`child_support_expense`); a household can report both, and collapsing them would count a payer's outflow as their income. Asserted side by side in tests so an edit can't merge them.
[...]

PR #1685, opened 2026-08-10
MFB-1312: adopt PolicyEngine's actual-receipt contract for SSI / TANF / SNAP / WIC

david-j-stoll:
## Context & Motivation

PolicyEngine used to feed **simulated** (would-be-eligible) SSI, TANF and SNAP into *other* programs' calculations — both as countable income and to confer categorical eligibility (SNAP categorical via SSI/TANF, WIC adjunctive, Head Start categorical, Medicaid's SSI category). Being *eligible* for a benefit is not the same as *receiving* it, so downstream eligibility and income were over-stated: a household PolicyEngine merely models as SSI-eligible was blocked from IL AABD by income they never got, and picked up categorical eligibility from a benefit they don't have.

As of `policyengine-us` **1.779.3** PE ties both to actual receipt. This adopts that contract across all four programs and every downstream consumer.

- Fixes [MFB-1312](https://linear.app/myfriendben/issue/MFB-1312/d-program-adopt-actual-receipt-vs-would-be-eligibility-contract-for)
- Unblocks [MFB-1315](https://linear.app/myfriendben/issue/MFB-1315) (KS programs with categorical scenarios) and [MFB-1260](https://linear.app/myfriendben/issue/MFB-1260) (MO SNAP)
- Related: #1606 (KS SNAP — deferred the SSI-categorical piece to here), #1681 ([MFB-1604](https://linear.app/myfriendben/issue/MFB-1604) `has_base_benefit` cleanup, merged — prerequisite), #1684 ([MFB-1571](https://linear.app/myfriendben/issue/MFB-1571) WIC income sources, open)
- [MFB-1382](https://linear.app/myfriendben/issue/MFB-1382) (SNAP/WIC income capture) was **cancelled**: PolicyEngine uses reported SNAP/WIC *amounts* for only two programs, neither of which MFB implements. So SNAP and WIC are two-state here (receipt boolean / take-up off) rather than the three-state shape the ticket originally described.

### The contract

| User reports | SSI / TANF | SNAP | WIC |
|---|---|---|---|
| Receiving, amount known | `ssi` / `tanf` = reported amount + `receives_X: true` | n/a — no amount capture | n/a — no amount capture |
| Receiving, no amount | `receives_X: true`, amount computed | `receives_snap: true`, amount computed | *(no input — see below)* |
| Receiving per the Current Benefits tile only | SSI: no member credited, `takes_up_ssi_if_eligible` held at default; TANF: `receives_tanf: true` | `receives_snap: true` | n/a |
| Not receiving | `takes_up_X_if_eligible: false` | `takes_up_snap_if_eligible: false` | *(no flag — see below)* |
| Display | `ssi_if_takes_up` / `tanf_if_takes_up` | `snap_if_takes_up` | `wic` (already ungated) |

## Changes Made

Three commits, one per workstream.

**1. Dependency layer** (`b333bde8`)

- New `dependencies/receipt.py` — single home for what "receiving" means, so all six inputs agree.
- New dependencies, all gated `min_pe_version = (1, 779, 3)`: `receives_ssi`, `receives_snap`, `receives_tanf`, `takes_up_ssi_if_eligible`, `takes_up_tanf_if_eligible`, and the `ssi_if_takes_up` / `snap_if_takes_up` / `tanf_if_takes_up` outputs (WIC keeps the ungated `wic` — see below). `TakesUpSnapIfEligibleDependency` already existed, defined-but-unwired with a hardcoded `True`; it now has a real `value()` and a version gate.
- `is_tanf_enrolled` is **not** sent — PE defaults it to `receives_tanf`.
- `Tanf` no longer requires the Current Benefits tile: a reported cash-assistance amount is itself proof of receipt. Without this we'd have sent `takes_up_tanf_if_eligible: false` alongside a household's own reported TANF amount.

**2. Program wiring** (`904d5553`)

- Shared `receipt_contract` bundle (next to `irs_gross_income`) applied to: SNAP (all 8 variants via `SNAP_BASE_INPUTS`), Head Start / Early Head Start, WIC, Medicaid, MSP, Lifeline, KS TANF, IL AABD, CO ANDCS, TX CEAP. PE requests are a single shared payload, so the bundle is applied broadly on purpose — behavior shouldn't depend on which programs a white label happens to enable.
- Three programs move to their would-be output: `Snap.pe_name → snap_if_takes_up`, `Tanf → tanf_if_takes_up`, `Ssi → ssi_if_takes_up`. `Wic` keeps `wic`, which is already ungated. **This half cannot ship separately** — once `takes_up_X_if_eligible: false` goes out, the receipt-gated fields read 0 for exactly the households the programs should be recommended to, and the frontend filters out anything valued at $0.
- `member.Ssi` now reaches all 7 SNAP states (only KS wired it before).
- **TX CEAP** migrated off `ssi_reported` / `use_reported_ssi`; `applicable_ssi` follows the `ssi` input.
- Deleted: `SsiReportedDependency`, `UseReportedSsiDependency`, `SsiAmountIfEligible`, the `spm.Snap` `1`-sentinel input and the `member.Wic` output dependency.

**3. Tests** (`9f99d390`)

- New `policyengine/tests/test_receipt_contract.py` — the cross-cutting wiring: every consumer adopts the bundle, each program reads the right output (including WIC staying ungated), every 1.779.3 field carries its gate, and the superseded reported-SSI inputs stay absent.
- New dependency tests covering member scoping (a spouse's SSI is not this member's), receipt from an amount without the tile, and the unattributable case.
- Updated the suites that pinned the old shape (`pe_name`, `ssi_reported`, the `snap: 1` sentinel).

**4. Guard against an unreadable output** (`46d26155`)

Found while responding to review. The `*_if_takes_up` fields are the **first version-gated `pe_outputs` in the codebase**, and gating an output is not like gating an input:

- a withheld *input* just leaves PolicyEngine to model the value itself — harmless, and the only case that existed before this PR;
- a withheld *output* is never returned, `Sim.value()` is a bare dict lookup, and `all_eligibility()` has no per-program error handling — so one unreadable program raises `KeyError` into `calc_pe_eligibility`'s catch-all, which returns an empty result. **Every** PolicyEngine program then drops off the screen, not just the gated ones.

Reproduced locally: at `pe_version="1.750.0"` the SNAP payload carries no `snap_if_takes_up` at all. Reachable two ways — an exact `PolicyEngineConfig` pin below a floor, or PE's `/versions/us` being unreachable while `/calculate` is healthy (separate endpoints; an unresolvable version withholds every gated field).

`_drop_unreadable_programs()` filters those programs out alongside the existing `can_calc()` pass and reports the drop to Sentry, so the failure degrades to "these are unavailable" instead of "no PolicyEngine programs at all". Also folds `pe_input`'s inline version resolution into the shared helper both paths now need.

**Scope of that degradation — 14 programs, not 4.** The gated outputs are declared on the federal base classes, but state variants inherit `pe_outputs`, so the drop set is every SNAP variant and every SSI variant:

```
snap_if_takes_up   snap, co_snap, il_snap, ks_snap, ma_snap, nc_snap, tx_snap, wa_snap   (8)
ssi_if_takes_up    ssi, ks_ssi, mo_ssi, tx_ssi, wa_ssi                                    (5)
tanf_if_takes_up   tanf                                                                   (1)
```

State TANF is **not** affected: `CoTanf` / `IlTanf` / `KsTanf` / `NcTanf` / `TxTanf` / `WaTanf` each override `pe_outputs` with their own `{st}_tanf` field, which carries no floor. Only the federal `tanf` calculator reads the gated output.

So if `/versions/us` is unreachable while `/calculate` is healthy, SNAP and SSI disappear across all their states simultaneously and silently from the user's side (omitted from results, not shown ineligible; only `missing_programs: true` is set). Still a large improvement on the pre-guard behavior — *every* PE program gone — and the Sentry `capture_message` names the withheld field, so it's diagnosable. But whoever sees that warning should expect two whole program families missing, not four programs. Version resolution failures are not cached, so every request re-attempts the fetch for the duration of an outage.

**5. SSI receipt is read from a reported amount only** (`224371b7`)

`receives_ssi` is person-level; the Current Benefits tile is household-level. A ticked tile with no dollar amount names no recipient, and with more than one plausible member the screener holds nothing to tell them apart — so no member is asserted as a recipient.

Guessing is not cheap. PolicyEngine treats `receives_ssi` as conclusive: measured against the live API, the flag alone flips `medicaid_category` to `SSI_RECIPIENT` with no demographic or income test — worth **$15,167** to a non-disabled 30-year-old at $90k income and **$10,222** to a 3-year-old. Crediting the wrong member both grants a benefit they don't qualify for and pushes their computed SSI into the income aggregate SNAP's income test reads, so a bad guess can add one benefit while reducing another.

Attributing it from a 65+/blind/disabled test would also put an SSI eligibility rule in the dependency layer, where it wouldn't track changes to SSI's actual rules.

Little is lost by not guessing. Every consumer tests `(ssi > 0) | receives_ssi` — `meets_snap_categorical_eligibility`, `is_ssi_recipient_for_medicaid` and its 209(b) variant. A tile-only household keeps `takes_up_ssi_if_eligible: true`, so PolicyEngine models their SSI, `ssi > 0`, and those tests fire anyway. The boolean only decides the case where PolicyEngine computes the amount as $0 (see Known limitations).

The tile is not discarded: `screen_reports_ssi_without_amount` holds take-up at PolicyEngine's default for the whole household, so a real recipient's SSI is never zeroed.

**WIC stays on the ungated `wic`.** `wic` is `wic_if_takes_up` gated on `takes_up_wic_if_eligible`, which defaults True and which we never send — and `receives_wic` is measurably inert — so the two fields return the same number for every payload we submit. There is no WIC take-up suppression for a would-be output to route around. Staying ungated also keeps WIC clear of the `1.779.3` floor, and so out of `_drop_unreadable_programs`.

## Testing

- **Migrations to run:** none
- **Configuration updates needed:** none. Requires the served PolicyEngine model to be ≥ 1.779.3 — see Deployment.
- **Environment variables/settings to add:** none
- **Automated:** `3145 passed, 8 skipped`. The 7 failures in `programs/management/commands/tests/test_import_all_program_configs.py` (a `content_hash` NOT NULL violation) are **pre-existing and unrelated** — they reproduce on untouched `origin/main`.
- **Manual testing steps** (each run twice: once reporting the benefit, once not):
  1. **KS SNAP Sc 8** — 1 person, age 45, disabled, SSI `$900`/mo, assets `$6,000`, ZIP `66102`/Wyandotte → **eligible, $1,080/yr**. Verified end-to-end against live PE: `meets_snap_categorical_eligibility: true`, `snap_if_takes_up` `$90.70`/mo.
  2. **KS SNAP Sc 22** — same, plus pension `$2,000`/mo → **eligible, $288/yr** (minimum allotment). Verified: categorical still fires even though PE computes the household's own SSI as `$0`. These are the two scenarios #1606 deferred.
  3. **Simulated suppression** — identical household that does *not* report SNAP but which PE would model as SNAP-eligible: Head Start / EHS categorical must **not** fire.
  4. **Reported-but-$0** — report receipt with income high enough that PE computes `$0`: categorical must still fire (via `receives_X`).
  5. **Regression guard** — a non-reporter still sees a would-be amount for SNAP / SSI / TANF / WIC rather than `$0`.
  6. **IL AABD** — age 67, disabled, `$0` income: previously `$0` (blocked by phantom simulated SSI), now **$10,153**.
  7. **Tile-only SSI, no amount** — tick the SSI tile, enter **no** SSI amount, 1 person aged 67, assets `$6,000`. Confirm no member is sent `receives_ssi: true`, `takes_up_ssi_if_eligible` stays `true` for everyone, and PolicyEngine's own modelled SSI still drives SNAP categorical eligibility. Then add a non-disabled working-age second adult and confirm neither member picks up the SSI-recipient Medicaid pathway.

> ### ⚠️ Read the categorical scenarios with this in mind
>
> SNAP categorical eligibility requires **every** member of the household to receive SSI/TANF — the federal pure-PA-household rule. Measured with the contract's take-up flags applied:
>
> | Household | `meets_snap_categorical_eligibility` |
> | -- | -- |
> | aged 67 reporting an SSI amount, alone | **true** |
> | 67 (SSI amount) + child 3 | false |
> | 67 (SSI amount) + adult 40 | false |
>
> This is why `ks/snap/spec.md` S8 is phrased "All Members Receive SSI" at household size 1. Multi-member households that look categorically eligible on `main` are firing off **simulated TANF** ($0 income + a child is TANF-eligible), which this contract deliberately suppresses — so those going to `false` is the intended change, not a regression. Note also that crediting a single member would not have been sufficient — the rule reads `spm_unit.all(receives_ssi)`.

## Deployment

- **Post-deployment scripts needed:** none
- **Production config updates:** none — **verified**. Both staging and prod ride PolicyEngine's `current` (1.784.3) unpinned, which satisfies the `1.779.3` floor, so every gated input is sent. Worth re-checking only if someone pins an exact version later: an exact pin below 1.779.3 silently withholds all of them and turns the contract into a no-op.
  ```
  heroku run --app cobenefits-api python manage.py shell -c \
    "from configuration.models import PolicyEngineConfig; print(repr(PolicyEngineConfig.current_version()))"
  ```
- **Admin updates needed:** none
- **Notify team/users of:** benefit **values change** for a large share of households — this is intended, but it is not a silent refactor:
  - **SNAP rises in all 7 states** for any household not reporting SSI (phantom simulated SSI stops counting as income). Measured `$62.50 → $298`/mo and `$243 → $546`/mo on two sweep households.
  - **IL AABD** `$0 → $10,153`, **CO OAP** `$456 → $12,384`, and state TANF (IL/KS/NC/TX) all rise, same cause.
  - **WIC eligibility narrows** in CO/IL/MA/NC/TX/MO: its adjunct test loses the computed-SNAP/TANF branch for non-reporters. The Medicaid branch and the income test are unaffected. The `wic` value itself is read from the same field as today — only the inputs feeding that test change.
  - Households reporting SNAP no longer have their SNAP pinned to `$1`/mo in PolicyEngine's income (the retired sentinel), so programs reading household benefit income shift slightly for them.

## Notes for Reviewers

Worth a close look: `dependencies/receipt.py` (the receipt rules and why each is drawn where it is), and `federal/pe/spm.py` / `federal/pe/member.py` for the output switches (and why WIC is not one of them).

Three ticket assertions turned out to be wrong when checked against the live private API; the ticket has been corrected:

1. **`applicable_ssi` is not inert.** `ssi_reported` + `use_reported_ssi` still drive it (11,928 → 6,000 with `ssi_reported=6000`). TX CEAP would *not* have silently broken — migrating it is cleanup, not a fix. What did change is IL AABD, which no longer reads it; only `takes_up_ssi_if_eligible` moves it.
2. **`receives_wic` is a no-op** — it changes nothing measurable, not even `wic` under `takes_up_wic_if_eligible: false`. Not wired. Together with our never sending `takes_up_wic_if_eligible`, this is why WIC needs no would-be output: there is no suppression to route around.
3. **An explicit `ssi`/`tanf` amount wins over the take-up flag** — the flag only suppresses PE's own computed value. So a household that reports an amount without ticking the tile is never silently zeroed.

On the bundle's **non-uniform version gating** (raised in review): flooring `member.Ssi` and `spm.Tanf` at 1.779.3 alongside the receipt fields would stop sending reported SSI/TANF amounts to older models — a regression, since those two predate the contract by years and are load-bearing today. The asymmetry is safe because *the suppressing half is the gated half*: below the floor we send amounts with no take-up flags, which is the pre-contract behavior, and there is no resolvable version at which we zero a benefit without also sending the evidence for it. The inverse shape is unreachable, and `test_the_suppressing_half_is_the_gated_half` pins that direction.

- **Known limitations:**
  - **A ticked SSI tile with no amount confers no categorical eligibility where PolicyEngine computes the household's SSI as $0.** Accepted, not a bug — please don't file it from QA. Such a household keeps `takes_up_ssi_if_eligible: true`, so PolicyEngine models their SSI and `(ssi > 0)` satisfies the categorical tests in the ordinary case; the gap is only the $0-computed one, where `receives_ssi` would have been the sole signal. Closing it needs per-member receipt capture at the tile, not an inferred recipient — the screener cannot currently say *who* in a multi-adult household receives SSI.
  - **SNAP receipt is tile-only** — there's no SNAP amount to fall back on, so an unticked tile reads as "not receiving".
  - TANF receipt is deliberately *not* derived into `CurrentBenefit`. "Cash Assistance Grant" is a broader label than TANF, and treating it as TANF receipt would drop the state's TANF program from those households' results via `already_has`.
  - **Pre-existing, not touched here:** `all_eligibility()` has no per-program error handling, so *any* calculator that raises empties the PE result for the whole screen — every PolicyEngine program at once, silently from the user's side (they're omitted from results, not shown as ineligible; the response only sets `missing_programs: true`). Commit `46d26155` removes the instance this PR would have introduced; the general fragility remains and probably deserves its own ticket.
- **Future considerations:**
  - Follow-ups from the ticket, intentionally not in this diff: IL AABD `spec.md` + scenario test, Head Start / EHS test directories (KS/TX/MA), and un-skipping the KS EHS/HS categorical Scenario 4 with its Known-limitation notes removed.
  - If SNAP/WIC amount capture is ever revived, the amount inputs belong in the `receipt_contract` bundle alongside `ssi` and `tanf`; `test_receipt_contract.py` pins that absence today so it can't drift unnoticed.
  - #1684 has merged, so WIC QA here runs against income-sensitive WIC.

## Summary by CodeRabbit

* **New Features**
  * Eligibility calculations now distinguish reported benefit receipt from modeled eligibility and take-up for SSI, SNAP, and TANF.
  * Programs including WIC, Medicaid, Head Start, Lifeline, MSP, and selected state benefits use consistent receipt information.
  * Benefit outputs now indicate when assistance is received or taken up.
* **Bug Fixes**
  * Improved handling of benefit receipts reported through income or benefit records.
  * Unsupported program versions are now excluded safely, with consistent version handling across calculations.

### DIFF
--- a/programs/programs/policyengine/calculators/dependencies/__init__.py
+++ b/programs/programs/policyengine/calculators/dependencies/__init__.py
@@ -68,3 +68,25 @@ wic_income = [
     member.Ssi,
     spm.Tanf,
 ]
+
+# PolicyEngine's actual-receipt contract (policyengine-us 1.779.3): countable income and
+# categorical eligibility follow the benefits a household reports receiving, not the ones
+# PolicyEngine simulates them as eligible for. See dependencies/receipt.py for the
+# reasoning; the amount inputs travel with the flags so a program adopting the contract
+# sends a coherent picture of each benefit.
+#
+# The set moves as a unit on purpose. Suppressing simulated SSI while leaving simulated
+# SNAP in place would just shift categorical eligibility onto the other phantom benefit,
+# and PE requests are a single shared payload — every program in a request sees these
+# inputs — so a partial adoption is a per-white-label inconsistency rather than a
+# narrower change.
+receipt_contract = [
+    member.Ssi,
+    member.ReceivesSsiDependency,
+    member.TakesUpSsiIfEligibleDependency,
+    spm.Tanf,
+    spm.ReceivesTanfDependency,
+    spm.TakesUpTanfIfEligibleDependency,
+    spm.ReceivesSnapDependency,
+    spm.TakesUpSnapIfEligibleDependency,
+]
--- a/programs/programs/policyengine/calculators/dependencies/__init__.py
+++ b/programs/programs/policyengine/calculators/dependencies/__init__.py
@@ -69,26 +69,17 @@ wic_income = [
     spm.Tanf,
 ]
 
-# PolicyEngine's actual-receipt contract (policyengine-us 1.779.3): countable income and
-# categorical eligibility follow the benefits a household reports receiving, not the ones
-# PolicyEngine simulates them as eligible for. See dependencies/receipt.py for the
-# reasoning; the amount inputs travel with the flags so a program adopting the contract
-# sends a coherent picture of each benefit.
+# PolicyEngine's actual-receipt contract: countable income and categorical eligibility follow
+# the benefits a household reports receiving, not the ones PolicyEngine simulates them as
+# eligible for. See dependencies/receipt.py.
 #
-# The set moves as a unit across *calculators* on purpose. Suppressing simulated SSI while
-# leaving simulated SNAP in place would just shift categorical eligibility onto the other
-# phantom benefit, and PE requests are a single shared payload — every program in a request
-# sees these inputs — so a partial adoption is a per-white-label inconsistency rather than a
-# narrower change.
+# Adopt it whole per calculator — suppressing one simulated benefit while leaving another just
+# shifts categorical eligibility onto the phantom that remains.
 #
-# The version gating within the bundle is deliberately NOT uniform: `ssi` and `tanf` predate
-# the contract by years and stay ungated, while the six receipt/take-up fields are floored at
-# 1.779.3. Flooring the amounts too would stop sending reported SSI/TANF to any older model —
-# a regression, since those amounts are load-bearing today (TX CEAP's income tier, SNAP
-# categorical from a reported amount). The asymmetry is safe because the *suppressing* half is
-# the gated half: below the floor we send amounts without take-up flags, which is exactly the
-# pre-contract behavior, and there is no version at which we suppress a benefit without also
-# sending the evidence for it. test_receipt_contract.py pins that direction.
+# The version gating is deliberately mixed: the amount inputs predate the contract and stay
+# ungated, so flooring them would stop sending reported SSI/TANF to older models. That is safe
+# because the gated half is the *suppressing* half — below the floor we send amounts without
+# take-up flags, never the reverse. test_receipt_contract.py pins that direction.
 receipt_contract = [
     member.Ssi,
     member.ReceivesSsiDependency,
--- a/programs/programs/policyengine/calculators/dependencies/__init__.py
+++ b/programs/programs/policyengine/calculators/dependencies/__init__.py
@@ -76,10 +76,11 @@ wic_income = [
 # Adopt it whole per calculator — suppressing one simulated benefit while leaving another just
 # shifts categorical eligibility onto the phantom that remains.
 #
-# The version gating is deliberately mixed: the amount inputs predate the contract and stay
-# ungated, so flooring them would stop sending reported SSI/TANF to older models. That is safe
-# because the gated half is the *suppressing* half — below the floor we send amounts without
-# take-up flags, never the reverse. test_receipt_contract.py pins that direction.
+# The version gating is deliberately mixed: the amount inputs stay ungated, since every
+# supported model reads them and flooring them would withhold reported SSI/TANF from older
+# ones. That is safe because the gated half is the *suppressing* half — below the floor we
+# send amounts without take-up flags, never the reverse. test_receipt_contract.py pins that
+# direction.
 receipt_contract = [
     member.Ssi,
     member.ReceivesSsiDependency,
--- a/programs/programs/policyengine/calculators/dependencies/__init__.py
+++ b/programs/programs/policyengine/calculators/dependencies/__init__.py
@@ -75,11 +75,20 @@ wic_income = [
 # reasoning; the amount inputs travel with the flags so a program adopting the contract
 # sends a coherent picture of each benefit.
 #
-# The set moves as a unit on purpose. Suppressing simulated SSI while leaving simulated
-# SNAP in place would just shift categorical eligibility onto the other phantom benefit,
-# and PE requests are a single shared payload — every program in a request sees these
-# inputs — so a partial adoption is a per-white-label inconsistency rather than a
+# The set moves as a unit across *calculators* on purpose. Suppressing simulated SSI while
+# leaving simulated SNAP in place would just shift categorical eligibility onto the other
+# phantom benefit, and PE requests are a single shared payload — every program in a request
+# sees these inputs — so a partial adoption is a per-white-label inconsistency rather than a
 # narrower change.
+#
+# The version gating within the bundle is deliberately NOT uniform: `ssi` and `tanf` predate
+# the contract by years and stay ungated, while the six receipt/take-up fields are floored at
+# 1.779.3. Flooring the amounts too would stop sending reported SSI/TANF to any older model —
+# a regression, since those amounts are load-bearing today (TX CEAP's income tier, SNAP
+# categorical from a reported amount). The asymmetry is safe because the *suppressing* half is
+# the gated half: below the floor we send amounts without take-up flags, which is exactly the
+# pre-contract behavior, and there is no version at which we suppress a benefit without also
+# sending the evidence for it. test_receipt_contract.py pins that direction.
 receipt_contract = [
     member.Ssi,
     member.ReceivesSsiDependency,
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -1,4 +1,5 @@
 from .base import Member
+from .receipt import SSI_INCOME_TYPE, member_reports_ssi, screen_reports_unattributed_ssi
 
 
 class AgeDependency(Member):
@@ -133,19 +134,37 @@ class Wic(Member):
     field = "wic"
 
 
+class WicIfTakesUp(Member):
+    """
+    PolicyEngine's would-be WIC entitlement — what the member is eligible for whether or
+    not they take it up. The WIC program reads this rather than ``wic``, which is
+    receipt-gated: ``takes_up_wic_if_eligible`` (or any downstream take-up suppression)
+    zeroes ``wic``, and every WIC calculator gates on that value being positive, so a
+    non-recipient would drop out of results entirely.
+    """
+
+    field = "wic_if_takes_up"
+    min_pe_version = (1, 779, 3)
+
+
 class Medicaid(Member):
     field = "medicaid"
 
 
 class Ssi(Member):
     """
-    SSI as both PE input and output.
+    The household's reported SSI amount, as PolicyEngine's person-level ``ssi`` input.
 
-    - If user reports SSI: use reported value
-    - If no reported SSI: return None so PE calculates eligibility
+    - If the member reports SSI: send the reported amount (annual; PE spreads it across
+      the months itself). PE uses it in place of its own computed SSI, so it counts as
+      income downstream and confers the categorical eligibility SSI receipt carries.
+    - Otherwise: None, so PE computes. Whether that computed amount is then *used*
+      depends on ``TakesUpSsiIfEligibleDependency`` — for a member who reports no SSI it
+      is suppressed, so PE's simulated SSI no longer counts as unearned income for
+      programs like IL AABD.
 
-    Warning: When PE calculates SSI, the value sometimes counts as unearned
-    income for downstream calculations (e.g., IL AABD).
+    An explicit amount wins over the take-up flag: PE's flag only suppresses its own
+    computed value.
     """
 
     field = "ssi"
@@ -156,8 +175,70 @@ class Ssi(Member):
     )
 
     def value(self):
-        ssi = self.member.calc_gross_income("yearly", ["sSI"])
+        ssi = self.member.calc_gross_income("yearly", [SSI_INCOME_TYPE])
         return None if ssi == 0 else ssi
+
+
+class SsiIfTakesUp(Member):
+    """
+    PolicyEngine's would-be SSI entitlement, independent of take-up. The SSI program
+    reads this rather than ``ssi``: with ``takes_up_ssi_if_eligible: False`` sent for
+    every non-recipient, ``ssi`` is 0 for exactly the people the program should be shown
+    to, and the frontend filters programs valued at $0 out of results.
+    """
+
+    field = "ssi_if_takes_up"
+    min_pe_version = (1, 779, 3)
+
+
+class ReceivesSsiDependency(Member):
+    """
+    Reported SSI receipt for this member.
+
+    Distinct from the ``ssi`` amount: a household can report receiving SSI while PE
+    computes their entitlement as $0 (high other income, say), and this is what keeps the
+    categorical eligibility that receipt confers — SNAP's SSI-recipient path, Medicaid's
+    SSI category — firing in that case.
+    """
+
+    field = "receives_ssi"
+    min_pe_version = (1, 779, 3)
+    dependencies = (
+        "income_type",
+        "income_amount",
+        "income_frequency",
+    )
+
+    def value(self):
+        return member_reports_ssi(self.member)
+
+
+class TakesUpSsiIfEligibleDependency(Member):
+    """
+    False for a member who reports no SSI, which stops PolicyEngine counting the SSI it
+    simulates for them as income they actually receive.
+
+    This is the lever behind the largest behavior change in the receipt contract: a
+    non-reporter whom PE models as SSI-eligible no longer carries phantom SSI income into
+    IL AABD, SNAP's income test, TX CEAP or the Medicaid/MSP SSI methodologies.
+
+    Left at PE's default (True) when the household reports SSI receipt that no member's
+    income stream accounts for — see ``screen_reports_unattributed_ssi``.
+    """
+
+    field = "takes_up_ssi_if_eligible"
+    min_pe_version = (1, 779, 3)
+    dependencies = (
+        "income_type",
+        "income_amount",
+        "income_frequency",
+    )
+
+    def value(self):
+        if member_reports_ssi(self.member):
+            return True
+
+        return screen_reports_unattributed_ssi(self.screen)
 
 
 class IsDisabledDependency(Member):
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -130,10 +130,6 @@ class MedicaidSeniorOrDisabled(Member):
     field = "is_optional_senior_or_disabled_for_medicaid"
 
 
-class Wic(Member):
-    field = "wic"
-
-
 class WicIfTakesUp(Member):
     """
     PolicyEngine's would-be WIC entitlement — what the member is eligible for whether or
@@ -387,40 +383,6 @@ class IsBlindDependency(Member):
         return self.member.visually_impaired or False
 
 
-class SsiReportedDependency(Member):
-    field = "ssi_reported"
-    dependencies = (
-        "income_type",
-        "income_amount",
-        "income_frequency",
-    )
-
-    def value(self):
-        return self.member.calc_gross_income("yearly", ["sSI"])
-
-
-class UseReportedSsiDependency(Member):
-    """
-    Tells PolicyEngine to count the household's *reported* SSI (ssi_reported)
-    instead of PE's *calculated* ssi wherever a program's countable income reads
-    applicable_ssi (the calculator opts in by including this in its pe_inputs).
-    We always send True so the screener's reported SSI drives the income test.
-
-    Person-level and global within a request: it flips applicable_ssi for every
-    program in that request, so only add it to calculators where reported SSI is
-    the correct measure.
-
-    min_pe_version gates this to the first model that defines use_reported_ssi;
-    sending it to an earlier model 400s the whole request.
-    """
-
-    field = "use_reported_ssi"
-    min_pe_version = (1, 742, 0)
-
-    def value(self):
-        return True
-
-
 class SsdiReportedDependency(Member):
     # Amount in "Social Security disability benefits (SSDI)"
     field = "social_security_disability"
@@ -447,10 +409,6 @@ class SsiCountableResourcesDependency(Member):
             ssi_assets = self.screen.household_assets / self.screen.num_adults()
 
         return int(ssi_assets)
-
-
-class SsiAmountIfEligible(Member):
-    field = "ssi_amount_if_eligible"
 
 
 class Andcs(Member):
--- a/programs/programs/policyengine/calculators/dependencies/member.py
+++ b/programs/programs/policyengine/calculators/dependencies/member.py
@@ -1,5 +1,5 @@
 from .base import Member
-from .receipt import SSI_INCOME_TYPE, member_reports_ssi, screen_reports_unattributed_ssi
+from .receipt import SSI_INCOME_TYPE, member_receives_ssi, screen_reports_unidentifiable_ssi
 
 
 class AgeDependency(Member):
@@ -194,7 +194,10 @@ class ReceivesSsiDependency(Member):
     Distinct from the ``ssi`` amount: a household can report receiving SSI while PE
     computes their entitlement as $0 (high other income, say), and this is what keeps the
     categorical eligibility that receipt confers — SNAP's SSI-recipient path, Medicaid's
-    SSI category — firing in that case.
+    SSI category — firing in that case. It is also the only signal for a household that
+    ticks the Current Benefits tile without entering an amount.
+
+    See ``member_receives_ssi`` for how a household-level tile is attributed to a person.
     """
 
     field = "receives_ssi"
@@ -203,10 +206,11 @@ class ReceivesSsiDependency(Member):
         "income_type",
         "income_amount",
         "income_frequency",
+        "age",
     )
 
     def value(self):
-        return member_reports_ssi(self.member)
+        return member_receives_ssi(self.screen, self.member)
 
 
 class TakesUpSsiIfEligibleDependency(Member):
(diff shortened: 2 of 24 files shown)


programs/programs/policyengine/calculators/dependencies/__init__.py
```diff
@@ -13,3 +13,25 @@
...
+receipt_contract = [
+    member.Ssi,
+    member.ReceivesSsiDependency,
+    member.TakesUpSsiIfEligibleDependency,
+    spm.Tanf,
+    spm.ReceivesTanfDependency,
+    spm.TakesUpTanfIfEligibleDependency,
+    spm.ReceivesSnapDependency,
+    spm.TakesUpSnapIfEligibleDependency,
+]
```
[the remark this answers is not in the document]
david-j-stoll: Re: mixed min_pe_version in receipt_contract
Verified against the current code. The observation is right, the proposed fix isn't — and chasing it turned up a real bug next door, now fixed in https://github.com/MyFriendBen/benefits-api/commit/46d26155d1f7ad7346b659f6c4db896b78756cc9.

Not applying the suggested change. Flooring member.Ssi and spm.Tanf at 1.779.3 would stop sending reported SSI/TANF amounts to any older model. Those two fields predate the contract by years and are load-bearing today (TX CEAP's FPG tier, SNAP categorical off a reported amount), so that's a regression, not a tightening. Removing the partial contract isn't on the table either.

The asymmetry is safe, and for a specific reason: the suppressing half is the gated half. Below the floor we send amounts with no take-up flags — exactly the pre-contract behavior — and there is no resolvable version at which we tell PolicyEngine to zero a benefit without also sending the evidence for it. The inverse (flags without amounts) is the dangerous shape and is unreachable. I've documented that in the bundle and pinned the direction with test_the_suppressing_half_is_the_gated_half, so a future edit can't quietly flip it.

The atomicity concern was pointing at something real, one layer over. The *_if_takes_up outputs are also gated at 1.779.3, and gating an output is not the same as gating an input:

a withheld input just leaves PolicyEngine to model the value itself — harmless;
a withheld output is never returned, Sim.value() is a bare dict lookup, and all_eligibility() has no per-program error handling — so one unreadable program raises KeyError into calc_pe_eligibility's catch-all and empties the PE result for the entire screen, every PolicyEngine program with it.
Reproduced locally: at pe_version="1.750.0" the SNAP payload carries no snap_if_takes_up at all. These are the first version-gated pe_outputs in the codebase, so nothing covered this path before. It's reachable two ways — an exact PolicyEngineConfig pin below a floor, or PE's /versions/us being unreachable while /calculate is healthy (separate endpoints; an unresolvable version withholds every gated field).

_drop_unreadable_programs() now filters those programs out alongside the existing can_calc() pass and reports the drop to Sentry, so the failure degrades to "these four are unavailable" instead of "no PolicyEngine programs at all". Six tests cover it, including the unresolvable-version case.

Worth flagging as pre-existing and out of scope: all_eligibility() having no per-program guard means any calculator raising takes down every PE program for that screen. This change removes the instance I introduced, not the general fragility.

programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -156,10 +171,72 @@ class Ssi(Member):
...
+    field = "receives_ssi"
+    min_pe_version = (1, 779, 3)
+    dependencies = (
+        "income_type",
+        "income_amount",
+        "income_frequency",
+    )
+
+    def value(self):
+        return member_reports_ssi(self.member)
```
catonph: This should also be true if they select the tile on step 8

david-j-stoll: Good catch — valid, fixed in 224371b7.

You're right that ticking the tile is a report of receipt, and we were discarding it: `receives_ssi` stayed False for everyone in a household that ticked SSI without entering an amount, so the categorical eligibility their receipt confers silently went missing. I'd shipped that as a documented limitation, which was the wrong call.

The tile now counts, but it's **attributed rather than broadcast**, because PolicyEngine treats `receives_ssi` as conclusive. Measured against the live private API, the flag alone flips `medicaid_category` to `SSI_RECIPIENT` with no demographic or income test applied — worth **$15,167** to a non-disabled 30-year-old at $90k income, and **$10,222** to a 3-year-old. So crediting every member would hand Medicaid to a toddler.

The rule now:

- a member reporting their own `sSI` amount receives SSI, unchanged;
- otherwise, if the tile is ticked and nobody reports an amount, credit only the members who could actually be the recipient — aged 65+, blind, or disabled, per 42 U.S.C. § 1382(a), which is also the screener data we have;
- a reported amount still explains the tile, so when anyone reports one the others stay non-recipients;
- tile ticked, no amount, and nobody aged/blind/disabled → unidentifiable, so take-up is left at PolicyEngine's default rather than zeroing a recipient we failed to spot.

Verified end to end through the real `pe_input()` payload: aged member, tile ticked, no amount, $6,000 assets → `meets_snap_categorical_eligibility: true` and **$298/mo**, where it previously returned `false` and $0.

**One thing worth knowing for QA, which surfaced while testing this.** Crediting the tile does *not* make a mixed household categorically eligible for SNAP. PolicyEngine requires **every** member to receive SSI/TANF — the federal pure-PA-household rule — so with the contract's take-up flags applied:

| Household | `meets_snap_categorical_eligibility` |
| -- | -- |
| aged 67 receiving SSI, alone | **true** |
| 67 (SSI) + child 3 | false |
| 67 (SSI) + adult 40 uncredited | false |
| 67 (SSI) + adult 40 **also credited** + child 3 | false |

That's also why `ks/snap/spec.md` S8 is phrased "All Members Receive SSI" at household size 1. Multi-member households that looked categorically eligible in my earlier probing were firing off **simulated TANF** ($0 income + a child = TANF-eligible), which this contract deliberately suppresses — so those cases correctly go to false, and QA shouldn't read that as a regression.

Full suite green (`2904 passed` across programs + screener; the 7 `test_import_all_program_configs` failures are pre-existing on `main`). Four new tests cover tile attribution, the non-plausible members, disability at any age, and the unidentifiable case.

programs/programs/policyengine/calculators/dependencies/receipt.py
```diff
@@ -0,0 +1,109 @@
...
+        return True
+
+    if not screen.has_base_benefit("ssi"):
+        return False
+
+    if screen.calc_gross_income("yearly", [SSI_INCOME_TYPE]) > 0:
+        # Somebody's amount already accounts for the tile.
+        return False
+
+    return _could_receive_ssi(member)
```
catonph: I'm very nervous about relying on this rough estimate to then determine this member receives SSI which will have real consequences to their results

catonph: Like maybe this isnt the member receiving SSI (maybe there are 2 members that *could*) and then they wont see SSI in their results

catonph: I know I originally flagged the other direction, but maybe we should rely exclusively on reported income for SSI? Because otherwise we risk hiding SSI for a member that doesn't already receive it....?

programs/programs/policyengine/calculators/dependencies/member.py
```diff
@@ -156,10 +161,68 @@ class Ssi(Member):
...
+
+
+class TakesUpSsiIfEligibleDependency(Member):
+    """
+    False for a member who reports no SSI, so PolicyEngine stops counting the SSI it
+    simulates for them as income they receive — the lever behind this contract's largest
+    behavior change, across IL AABD, SNAP's income test, TX CEAP and the Medicaid/MSP SSI
+    methodologies.
+
+    Left at PolicyEngine's default when the household's report can't be pinned on anyone —
```
catonph: Can we make this comment clearer - reading this alone I dont know what this field does

programs/programs/policyengine/calculators/dependencies/receipt.py
catonph: Also what if rules about SSI change and we have that sort of hidden within this helper function

programs/programs/policyengine/calculators/dependencies/spm.py
```diff
@@ -148,33 +172,17 @@ def value(self):
...
-    ssi (``ssi_reported`` / ``use_reported_ssi``), so this is the available signal.
+class SnapIfTakesUp(SpmUnit):
     """
+    PolicyEngine's would-be SNAP entitlement, independent of take-up — what the SNAP program
+    reads, since ``snap`` is zeroed for every household not already receiving it.
 
-    field = "snap"
+    Replaced a ``snap: 1`` receipt sentinel that pinned PolicyEngine's computed SNAP to $1/mo
+    for anyone reporting receipt, polluting the income other programs read.
+    """
```
catonph, line 192: delete this

programs/programs/policyengine/calculators/dependencies/spm.py
```diff
@@ -203,27 +211,76 @@ class Lifeline(SpmUnit):
...
+    field = "tanf_if_takes_up"
+    min_pe_version = (1, 779, 3)
+
+
+class ReceivesTanfDependency(SpmUnit):
+    """
+    Reported TANF receipt, conferring the categorical eligibility TANF carries (SNAP, Head
+    Start, WIC's adjunct test) even where PolicyEngine computes the household's own TANF as $0.
+
+    ``is_tanf_enrolled`` defaults to this, so it need not be sent separately.
```
catonph, line 267: Can we make this a little clearer: PolicyEngine derives is_tanf_enrolled from receives_tanf, so there's no separate enrollment input to send.

elliottwaissbluth:
fwiw here's a claude code review

## Code Review: PR #1685 — MFB-1312: adopt PolicyEngine's actual-receipt contract for SSI / TANF / SNAP / WIC

### Summary
A careful, well-tested implementation of the receipt contract, verified against the ticket's requirements (including the 2026-08-10 corrections). Every file in the diff was read, and the changed test suites were run against a local Postgres: all 245 dependency/contract tests pass. The 12 failures seen in the broader PE suites are purely environmental — they require live access to PolicyEngine's `/versions/us` (see Suggestion #4). Every ticket item for sub-issues 1 and 2 is delivered: take-up flags wired and version-gated, `receives_ssi` added with per-member attribution, the retired SSI bridge fully removed (grep-verified — no leftover references outside docstrings), `Ssi` reaches all 7 SNAP states via `SNAP_BASE_INPUTS`, TX CEAP migrated, and the four programs read `X_if_takes_up`. Also confirmed the `wa_snap.base_program` prerequisite landed (migration 0165 in #1681) — without it the SNAP take-up flag would have zeroed real WA recipients.

### Critical Issues
None found.

### Suggestions

| # | File | Line | Suggestion | Category |
|---|------|------|------------|----------|
| 1 | `dependencies/tests/test_member.py` | 1659 | The docstring justifying tile attribution says "one credited member is enough for the cross-program signal (SNAP's categorical eligibility fires off a single member)." That's wrong for SSI: PE's `meets_snap_categorical_eligibility` uses `spm_unit.all(receives_ssi)` — a pure-PA rule requiring **every** member (checked against the PE source; the PR description itself states the all-members rule). The *behavior* is correct — a mixed household shouldn't get pure-PA categorical — but the stated rationale is inverted for SSI and could mislead someone into "fixing" attribution later. TANF is the unit-level single-signal case. | Correctness (docs) |
| 2 | `dependencies/receipt.py` | 91 | Tile-only attribution credits **every** plausible member: two aged/disabled members in a household that ticked the tile with no amount both get `receives_ssi: true`, and the tests show that flag alone conclusively confers the SSI-recipient Medicaid pathway (no income test). If only one actually receives SSI, the other is over-granted Medicaid — the same class of error this ticket exists to remove, just via reported receipt. The symmetric edge exists too: one member's amount "explains the tile," so a second genuine recipient without an amount gets `takes_up: false`. Both are inherent screener ambiguity and reasonable defaults, but worth an explicit note in the known-limitations list (only the unattributable case is documented) and a QA case each. | Correctness (edge case) |
| 3 | `dependencies/spm.py` | 231, 234 | `Tanf` no longer requires the tile — a `cashAssistance` amount alone now asserts `receives_tanf` and pins the `tanf` input, which drives `tanf_if_takes_up` to 0. The known-limitation note calls this "harmless behind `already_has`," but TANF receipt is deliberately *not* derived into `CurrentBenefit` — so an amount-only household has TANF silently vanish from results (value 0, filtered) with no "already receiving" badge. Probably the right outcome (don't offer what they have), but the "harmless behind already_has" claim doesn't hold for exactly these households; worth stating and QA-ing. Related: `cashAssistance` ("Cash Assistance Grant") is a broader UI label than TANF, and an amount alone now also fires SNAP's TANF categorical eligibility. | Correctness (edge case) |
| 4 | `tests/test_pe_input.py` | 284 | Twelve existing tests (this one, plus TX/MO `pe_input` integration tests) now assert gated fields like `snap_if_takes_up` in the payload, which requires a live `/versions/us` resolution — they fail offline and will fail CI during a PE `/versions` outage. The new tests in this PR mock the resolution correctly; consider passing `resolved_version=`/mocking in these updated ones too, or accept the live-API convention explicitly. | Test reliability |
| 5 | `programs/policyengine/versions.py` | 48 | Post-contract, an unresolvable `/versions/us` doesn't just withhold optional inputs anymore — `_drop_unreadable_programs` removes SNAP/SSI/TANF/WIC from every screen. With a 1-hour cache TTL and no negative-result caching, an outage longer than the TTL degrades four core programs. Consider keeping a last-known-good resolution (stale-while-revalidate) as a follow-up. The guard itself is right — far better than the screen-wide KeyError it prevents. | Resilience |
| 6 | `dependencies/member.py` | 620 | `SocialSecurityIncomeDependency`'s docstring still points to the removed `SsiReportedDependency`. | Maintainability |

### What Looks Good
- Resolving the PE version once in `calc_pe_eligibility` and threading it through both the drop pass and payload build eliminates the two-resolution race cleanly, and `_drop_unreadable_programs` is exactly the right guard for the first gated *outputs* in the codebase.
- The asymmetric version gating (suppressing half floored, amount half ungated) is well-reasoned and pinned by a test that explains *why*.
- The unidentifiable-SSI fallback honors the ticket's "unknown ≠ not receiving" constraint.
- `receipt.py` is a model of documented trade-offs, and the cross-cutting `test_receipt_contract.py` (every consumer adopts the bundle, the four programs read would-be outputs, the retired bridge stays gone) is the kind of wiring test that keeps a shared-payload architecture honest.
- Security-wise nothing concerning: no injection surface, and the Sentry message carries program names only.

### Verdict
**Approve with minor changes** — #1's docstring fix should land before merge since it misdocuments the load-bearing federal rule; #2 and #3 need only documentation/QA-matrix acknowledgment, not code. The open pre-flight from the ticket (prod PE pin) is reported cleared as of 2026-08-10. Remaining sub-issue-3 work (IL AABD spec/test, Head Start test dirs, KS Scenario 4 un-skip, per-state QA) is explicitly deferred and tracked, which matches the ticket.

Sources: [MFB-1312](https://linear.app/myfriendben/issue/MFB-1312/d-program-adopt-actual-receipt-vs-would-be-eligibility-contract-for) · [PR #1685](https://github.com/MyFriendBen/benefits-api/pull/1685)

catonph (approved):

PR #1687, opened 2026-08-11
MFB-1602: fail-loud county validation guard in import_program_config

david-j-stoll:
## Context & Motivation

Navigator `counties` filters are an exact-string match against the county the screener stores (`Screen.county`), which comes from each white label's `counties_by_zipcode` map. The suffix convention is **per-white-label** (`"Jackson County"` for MO/CO/WA/KS/NC, bare `"Cook"` for IL/TX). When a config lists a county in the wrong convention, the filter silently never matches and the navigator referral is dropped with no error — the class of bug documented in MFB-1602 (found in CO/WA/MO).

This PR adds the **fail-loud import-time guard** suggested in MFB-1602 so a mismatched county name errors at import instead of silently shipping a dead referral.

- Related Linear ticket: MFB-1602 (guard — this does **not** fully close it; the `initial_config.json` normalization is separate)
- Related Linear ticket: MFB-1624 (the data-only cleanup that fixed the already-shipped bare rows in the DB — already applied to staging + production)

## Changes Made

- `programs/management/commands/import_program_config.py`:
  - `_get_valid_county_names(white_label)` — returns the valid county-name set from the white label's `counties_by_zipcode` `Configuration` row. Filters `active=True` and orders `-id` for a deterministic read. Raises `CommandError` (chaining the cause) if that config is malformed JSON, and raises if it isn't a JSON object (array/string/number/bool); returns `None` (validation skipped) when the row is absent, `null`, or an empty object.
  - `_validate_counties(config, white_label, existing_program, overriding)` — validates each navigator `counties` entry against that set and raises `CommandError` (before any writes, and in `--dry-run`) listing every offending name with a **symmetric** `did you mean 'X'?` hint (suffixed↔bare); non-string entries are reported as legible errors rather than crashing. It is called **only after the import is known to proceed** (skipped for the "program already exists, no `--override`" no-op path) and validates **only the navigators this run will actually (re)create** — via `_navigator_counties_written`, which mirrors `_delete_program_and_related`'s retention criterion — so it never blocks over counties the existing-navigator path ignores.
  - Guard is **convention-driven, not "always suffix"**: bare-map white labels accept bare names, and legit bare entries like MO `St. Louis City` pass automatically.
- `programs/management/commands/tests/test_import_program_config.py` — guard tests:
  - wrong convention (bare name, suffixed map) → raises, nothing created
  - wrong convention under `--dry-run` → still raises, nothing created
  - matching convention → imports and links the county
  - bare-convention map accepts bare / rejects suffixed (asserts the reverse suggestion)
  - no `counties_by_zipcode` config → validation skipped, import proceeds
  - non-string county entry (e.g. `null`) → raises a legible `CommandError`, not `AttributeError`
  - `counties_by_zipcode` config that isn't a JSON object → raises a legible `CommandError`
  - pre-existing navigator + bad county on a non-override import → does **not** block (counties aren't written on that path)

## Testing

- Migrations to run: none
- Configuration updates needed: none (reads the existing `counties_by_zipcode` config; no schema change)
- Manual testing steps:
  - `python manage.py test programs.management.commands.tests.test_import_program_config` → `Ran 14 tests ... OK`
  - Or import a config whose navigator lists a wrong-convention county (e.g. bare `"Jackson"` on `mo`) and confirm it errors with a "did you mean 'Jackson County'?" message and creates nothing.

## Deployment

- Admin updates needed: none
- Notify team/users of: nothing user-facing. Note the county data cleanup this complements (MFB-1624) was already applied to staging + production separately.

## Notes for Reviewers

- Addressed review feedback across several rounds:
  - **CodeRabbit**: reverse (suffixed→bare) suggestion, the `active=True` + deterministic config lookup, and the `--dry-run` guard test. **Declined** adding bare forms of suffixed keys to the valid set — that would defeat the guard (a bare name never matches the suffixed household county at runtime, `"Jackson County" in "Jackson"` is `False`), letting a silently-broken navigator import succeed.
  - **@catonph (round 1)**: removed ticket refs from committed code; non-string county entries are reported legibly instead of crashing; simplified the suggestion logic; and the guard validates only the counties this run actually writes (skips already-existing navigators whose counties are ignored).
  - **@catonph (round 2)**: malformed `counties_by_zipcode` JSON now raises `CommandError` (cause chained) and a non-object value raises rather than reaching `.values()`; and the validation is aligned with the navigator write plan — invoked only once the import is known to proceed, and scoped to navigators that will be (re)created using the same retention criterion as `_delete_program_and_related`.
- Known limitations:
  - Scoped to `import_program_config` and validates only navigator `counties` (the sole place this command imports counties). Other county-writing paths — `bulk_import` (translations) and `add_counties` — are not guarded here.
  - When a white label has no (or an empty) `counties_by_zipcode` config, validation is skipped (warns) so a fresh environment isn't blocked before `add_config` runs.
- Future considerations:
  - Normalizing the seed `initial_config.json` files (suffix CO/WA/MO, drop the `["Denver","Denver County"]` workaround, keep `St. Louis City` bare) is still needed to fully close MFB-1602.
  - `_delete_program_and_related` checks the legacy `Navigator.programs` M2M while `_import_navigators` associates via the newer `ProgramNavigator` through-model, so its "keep navigators shared with other programs" retention effectively never triggers on the current association path — looks like a pre-existing bug worth a separate ticket.
  - The same validation could be extended to the `bulk_import` path.

## Summary by CodeRabbit

* **New Features**
  * Added validation for navigator county names during program configuration imports.
  * County names are checked against the active configuration, with helpful suggestions for mismatches.
  * Validation also applies to dry runs and links matching counties automatically.

* **Bug Fixes**
  * Imports now handle missing or invalid county configuration safely.
  * Malformed county entries produce clear errors, while existing navigators are not unnecessarily revalidated.


programs/management/commands/tests/test_import_program_config.py
```diff
@@ -221,3 +222,135 @@ def test_transaction_rollback_on_category_creation_error(self):
 
         Path(config_file).unlink()
         Path(invalid_config_file).unlink()
+
+    # --- MFB-1602: navigator county validation guard ---
```
catonph: No ticket numbers

programs/management/commands/import_program_config.py
```diff
@@ -183,6 +189,77 @@ def handle(self, *args: Any, **options: Any) -> None:
...
+        config_obj = (
+            Configuration.objects.filter(name="counties_by_zipcode", white_label=white_label, active=True)
+            .order_by("-id")
+            .first()
+        )
+        if config_obj is None:
+            return None
+
+        data = config_obj.data
+        if isinstance(data, str):
```
catonph, line 209: **`data` is always a `str` here, and `not data` silently skips validation on a populated map.**

`OrderedJSONField.get_prep_value` does `json.dumps`, so Postgres stores a JSON *string*; `from_db_value` `json.loads` returns that string back. I confirmed against a test DB — after `Configuration.objects.create(data={"11111": {...}})`, a refetch gives:

```
type(obj.data) -> <class 'str'>
repr(obj.data) -> '{"11111": {"Jackson County": "Jackson County"}}'
```

So `isinstance(data, str)` will always be true.

Two concrete issues in these four lines:

1. `if not data: return None` conflates "map is empty" with "skip validation entirely." Returning `None` makes the caller print the *no config found* warning even though the row exists. Those are different operator situations and deserve different messages.
2. A malformed row raises a bare `JSONDecodeError` traceback rather than a `CommandError`, which is the one failure mode this PR is specifically about making legible. Wrapping it keeps the fail-loud contract intact:

```python
try:
    data = json.loads(data) if isinstance(data, str) else data
except json.JSONDecodeError as e:
    raise CommandError(
        f"'counties_by_zipcode' config for white label '{white_label.code}' is not valid JSON: {e}"
    )
```

Related: since `Optional[set]` already overloads `None` as "no config" while `set()` would mean "empty map," and the empty case is currently also treated as skip, the tri-state isn't buying anything as written.

programs/management/commands/import_program_config.py
```diff
@@ -183,6 +189,77 @@ def handle(self, *args: Any, **options: Any) -> None:
...
+        for nav_config in config.get("navigators", []):
+            if not isinstance(nav_config, dict):
+                continue
+            external_name = nav_config.get("external_name", "<unnamed>")
+            for county_name in nav_config.get("counties", []) or []:
+                if county_name in valid_names:
+                    continue
+                suggestion = ""
+                suffixed = f"{county_name} County"
+                bare = county_name[: -len(" County")] if county_name.endswith(" County") else None
```
catonph: **A `None` entry in `counties` crashes with an unhandled `AttributeError` instead of a `CommandError`.**

Verified — a config with `"counties": [null]` produces:

```
AttributeError: 'NoneType' object has no attribute 'endswith'
```

The loop guards `isinstance(nav_config, dict)` on line 239 but never type-checks the county entry itself, so a non-string slips through to `.endswith()`. Same for an int or a nested list. Given the whole point of this guard is turning silent config errors into legible ones, we should probably do something other than a raw traceback from the validator.

A type check folded into the problems list handles it and keeps the error style consistent:

```python
for county_name in nav_config.get("counties", []) or []:
    if not isinstance(county_name, str):
        problems.append(f"navigator '{external_name}': {county_name!r} is not a string")
        continue
    if county_name in valid_names:
        continue
```

Minor, same lines: the suffix/bare suggestion is ~10 lines of branching for one hint string. Computing the opposite-convention form once collapses it symmetrically:

```python
other = (
    county_name[: -len(" County")]
    if county_name.endswith(" County")
    else f"{county_name} County"
)
suggestion = f" (did you mean '{other}'?)" if other in valid_names else ""
```

That also drops the `bare and ...` truthiness guard, which exists only to handle a county literally named `"County"`.

programs/management/commands/import_program_config.py
```diff
@@ -183,6 +189,77 @@ def handle(self, *args: Any, **options: Any) -> None:
...
+            self.stdout.write(
+                self.style.WARNING(
+                    f"  Warning: no 'counties_by_zipcode' configuration found for white label "
+                    f"'{white_label.code}'; skipping navigator county validation."
+                )
+            )
+            return
+
+        problems = []
+        for nav_config in config.get("navigators", []):
```
catonph, line 278: **Counties are validated for *every* navigator, but `_import_navigators` only ever persists them for newly-created ones — so this can block an import over a field that path ignores.**

In `_import_navigators` (line ~836, outside this diff), the county-writing block sits inside the `else` branch — the one taken only when `Navigator.objects.filter(external_name=...).first()` misses. An existing navigator takes the `if` branch, gets appended to `navigators_to_associate`, and its `counties` from the config are never read.

I confirmed both halves against a test DB with a pre-existing navigator named `preexisting`:

- config lists `"counties": ["Jackson"]` (bad) → **raises** `Invalid county name(s) for white label 'test_wl'`
- config lists `"counties": ["Jackson County"]` (valid) → import succeeds, and `nav.counties` is still `[]` afterward

So an operator hits the error, corrects the name as the message instructs, re-runs, sees a green import — and the county still isn't attached. The guard is pointing at real config rot, but the remediation it implies doesn't exist on that path.

Two defensible resolutions:

1. **Validate only what gets written** — skip entries whose navigator already exists, mirroring `_import_navigators`' own branch. Narrow, but silently tolerates bad names in configs for existing navigators.
2. **Keep validating everything and note the gap** — still raise, but warn when the navigator already exists that `counties` won't be applied (or make the existing-navigator path update counties too).

I'd lean toward 2: the bad name in the config is genuinely worth surfacing, and the honest fix is that the existing-navigator path arguably *should* sync counties. Either way the current combination is the one state that misleads.

catonph (commented):
Three inline findings from an empirical pass (ran the branch tests — 11/11 pass — and probed each case against a real test DB). Finding 1 from my review (the 31 seed-config entries that now hard-fail) is intentional per the ticket and left out here.

catonph (changes requested):

catonph (approved):

PR #1689, opened 2026-08-11
MFB-1603: default legal links in base config (fixes empty KS/MO/WA links)

david-j-stoll:
## Context & Motivation

KS, MO, and WA never overrode `privacy_policy` or `consent_to_contact`, so all three inherited an **empty-string** default from `base.py`. The result is a link with an empty `href` — it looks like a link and goes nowhere — in the footer, the step-1 disclaimer, and the sign-up consent copy. These are the legal links.

**KS and WA are live in production with `{"en-us": ""}` in their Configuration rows right now.** This predates the MO work; it surfaced during MO staging QA but is not MO-specific and is not a launch blocker introduced by MO.

Nothing caught it, for two reasons worth naming:

1. `base.py` supplied a value, so no error or missing-key warning ever surfaced.
2. `health_check._check_links` skips empty links outright (`if link == BLANK_TRANSLATION_PLACEHOLDER or link == "": continue`), so the link checker cannot flag this class of failure.

- Fixes [MFB-1603](https://linear.app/myfriendben/issue/MFB-1603/privacy-policy-and-terms-footer-links-are-empty-for-mo-ks-and-wa)
- Related PR: MyFriendBen/benefits-calculator#2180 (frontend fallback hardening — the two together are the full fix)

## Changes Made

The fix is entirely in the base class. Per review feedback, **no state config files are touched** — state configs stay minimal and carry only deliberate overrides, matching how `referrer_data` and the income/expense options already work.

- **`base.py`** — `consent_to_contact` and `privacy_policy` default to the generic MyFriendBen pages (en-us + es) instead of empty strings. KS, MO, and WA inherit working links with no per-state declaration, and any future white label degrades to a valid page rather than a dead one.
- **`_template.py`** — documents both keys as inherited, with a commented-out override example, matching how the template already presents other base-supplied fields (`income_options_by_category`, `expense_options_by_category`).
- **`configuration/tests.py`** — new `TestLegalLinkConfiguration` asserting, for every white label in `white_label_config`, that the **effective** value of both keys (inherited or overridden) has an `en-us` entry and that every locale maps to an absolute `https` URL.

The test parses each URL rather than prefix-matching `"https://"`, so `"https://"` and `"https:///privacy-policy/"` — which name no host and reach the browser as dead links — fail, while a valid uppercase scheme passes.

No UTM parameters: CESN is the only white label that appends them, which reads as partner-network attribution rather than a state pattern.

## Testing

- **Migrations to run:** none
- **Configuration updates needed:** `python manage.py add_config ks mo wa` (or `--all`). These values reach the frontend as `Configuration` rows written by `add_config`, so merging alone changes nothing at runtime.
- **Environment variables/settings to add:** none
- **Manual testing steps:**
  1. `pytest configuration/` — 47 passed locally
  2. `black --line-length 120 --check configuration/` — clean (`_template.py` is Black-excluded in `pyproject.toml`; the change there is comments only)
  3. Confirm inheritance resolves for the three affected states — each should return the generic en-us + es URLs:
     ```python
     from configuration.white_labels import white_label_config
     for c in ("ks", "mo", "wa"):
         print(c, white_label_config[c].privacy_policy, white_label_config[c].consent_to_contact)
     ```
  4. After `add_config`, `GET /api/configuration/ks/` should show the real URLs where it currently returns `{"en-us": ""}`

## Deployment

- **Post-deployment scripts needed:** none beyond `add_config --all`, which `deploy-staging.yml` and `deploy-production.yml` already run unconditionally on every deploy. No manual step required.
- **Production config updates:** covered by that automatic `add_config --all` on the production deploy. Verify afterward that ks/mo/wa return the real URLs.
- **Admin updates needed:** none. Worth knowing a Django admin edit would *not* have been a durable fix: `add_config` writes with `update_or_create(defaults={"data": ...})`, so any admin edit to these rows is overwritten on the next deploy.
- **Notify team/users of:** KS and WA were serving dead privacy policy / terms links in production until this deploys, which may matter to whoever owns legal/compliance sign-off.

## Notes for Reviewers

- **Please confirm the URL choice.** The generic MyFriendBen pages are now the default for every white label that does not override. All four URLs return 200. If KS, MO, or WA need state-specific policy pages, they should override in their own config.
- **Known limitations:**
  - Only `en-us` and `es` are populated, matching every other white label. The other 12 supported locales fall back to `en-us` in the frontend — pre-existing behavior, not a regression.
  - Inheriting by default means a new white label gets working links without anyone deciding the generic MyFriendBen terms are legally right *for that partner*. The links resolve; they may name the wrong entity. That belongs on the new-white-label checklist rather than in a test, given the minimize-state-config convention.
  - The test reads config classes, so it cannot see a value edited to empty through Django admin. `health_check` still skips empty links, so that path stays uncovered; the config test catches the realistic case at CI time, before deploy.
- **Future considerations:**
  - `co`, `il`, `ma`, `nc`, `tx`, and `_default` still declare both keys explicitly with these exact generic URLs — the same redundancy just removed from ks/mo/wa. Fully applying the minimize-state-config convention means deleting those too, leaving only `cesn` (whose UTM-tagged URLs are a real override). Left out of this PR deliberately: outside MFB-1603 and it touches five live states. Happy to open a follow-up.
  - `_default` currently has no `privacy_policy`/`consent_to_contact` rows in production at all, meaning `add_config --all` has not run since those keys were added to `_default.py`. The next deploy creates them. No action needed — and `_default` was never broken precisely because a *missing* key hits the frontend fallback that an empty string bypassed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default privacy policy and contact-consent links for English and Spanish.
  * Ensured legal links use valid, absolute HTTPS URLs with an English fallback.
  * Added validation to confirm all white-label configurations provide required legal links.

* **Documentation**
  * Documented how to override legal links for individual languages.


configuration/white_labels/ks.py
```diff
@@ -1040,6 +1040,21 @@ def get_white_label(self) -> WhiteLabel:
...
+    consent_to_contact = {
+        "en-us": "https://www.myfriendben.org/terms-and-conditions/",
+        "es": "https://www.myfriendben.org/terminos-condiciones/",
+    }
+
+    privacy_policy = {
+        "en-us": "https://www.myfriendben.org/privacy-policy/",
+        "es": "https://www.myfriendben.org/privacidad/",
+    }
+
```
catonph: Don't need these now that you set it in the base

configuration/white_labels/mo.py
```diff
@@ -1435,6 +1435,21 @@ def get_white_label(self) -> WhiteLabel:
...
+    consent_to_contact = {
+        "en-us": "https://www.myfriendben.org/terms-and-conditions/",
+        "es": "https://www.myfriendben.org/terminos-condiciones/",
+    }
+
+    privacy_policy = {
+        "en-us": "https://www.myfriendben.org/privacy-policy/",
+        "es": "https://www.myfriendben.org/privacidad/",
+    }
+
```
catonph: Don't need these now that you set it in the base

configuration/white_labels/wa.py
```diff
@@ -1295,6 +1295,21 @@ def get_white_label(self) -> WhiteLabel:
...
+    consent_to_contact = {
+        "en-us": "https://www.myfriendben.org/terms-and-conditions/",
+        "es": "https://www.myfriendben.org/terminos-condiciones/",
+    }
+
+    privacy_policy = {
+        "en-us": "https://www.myfriendben.org/privacy-policy/",
+        "es": "https://www.myfriendben.org/privacidad/",
+    }
+
```
catonph: Don't need these now that you set it in the base

configuration/tests.py
```diff
@@ -150,3 +152,64 @@ def test_get_feature_flags_handles_empty_feature_flags(self):
...
+    KS, MO, and WA each shipped with these keys unset and inherited an empty-string
+    default, which renders as a link with an empty href — visibly a link, silently going
+    nowhere. Nothing caught it: the base class supplied a value, so no error was raised,
+    and the health check skips empty links entirely.
+
+    Both keys now have real defaults in the base class, so a missing override degrades to
+    the generic MyFriendBen pages instead of a dead link. These tests additionally require
+    each white label to declare both keys in its own class body, so that using the generic
+    pages stays a deliberate choice — a new white label cannot inherit legal links by
+    accident.
```
catonph: This isn't the pattern we are currently using and I think we should stay consistent: we try to minimize anything in the state-level configuration files unless a deliberate override is necessary

catonph (changes requested):
See in-line comments!

configuration/white_labels/_template.py
```diff
@@ -304,18 +304,21 @@ def get_white_label(self) -> WhiteLabel:
     }
```
catonph: Do you mind removing this while you're in here

configuration/white_labels/base.py
```diff
@@ -467,13 +467,20 @@ def get_white_label(self) -> WhiteLabel:
     category_benefits = {}
 
     # Links to consent/terms pages, keyed by language code.
+    # These default to the generic MyFriendBen pages rather than empty strings: an empty
+    # value renders as a link with an empty href, which is what KS, MO, and WA shipped for
```
catonph: Let's remove this comment

configuration/white_labels/base.py
```diff
@@ -467,13 +467,20 @@ def get_white_label(self) -> WhiteLabel:
...
+    # months. White labels inherit these and override only if they have their own policy
+    # pages. configuration.tests.TestLegalLinkConfiguration checks the effective value for
+    # every white label, so an override that is empty or relative fails CI.
```
catonph: This part is also unnecessarily verbose - this is a pattern we use all the time

configuration/tests.py
```diff
@@ -150,3 +152,50 @@ def test_get_feature_flags_handles_empty_feature_flags(self):
...
+    KS, MO, and WA each shipped with these keys unset and inherited an empty-string
+    default, which renders as a link with an empty href — visibly a link, silently going
+    nowhere. Nothing caught it: the base class supplied a value, so no error was raised,
+    and the health check skips empty links entirely.
+
+    Both keys now default to the generic MyFriendBen pages in the base class, so a white
+    label that does not override them inherits working links. These tests assert the
+    effective value for every white label, inherited or overridden, so a white label that
+    points these at an empty or relative value fails CI instead of shipping a dead link.
+    """
```
catonph, line 161: Unnecessarily verbose doc string

david-j-stoll: Made it one sentence

catonph (approved):

catonph (approved):

PR #1698, opened 2026-08-13
Remove stale Trump Account availability warning (7 states)

catonph:
## Context & Motivation

Every Trump Account config except MO carries this warning:

> Trump Accounts are expected to become available on or after July 4, 2026. Check IRS.gov/trumpaccounts for updates.

That date has passed and the premise is wrong. Treasury's portal at `trumpaccounts.gov` is live and accepting Form 4547 submissions today, so the warning now tells users to wait for something they can already do.

The original text also conflated two milestones: July 4, 2026 was when **contributions** begin, not when accounts became available. Activation emails have been going out in phases to people who filed earlier.

MO deliberately omits this warning (flagged during MO's discovery review as stale — do not copy). This brings the other seven in line with MO rather than the reverse.

Sources: [Treasury — Launch of the Trump Accounts App](https://home.treasury.gov/news/press-releases/sb0508), [Treasury/IRS proposed regulations](https://www.irs.gov/newsroom/treasury-irs-issue-proposed-regulations-on-how-to-open-initial-trump-accounts-under-the-one-big-beautiful-bill)

## Changes Made

Removed the top-level `warning_message` block from seven configs: `co`, `il`, `ks`, `ma`, `nc`, `tx`, `wa`. Seven lines deleted per file, nothing else touched.

## Testing

- Migrations to run: none
- Configuration updates needed: see Deployment — the config change alone is **not** sufficient
- Manual testing: removed the row in the local DB and confirmed the warning no longer renders

## Deployment

**A re-import will not remove the existing warning.** `_delete_program_and_related` only deletes a warning message when `warning.programs.count() == 1`. This row (`trump_account_notices`) is shared by all seven programs, so every per-state override re-import keeps it.

It must be deleted directly in each environment:

```bash
heroku run --app cobenefits-api-staging -- \
  python manage.py shell -c "from programs.models import WarningMessage; \
  WarningMessage.objects.filter(external_name='trump_account_notices').delete()"
```

Then the same on `cobenefits-api` for production. Deleting the shared row clears it for all seven white labels at once.

## Notes for Reviewers

The `programs.count() == 1` guard is correct behavior — it protects shared entities from a single program's re-import. It just means a shared row can never be cleaned up through the config path, which is worth knowing for any future shared-entity removal.

## Summary by CodeRabbit

* **New Features**
  * Added support for Washington’s 2026 Food Assistance Program, including eligibility details, application information, required documents, and navigator resources.
  * Added Washington FAP benefit calculations using SNAP-based rules.

* **Updates**
  * Updated Washington SNAP eligibility requirements to accept only citizenship or qualifying permanent-resident status.
  * Removed outdated warning messages from Trump Account program information across supported states.

### DIFF
--- a/programs/management/commands/import_program_config_data/data/co_trump_account_initial_config.json
+++ b/programs/management/commands/import_program_config_data/data/co_trump_account_initial_config.json
@@ -49,10 +49,5 @@
             "link_url": "https://www.irs.gov/trumpaccounts",
             "link_text": "IRS Trump Accounts page"
         }
-    ],
-    "warning_message": {
-        "external_name": "trump_account_notices",
-        "calculator": "_show",
-        "message": "Trump Accounts are expected to become available on or after July 4, 2026. Check IRS.gov/trumpaccounts for updates."
-    }
+    ]
 }
--- a/programs/management/commands/import_program_config_data/data/il_trump_account_initial_config.json
+++ b/programs/management/commands/import_program_config_data/data/il_trump_account_initial_config.json
@@ -49,10 +49,5 @@
             "link_url": "https://www.irs.gov/trumpaccounts",
             "link_text": "IRS Trump Accounts page"
         }
-    ],
-    "warning_message": {
-        "external_name": "trump_account_notices",
-        "calculator": "_show",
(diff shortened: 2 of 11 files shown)


catonph:
Consolidated into #1695 under MFB-1226 — the MO config addition and the seven-state warning removal are the same piece of work.

PR #1700, opened 2026-08-14
fix: hash passwords saved through the user admin

catonph:
## Problem

Admin users keep reporting that they cannot log in to `api.myfriendben.org/admin/` and ask for a password reset. It is not a forgotten password — the admin is corrupting stored password hashes.

`CustomUserAdmin` extended only unfold's `ModelAdmin`, not `django.contrib.auth.admin.UserAdmin`. `UserAdmin` is what supplies the password-specific form handling: `UserCreationForm` / `UserChangeForm`, the `ReadOnlyPasswordHashField`, and the separate change-password view.

Without it, `password` was rendered as an ordinary editable `CharField` pre-filled with the stored hash. On save, whatever the form submitted was written to `User.password` verbatim and unhashed. `check_password()` then failed, and the admin login reported generic invalid credentials — which reads as a forgotten password rather than a damaged record.

This needs no human error to trigger: browser password managers readily autofill a field named `password`, so simply opening a user, editing an unrelated field, and saving could destroy that user's password. Demonstrated against the pre-fix admin:

```
password field class : CharField
form valid           : True
stored password      : autofilled-plaintext
check_password works : False
```

## Fix

- Mix in `UserAdmin` so the change form uses `ReadOnlyPasswordHashField` and password changes go through `AdminPasswordChangeForm`.
- Declare `fieldsets` / `add_fieldsets` explicitly, since `UserAdmin`'s defaults reference the `username` field that this model replaces with `email_or_cell`.
- `SecureAdmin` is listed first in the MRO so its white-label permission and queryset overrides continue to win. `User` has no `white_label` attribute and `always_can_view` is `False`, so non-superusers resolve `False` on every permission hook — adding `UserAdmin` does not widen access.
- Add `email_or_cell` to `search_fields`. Search previously covered only the nullable `email` field, which is empty for most screener-created users, so admin lookups usually returned nothing — a large part of why these reports were painful to diagnose.
- Narrow `ordering` from `("email_or_cell", "email")` to `("email_or_cell",)`. `email_or_cell` is `unique=True`, so the second key was never reachable as a tiebreaker.
- Drop the `password=password` kwarg from `UserManager.create_user`. `set_password()` immediately overwrites it, so there is no behavior change, but assigning an unhashed value to `User.password` is the same footgun this PR fixes in the admin.

## Tests

Seven tests in `authentication/tests.py`. **Reverting the `admin.py` fix fails 5 of them.**

The two end-to-end tests submit a password value that *differs* from the stored hash, reproducing the autofill scenario above. This matters: an earlier revision of this PR submitted the existing hash back, and because writing the hash over itself is a no-op even on a writable `CharField`, both tests passed with the bug fully reinstated. They proved nothing. Caught in review.

`test_fieldsets_cover_every_editable_field` guards a hazard introduced by this PR itself — explicit `fieldsets` replace Django's auto-generated field list, so any field not listed silently vanishes from the admin, including fields added to the model later. It diffs the model's editable fields against the declared fieldsets. (This caught a real regression in review: `external_id`, the HubSpot contact ID, had been dropped.)

The end-to-end test overrides `STATICFILES_STORAGE`, because rendering the admin templates resolves unfold's static assets and requires a `collectstatic` manifest that is not built in the test environment. It matches the deprecated setting already in `benefits/settings.py:306`; both should move to `STORAGES` together at the Django 5.1 upgrade.

## Validation

- `pytest authentication/ configuration/ screener/tests/test_has_benefits_programs.py screener/tests/test_current_benefits_toggle.py screener/tests/test_referral_sources.py` — **88 passed, 0 failed**
- `manage.py check` — 0 issues
- `black` — clean

Correcting an earlier claim in this description: it previously reported "17 pre-existing failures on `main`." That was wrong. Those failures came from a dirty local working tree during the earlier run, not from `main`; the suites pass clean.

## Follow-up: audit prod for already-corrupted hashes

Run and complete. Two rows matched, both `is_staff = false`, so **no other admin account was locked out** — the damage was limited to the one reported user, since reset.

For the follow-up ticket, the more reliable detector is to match rows whose `password` contains no `$` separator, rather than enumerating known-good prefixes. Enumeration misses legacy `md5$`/`sha1$` hashers and, given the corruption mode, would let through arbitrary plaintext that happens to start with a hasher prefix. `set_unusable_password()` values (`!`-prefixed) must be excluded either way.

### DIFF
--- a/authentication/admin.py
+++ b/authentication/admin.py
@@ -136,6 +136,8 @@ class CustomUserAdmin(SecureAdmin, UserAdmin):
             },
         ),
         ("Contact preferences", {"fields": ("tcpa_consent", "explicit_tcpa_consent", "send_offers", "send_updates")}),
+        # external_id is the HubSpot contact ID, used when debugging CRM sync.
+        ("Integrations", {"fields": ("external_id",)}),
         ("Important dates", {"fields": ("last_login", "date_joined")}),
     )
 
--- a/authentication/admin.py
+++ b/authentication/admin.py
@@ -1,6 +1,6 @@
 from django.contrib import admin
 from django.contrib.auth.models import Group
-from django.contrib.auth.admin import GroupAdmin
+from django.contrib.auth.admin import GroupAdmin, UserAdmin
 from django.core.exceptions import PermissionDenied
 from rest_framework.authtoken.models import TokenProxy
 from rest_framework.authtoken.admin import TokenAdmin
@@ -106,12 +106,40 @@ class SecureAdmin(ModelAdmin):
         return request.user.is_superuser
 
 
-class CustomUserAdmin(SecureAdmin):
-    search_fields = ("email",)
-    ordering = ("email_or_cell", "email")
+# UserAdmin supplies the password-hashing forms (UserCreationForm/UserChangeForm) and
+# renders the stored hash as a read-only field with a link to the change-password view.
+# Without it the password renders as a plain editable CharField and whatever is submitted
+# is saved unhashed, locking the account out.
+class CustomUserAdmin(SecureAdmin, UserAdmin):
+    search_fields = ("email_or_cell", "email", "first_name", "last_name")
+    ordering = ("email_or_cell",)
     filter_horizontal = ["white_labels", "user_permissions"]
 
     list_display = ("email_or_cell", "is_staff")
+
+    # UserAdmin's default fieldsets reference the "username" field, which this model
+    # replaces with email_or_cell, so they are declared explicitly.
+    fieldsets = (
+        (None, {"fields": ("email_or_cell", "password")}),
+        ("Personal info", {"fields": ("first_name", "last_name", "email", "cell", "language_code")}),
+        (
+            "Permissions",
+            {
+                "fields": (
+                    "is_active",
+                    "is_staff",
+                    "is_superuser",
+                    "white_labels",
+                    "groups",
+                    "user_permissions",
+                )
+            },
+        ),
+        ("Contact preferences", {"fields": ("tcpa_consent", "explicit_tcpa_consent", "send_offers", "send_updates")}),
+        ("Important dates", {"fields": ("last_login", "date_joined")}),
+    )
+
+    add_fieldsets = ((None, {"classes": ("wide",), "fields": ("email_or_cell", "password1", "password2")}),)
 
 
 class CustomGroupAdmin(SecureAdmin, GroupAdmin):
--- a/authentication/models.py
+++ b/authentication/models.py
@@ -13,7 +13,7 @@ class UserManager(BaseUserManager):
         if not email_or_cell:
             raise ValueError("Users must have an email address or cell phone number")
 
-        user = self.model(email_or_cell=email_or_cell, password=password, explicit_tcpa_consent=False)
+        user = self.model(email_or_cell=email_or_cell, explicit_tcpa_consent=False)
 
         user.set_password(password)
         user.save(using=self._db)
--- a/authentication/tests.py
+++ b/authentication/tests.py
@@ -1,3 +1,137 @@
-# from django.test import TestCase
+from django.contrib.admin.sites import AdminSite
+from django.contrib.auth.forms import AdminPasswordChangeForm
+from django.test import TestCase, override_settings
+from django.urls import reverse
 
-# Create your tests here.
+from .admin import CustomUserAdmin
+from .models import User
+
+
+class UserAdminPasswordTests(TestCase):
+    """
+    The admin must never write an unhashed value to User.password.
+
+    CustomUserAdmin previously extended only ModelAdmin, so `password` rendered as an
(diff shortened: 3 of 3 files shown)


catonph:
Thanks — the `external_id` regression was real and accidental, not deliberate. Fixed in 76e6548 along with the three minor items.

## 🔴 `external_id` regression — fixed

Confirmed independently before fixing; it was the only editable field omitted:

```
MISSING from fieldsets: ['external_id']
```

Restored under its own `Integrations` section rather than `Personal info`, since it is a CRM identifier rather than user-entered data.

I also added a guard test, because the underlying hazard is that explicit `fieldsets` silently drops any field not listed — including fields added to the model later. `test_fieldsets_cover_every_editable_field` diffs the model's editable fields against the declared fieldsets. Verified it catches the exact regression you found: removing the `Integrations` entry fails with

```
'external_id' : editable User fields missing from CustomUserAdmin.fieldsets
```

## 🟡 Minor items

1. **`RequestFactory`** — done. You were right that it was issuing a real `GET /` per call; the 404 log noise is gone.
2. **`ordering`** — leaving as `("email_or_cell",)`, agreed it is harmless given `unique=True`. Now noted in the PR body so it does not read as unexplained.
3. **Strengthened e2e assertion** — done, now asserts `first_name == "Updated"` so a form-error re-render cannot pass as success.
4. **`STATICFILES_STORAGE` deprecation** — leaving as-is to match `benefits/settings.py:306`; it should move to `STORAGES` as one change at the Django 5.1 upgrade rather than diverging here.

## `create_user` footgun

Folded in — dropped the `password=password` kwarg from `self.model(...)`. `set_password()` overwrote it, so no behavior change, but it is the same "assign an unhashed value to `User.password`" shape as the bug this branch fixes. The suites that call `create_user` still pass.

## Validation

- `pytest authentication/` — **7 passed**
- `manage.py check` — 0 issues
- `black` — clean
- Surrounding suites: **17 pre-existing failures on `main`, 17 after** — unchanged, no new ones.

## Blast radius

Agreed on making it a real follow-up ticket rather than a maybe, and thank you for the `!%` exclusion — `set_unusable_password()` values would otherwise have shown up as false positives. Using your query, adjusted for the remaining hashers in `PASSWORD_HASHERS`:

```sql
SELECT id, email_or_cell, is_staff FROM authentication_user
WHERE password NOT LIKE 'pbkdf2\_%' AND password NOT LIKE 'argon2%'
  AND password NOT LIKE 'bcrypt%'   AND password NOT LIKE 'scrypt%'
  AND password NOT LIKE '!%'        AND password <> '';
```

I have not run this against prod — flagging for @catonhauer to authorize, since it is the piece that tells us whether other admins are already locked out and simply have not reported it.

catonph:
You were right, and it was worse than reported: on my machine reverting `admin.py` failed **0** tests, not 2. Fixed in f8bd065d.

## The two tests proved nothing

Confirmed your diagnosis exactly. Instrumenting the pre-fix admin:

```
password field class : CharField
form valid           : True
stored password      : autofilled-plaintext
check_password works : False
```

That is the production failure, and submitting `self.user.password` back meant the corrupting write was a no-op, so `check_password()` succeeded either way. Both tests were structurally incapable of failing.

Both now submit `"autofilled-by-the-browser"`. Reverting the fix fails **5 tests**, including both end-to-end guards:

```
FAILED test_add_form_hashes_password
FAILED test_fieldsets_cover_every_editable_field
FAILED test_password_field_is_not_editable_as_plain_text
FAILED test_saving_user_via_admin_form_preserves_password   ← was passing
FAILED test_change_view_post_preserves_password             ← was passing
```

## Why my "3 fail" number was wrong

Worth recording, since you flagged the number as load-bearing. My original check stashed `admin.py` — but `git stash push` on a single file interacted badly with a pre-existing unrelated stash entry, and the pop merged foreign changes into `screener/models.py`. Some runs executed against a conflicted tree. I restored the tree from `HEAD` and confirmed the unrelated WIP stash is intact.

The same dirty tree is where the **"17 pre-existing failures on `main`"** claim came from. That was also wrong — those suites pass clean, **88 passed, 0 failed**. Both numbers are corrected in the description.

## Search test

Replaced the `search_fields` literal assertion with a `get_search_results()` call, so it verifies search behavior rather than config.

## Not changed

- **`create_superuser` setting `is_admin`** — confirmed it is not a field on the model; pre-existing dead assignment, agreed out of scope.
- **`test_password_field_is_not_editable_as_plain_text`** — kept. Indirect as you note, but it now backs up two genuine end-to-end guards rather than carrying the suite alone.

## Follow-up SQL

Adopting the inverted check — matching rows with no `$` separator. Enumerating prefixes misses `md5$`/`sha1$` and would let through plaintext that happens to begin with a hasher prefix, which is a live concern given the corruption mode. Noted in the description for the ticket.

The audit itself came back clean: 2 rows, both `is_staff = false`, no other admin locked out.

PR #1708, opened 2026-08-19
MFB-1567: Retire the legacy validation system

catonph:
Closes [MFB-1567](https://linear.app/myfriendben/issue/MFB-1567/test-machinery-retire-old-validation-dbharnessworkflow).

Removes the importable-validation database, its harness, management commands, REST API, and the CI steps that invoked them. Spec-driven tests in `programs/programs/{state}/{program}/tests/` are the coverage mechanism now.

**Paired frontend PR:** MyFriendBen/benefits-calculator#2184 — merge this one first, since it removes `validations` from the results payload.

## What was actually still running

The ticket described this as dormant. It wasn't, quite:

- **Staging** ran `manage.py validate` on every deploy, writing to a Google Sheet. Wrapped in `continue-on-error: true`, so it never blocked anything and nobody read the output.
- **Production**'s step was stubbed to `run: exit 1` per a MFB-470/472 timeout TODO — but the final-status gate still required `steps.validations.outcome == "success"`. A step hardcoded to fail can never be `success`, so **every production deploy has been reporting `completed-with-warnings` in Slack.** Fixed here as a side effect.
- The **`api/validations/` REST endpoint** was live and used by an admin-only frontend workflow (create/remove expected values from the results page). Not mentioned in the ticket's "Current surface" section.

## Removed

**App** — model, viewset, serializers, permissions, admin, `api/validations/` route, 4 management commands (`validate`, `import_validations`, `pull_validations`, `update_validations`), 67 data JSONs, schemas, READMEs.

**Results contract** — `validations` array out of the payload ([`screener/views.py`](screener/views.py), [`screener/serializers.py`](screener/serializers.py)) and `validations__isnull=True` out of `pull_screens`.

**`Screen.frozen`** — the property, the `ScreenSerializer.update` write guard it gated, and the serializer field. It was settable *only* by the presence of validation rows, so once the table is gone it is permanently `False` and every reader is unreachable.

**`get_reference_date()`** — had one non-default branch, fed by the earliest validation's `created_date`. Every non-frozen screen already took the `timezone.now().date()` path, so removal is behavior-preserving except for screens that had validation rows. Inlined at its six call sites (`screener/models.py`, `ma/bsp`, `mo/ssdi`, `ks/ssdi`, `wa/ssdi`, `federal/trump_account`), and `age_from_date()` drops the `reference_date` parameter that existed to receive the injected value.

**CI** — staging step, both composite actions (`heroku-validations`, `heroku-pull-validations`), prod's stub step and its gate term, and slack-notify's validation inputs + spreadsheet link block.

## Test-date pinning

The two suites that patched `get_reference_date` (and three that mocked it) used it as an injection seam to keep fixed birth months from drifting. With the method gone that seam disappears, so the five age-sensitive suites now pin time with `freezegun` (`freezegun==1.5.5`, new test dependency):

- `ma/bsp` — class-level `@freeze_time(FROZEN_DATE)`, replacing `patch.object`
- `mo/ssdi`, `ks/ssdi`, `wa/ssdi` — `@freeze_time(REFERENCE_DATE)` on the eligibility/value classes
- `federal/trump_account` — per-call `freeze_time` in the pregnancy helper, which *varies* the date across six pilot-window boundary cases rather than pinning it

## Migration sequencing — important for deploy

`validations/migrations/0005_delete_validation` emits `DROP TABLE "validations_validation" CASCADE`. The app package, `models.py`, and `INSTALLED_APPS` entry are **deliberately retained** so that migration actually applies. Dropping the app in the same commit would leave the table orphaned in every environment with dangling `django_migrations` rows.

**Follow-up required:** once 0005 has been deployed everywhere, delete the `validations/` package and its `INSTALLED_APPS` entry. There's a comment at the registration site noting this.

## Verification

- `3309 passed` (full suite, fresh DB)
- `manage.py check` — no issues; `makemigrations --check` — no changes detected
- `sqlmigrate validations 0005` confirms the `DROP TABLE`
- `black --check` clean; all three deploy YAML files parse, and the edited `jq` payload was executed to confirm valid JSON

## Deliberately left alone

Five spec.md files still contain validation-*workflow prose* rather than dead file links — `ks/chip`, `wa/nslp`, `wa/lifeline`, `wa/orca_lift`, `wa/senior_disabled_pte`. These are historical discovery records, and some carry substantive findings (`wa/nslp` records two real coverage gaps against `test_case_schema.json`; `ks/chip` cites it as the authority for KS bare-county naming). Rewriting them means editorial judgment on someone else's research notes, and nothing imports a spec.md, so the staleness is inert. Worth a small follow-up by whoever owns that research.

Two references were already stale before this PR: `ks/snap/spec.md` and `wa/ctc/spec.md` pointed at `ks_snap.json` / `wa_ctc.json`, which never existed in the data directory.

## AC #4 — verified against production

The ticket asks whether any program depended on a validation JSON for coverage not yet in a spec. Two things were checked:

**Are the frozen screens real user data?** No. Production has 1039 validation rows across **537 distinct screens, and all 537 are flagged `is_test_data`** (418 via the import command's explicit `is_test=True`, the remaining 119 via a test-ish `referral_source`/`referrer_code`, which `Screen.set_screen_is_test()` folds into the same flag). `import_validations` was the only writer and always set `is_test=True`, so the frozen set is test fixtures by construction. Most recent write was 2026-06-11.

This matters because dropping the table has two consequences for frozen screens — they become editable (the `ScreenSerializer.update` guard stops firing) and their ages recompute against today instead of the earliest validation timestamp. Both are inert on test fixtures, which is what all 537 are.

**Is spec coverage in place?** Per the ticket author, the in-flight spec backfill covers this, so the JSONs are deleted rather than ported. Git history retains all 67 files if a backfill author wants prior expected values. For the record: 23 of the 63 import JSONs had no matching `spec.md` at deletion time (concentrated in MA/TX/IL) and 6 more matched only a *different* state's spec.

## Reviewer note

CodeRabbit skipped this PR automatically — 124 files exceeds its 100-file limit — so there's no automated pass on the backend. The bulk is deletions (67 JSONs + the app), but the hand-written parts worth a human look are:

- `screener/models.py` — the six inlined `get_reference_date()` call sites and `age_from_date()`'s signature change
- `.github/workflows/deploy-production.yml` — the gate-term removal
- `.github/actions/slack-notify/action.yml` — the `jq` payload edit (executed locally to confirm valid JSON, but it's a shell-embedded heredoc, so worth eyeballing)


david-j-stoll:
## Leftover references to the retired validation system

Swept `caton/mfb-1567-retire-validations` in both repos. benefits-calculator is clean — every remaining hit there is form validation or `useThemeValidation`. In benefits-api, **18 spec.md files** still point at deleted validation JSON, the deleted runner, or the deleted `test_case_schema.json`.

### spec.md files still referencing validations

| `programs/programs/…/spec.md` | lines | dead thing it points at |
| --- | --- | --- |
| `tx/ccad` | 63, 67, 109, 195 | `tx_ccad.json`, `[validation]` scenario tags |
| `ks/tanf` | 215, 236, 242, 379, 449 | `ks_tanf.json` |
| `ks/k40h` | 226, 241, 256 | "✓ in validation JSON" |
| `ks/chip` | 21 | "**Validations must use 255%** as the eligible/ineligible boundary" |
| `wa/nslp` | 37, 41, 57, 92, 302, 339, 341 | `test_case_schema.json`, `wa_nslp.json` |
| `wa/apple_health` | 173, 200, 202, 226, 276 | `wa_apple_health.json` |
| `wa/apple_health_for_kids` | 93, 103 | "Validation JSON convention", "update the validation JSON" |
| `wa/orca_lift` | 160, 330, 332 | `wa_orca_lift_test_cases.json` |
| `wa/cdctc` | 18, 96 | `wa_cdctc.json` validations draft |
| `wa/wsos_bas` | 158, 163, 175 | "the validation uses…", "if the validation system supports…" |
| `wa/wsos_grd` | 179 | "Matches the JSON validation edge case" |
| `wa/wap` | 117 | "Validation scenarios assert `value: 7669`" |
| `wa/senior_disabled_pte` | 101 | "For validation scenarios…" |
| `wa/head_start` | 373 | "exclude when picking validation cases" |
| `wa/lifeline` | 139 | `## Validation Scenarios` heading |
| `wa/ssi` | 481 | "not represented in the JSON validation suite" |
| `il/silver_access` | 72 | "…the unit the calculator returns and the validation suite asserts" |
| `federal/trump_account` | 21 | "cannot be tested via the validation runner" |

Most of these are one-clause rewordings rather than section deletions. Two carry load-bearing facts that should survive the edit — `ks/chip:21` (the 255% FPL boundary) and `il/silver_access:72` (annual-vs-monthly units) — so those just need "Validations"/"the validation suite" → "tests". Only three are genuinely structural: `wa/nslp:339-341` ("Representative Validation Scenarios Used in `wa_nslp.json`"), `wa/orca_lift:330-332` ("Extracted for JSON validation"), and the `wa/lifeline:139` heading.

For reference, these mentions of "validation" are **generic** and should be left alone: `mo/pts/spec.md:70`, `ks/snap/spec.md:420,439,458`, `wa/snap/spec.md:411`, `wa/tanf/spec.md:326,328`, `wa/wap/spec.md:309`, and the `### Validation` sections in `import_program_config_data/README.md` / `import_urgent_need_config_data/README.md`.

### Stale code comments

Five comments in shipped code still describe the retired system as live:

- `programs/programs/wa/wsos_bas/calculator.py:80` — "Validations only cover sizes 1 and 4 today; this fallback exists so…"
- `programs/programs/wa/wsos_grd/calculator.py:65` — "Validations only cover sizes 1 and 3 today; this fallback exists…"
- `programs/programs/wa/wsos_bas/tests.py:21` — "persisted validation suite (`wa_wsos_bas.json`) covers Scenarios 1, 2, …"
- `programs/programs/wa/nslp/tests.py:37` — "Spec / validation: HH 3, one school-age child…"
- `programs/framework/pe_dependencies/tests/test_spm.py:248` — "A heating-only expense (the validation scenario shape) flows through to the cap."

The two calculator ones matter most: they justify a live fallback branch by citing coverage in a file that no longer exists, so the next reader can't verify the rationale.

### `validations` is still registered in `INSTALLED_APPS`

`benefits/settings.py:130` still has `"validations.apps.ValidationsConfig"`, and `validations/` still ships `apps.py`, an empty `models.py`, and migrations `0001`–`0005`.

That looks correct **for this release** — the app has to stay installed for `0005_delete_validation` to run — but it does mean the package can't be deleted in this PR, and an empty `models.py` under an installed app is easy to mistake for an oversight later. Could we confirm that's the intent, and get a follow-up on the board to drop `validations/` (app, `apps.py`, `models.py`, and the migration chain) once `0005` has run in every environment? Without it the retirement stops half-done.

PR #1727, opened 2026-08-26
MFB-1672: restrict the state dropdown per referrer and brand the UWGKC 2-1-1

david-j-stoll:
## Context & Motivation

- Fixes [MFB-1672](https://linear.app/myfriendben/issue/MFB-1672/handle-moks-kansas-city-overlap-in-landing-page)
- Related PR: MyFriendBen/benefits-calculator#2187 (frontend half — renders what this config declares)

United Way of Greater Kansas City's 2-1-1 serves the Kansas City metro, which spans the state line,
so its referral links need to land on a state dropdown offering exactly Kansas and Missouri.
Neither state is publicly launched, so this can't be a filter over the launched set — it needs a
per-referrer override.

The dropdown's contents were entirely frontend-side: a hardcoded `code -> name` map plus a
hardcoded public-state list with `ks` and `mo` commented out. Both went stale whenever a white
label was added or launched. So this PR also moves that knowledge into the white label configs,
where the states are already declared.

The ticket's second acceptance criterion — 2-1-1 branding for KS and MO — is covered here too, as
the theme, logo and share-link config the frontend PR renders.

### State dropdown

- `base.py`: two class attributes on `ConfigurationData` — `is_state` (whether the dropdown can
  offer this white label at all) and `publicly_launched` (whether it shows without a referrer).
  `publicly_launched` defaults to `False`, so launching is opt-in: a state under construction
  can't reach the public dropdown by omission.
- `co.py` / `il.py` / `ma.py` / `nc.py` / `tx.py` / `wa.py`: `publicly_launched = True`, opting the
  six live states in. The derived catalog is byte-identical before and after that flip.
- `cesn.py`: `is_state = False`, since it's a Colorado sub-brand rather than a state screener and
  should never be selectable as a state.
- `ks.py` / `mo.py`: `publicly_launched = False` — reachable at `/ks` and `/mo` and offered to the
  2-1-1, but not in the public dropdown. Now the same as the default, kept explicit because these
  two are the point of the PR.
- `white_labels/__init__.py`: `state_options()` derives the dropdown catalog from the registry —
  `[{"code", "name", "public"}]`, name taken from each state's existing `state["name"]`, sorted by
  display name.
- `add_config.py`: saves that catalog as a new `state_options` configuration row. The write sits
  above the `is_default` guard, so **every** white label gets an identical copy, and the frontend
  reads the row for whichever label is loaded — which is why `--all` is required.

### The `uwgkc` referrer

Stakeholders confirmed one referrer code for both states, reached as
`/ks/step-1?referrer=uwgkc` and `/mo/step-1?referrer=uwgkc`. Each state path loads its own config,
so the referrer is declared in both — sharing the theme and `stateOptions`, while `logoSource`,
`logoAlt`, `shareLink` and `stateName` differ per state:

- `ks.py` / `mo.py` `referrer_data`: a `uwgkc` entry under `theme` (`"uwgkc"`), `logoSource`
  (`"KS211_MFBLogo"` / `"MO211_MFBLogo"` — UWGKC supplied a lockup per state, same 2-1-1 signature
  differing only in the state label), `logoAlt`, `shareLink` (state-specific, carrying the referrer
  through a share), `stateOptions` (`["ks", "mo"]`), `uiOptions` (`["white_header"]`), `logoClass`
  (`"uwgkc-logo-size"`, since the lockup is far wider than the default logo slot allows for), and a
  blank `stateName` — the lockup already reads "Kansas" / "Missouri", so the header's separate
  state name rendered the word a second time underneath the artwork.
- `uiOptions: ["white_header"]` is load-bearing, not cosmetic: the supplied lockups carry
  MyFriendBen's `#293457` navy wordmark, which sits at 1.08:1 against the theme's `#21296B` header
  and is effectively invisible. `white_header` is the existing option `cesn` already uses — it
  whitens the bar and flips the header text and icons to black, where the wordmark is 12.2:1.
- `_default.py`: stays the generic `{"default": []}` — every publicly launched state. It backs the
  state-agnostic `/select-state` route, so it has no business naming one partner's states, and the
  `uwgkc` entry lives only where the screener reads it. The tradeoff is that a bare
  `/?referrer=uwgkc` link can't narrow the dropdown; UWGKC's links are state-scoped, so nothing in
  the agreed flow depended on that.

### Tests and docs

- `configuration/tests.py`: `TestStateOptionsConfiguration` covers five things — the catalog
  matches a pinned `{code: public}` map, non-state white labels are excluded from it, `_default`
  declares exactly the generic `{"default": []}` and nothing partner-specific, every code a
  referrer override names exists in the catalog, and a referrer offering two or more states is
  configured with `theme`, `logoSource` and `logoClass` in *each* of them and offers the same set
  of states from each. That last one is the failure mode this design invites: configure `uwgkc` in
  `ks.py` only and the Missouri path silently serves generic branding with an unscoped dropdown.

  The catalog assertion is pinned rather than recomputed from `is_state`/`publicly_launched`, which
  would only restate the predicate `state_options()` filters on. Adding a state now needs a
  deliberate edit to `EXPECTED_CATALOG` — the right place to decide whether it belongs in the
  public dropdown yet.

  On the `stateOptions` half of that check, CodeRabbit correctly caught that the first version only
  asserted the referrer *key* was present, so `ks.py` could offer `["ks"]` while `mo.py` offered
  `["ks", "mo"]` and the guard stayed green — entering from `/ks` would then hide Missouri, the
  exact misconfiguration the test exists for. It now compares membership (order left free, since a
  state may reasonably list itself first), and gates on a referrer naming two or more states rather
  than on the declaring state appearing in its own list — which also covers a referrer pointing at
  states other than its own, previously skipped entirely.
- `_template.py`: points at `styleController.ts` for theme names instead of duplicating the list,
  corrected the `stateOptions` note (it claimed only `_default` is read), and defaults
  `publicly_launched` to `False` with a TODO to flip it at launch.
- `white_labels/README.md`: rewrote the `stateOptions` section for the same reason, and corrected
  where the catalog lands — an earlier revision of this PR said it was written to the `_default`
  row and offered `add_config _default` as a shortcut. Both were wrong: the write precedes
  `add_config`'s `is_default` guard, so every white label gets a copy and the frontend reads its
  own, meaning that shortcut would have left every state row stale.

No migrations and no model changes — a derived config row, two class flags, and referrer config.

## Testing

- Migrations to run: none
- Configuration updates needed: `python manage.py add_config --all` (the catalog row is written for
  every white label)
- Manual testing steps:
  1. `python manage.py add_config --all`
  2. `GET /api/configuration/_default/` and confirm `state_options` lists all eight states with
     `ks` and `mo` at `"public": false`, and that `referrer_data.stateOptions` is just
     `{"default": []}`
  3. `GET /api/configuration/ks/` and `/mo/` and confirm each carries a `uwgkc` entry under
     `referrer_data.theme`, `logoSource`, `logoAlt`, `shareLink`, `stateOptions` and `uiOptions`,
     with the state's own logo key
  4. With the frontend branch running, `/ks/step-1?referrer=uwgkc` and
     `/mo/step-1?referrer=uwgkc` render United Way branding; `/` with no referrer offers the same
     six states as production
  5. `python manage.py test configuration` (53 passing locally)

## Deployment

- Post-deployment scripts needed: `python manage.py add_config --all` on staging, then production.
  `--all` is required, not a convenience: `state_options` is written per white label, so a
  single-label run leaves every other row stale, and until it runs at all the row doesn't exist and
  the frontend renders an empty dropdown — see Notes for Reviewers. Nine configs changed here
  (`_default`, the six launched states, `ks` and `mo`).
- Production config updates: the same `add_config --all` run.
- Admin updates needed: **yes.** `add_config` does not write `Referrer` rows, and they're
  `unique(white_label, referrer_code)` — so `uwgkc` needs a row created under **both** `ks` and
  `mo` for referral-source attribution. The dropdown and branding work off the URL param without
  them; attribution does not.
- Notify team/users of: UWGKC needs the final links —
  `screener.myfriendben.org/ks/step-1?referrer=uwgkc` and `.../mo/step-1?referrer=uwgkc`.
- **This unblocks a red CI job on the frontend PR.** MyFriendBen/benefits-calculator#2187 has 7
  failing `e2e-tests` specs, all of them timing out on an empty state dropdown, because that suite
  runs against the deployed API and `state_options` does not exist there yet. After this merges and
  `add_config --all` has run, re-run that job and confirm all 7 pass with no code change — that is
  the staging QA gate for the ordering tradeoff below.

## Notes for Reviewers

- **The referrer code is `uwgkc`**, one code across both states, per stakeholders. An earlier
  revision of this PR used separate `211ks` / `211mo` codes declared in `_default.py`; the switch is
  why the config now lives in `ks.py` and `mo.py`, with `_default.py` left generic.
- **Deploy ordering matters now**, unlike the first version of this PR. The frontend no longer has
  a hardcoded fallback list, so between the frontend deploy and `add_config` the dropdown would be
  empty. Running `add_config --all` first, or in the same release, avoids it. The frontend PR's red
  `e2e-tests` job is this tradeoff made visible — it runs against the deployed API, so it cannot go
  green until this side ships. Worth a second opinion on whether that tradeoff is right, or whether
  the frontend should keep a minimal fallback, which would also decouple its e2e suite from API
  deploy order.
- Known limitations:
  - A `uwgkc` user who clicks "click here for other state options" on the zipcode step loses the
    referrer: that link is a plain `<a href="/select-state">` with no query string, so the reload
    falls back to the public list — which contains neither KS nor MO. Deliberately out of scope
    here; the cross-state link needs the referrer propagated through the state page.
  - State names are plain strings, not translation keys, matching the hardcoded English names they
    replace. Making them translatable is a follow-up.
  - `publicly_launched` is a launch decision, so it can't be derived — but it now lives beside the
    state it describes rather than in frontend code.
- Future considerations:
  - The "single screener covering both states" framing in the original ticket description isn't
    addressed here: `state_code` is a per-white-label class constant on every PE calculator, so a
    combined white label would need it derived per household from the ZIP. Two screeners plus one
    shared referrer gets the KC-metro user to the right state screener instead.
  - Cross-state program visibility isn't needed — `ks` and `mo` already duplicate the KC-metro
    resources, and the four programs `ks` was missing (ACA PTC, federal EITC, WIC, WAP) are tracked
    in MFB-1052, MFB-1048, MFB-1051 and MFB-1067. See the comment on MFB-1672.

## Summary by CodeRabbit

- **New Features**
  - State dropdown options are now generated from configured, publicly launched states.
  - Kansas and Missouri remain accessible by direct links and can be offered through eligible referrer links without appearing in the public dropdown.
  - Added United Way of Greater Kansas City branding and cross-state options for Kansas and Missouri.

- **Bug Fixes**
  - Referrer-specific state selections now apply consistently across state entry points.

- **Documentation**
  - Clarified state visibility, referrer configuration, theming, and rollout guidance.

### DIFF
--- a/configuration/tests.py
+++ b/configuration/tests.py
@@ -189,3 +189,36 @@ class TestLegalLinkConfiguration(SimpleTestCase):
                             f"absolute https URL: {link!r}. An empty string here renders as a link "
                             "with an empty href.",
                         )
+
+
+class TestStateOptionsConfiguration(SimpleTestCase):
+    """
+    Guards the referrer-scoped state dropdown config. The dropdown renders before a state is
+    chosen, so it only ever reads the "_default" white label, and a code that is not a real
+    white label silently falls back to the public state list — a typo here would look like the
+    feature was never configured.
+    """
+
+    def test_default_white_label_declares_state_options(self):
+        """The "_default" config is the only one the state dropdown reads."""
+        self.assertIn("stateOptions", white_label_config["_default"].referrer_data)
+
+    def test_state_options_codes_are_real_white_labels(self):
+        """Every state code offered to a referrer must be a white label the screener can route to."""
+        for code, white_label_data in white_label_config.items():
+            state_options = white_label_data.referrer_data.get("stateOptions", {})
+
+            for referrer, states in state_options.items():
+                for state in states:
+                    with self.subTest(white_label=code, referrer=referrer, state=state):
+                        self.assertIn(
+                            state,
+                            white_label_config,
+                            f'"{referrer}" in white label "{code}" offers state "{state}", which '
+                            "is not a white label. The dropdown would drop it.",
+                        )
+                        self.assertNotEqual(
+                            state,
+                            "_default",
+                            f'"{referrer}" in white label "{code}" offers "_default", which is not a state.',
+                        )
--- a/configuration/tests.py
+++ b/configuration/tests.py
@@ -7,7 +7,7 @@ from urllib.parse import urlparse
 from django.test import SimpleTestCase, TestCase
 from configuration.models import Configuration
 from configuration.serializers import ConfigurationSerializer
-from configuration.white_labels import white_label_config
+from configuration.white_labels import state_options, white_label_config
 from screener.models import WhiteLabel
 from screener.feature_flags import FeatureFlagConfig
 
@@ -192,33 +192,38 @@ class TestLegalLinkConfiguration(SimpleTestCase):
 
 
 class TestStateOptionsConfiguration(SimpleTestCase):
-    """
-    Guards the referrer-scoped state dropdown config. The dropdown renders before a state is
-    chosen, so it only ever reads the "_default" white label, and a code that is not a real
-    white label silently falls back to the public state list — a typo here would look like the
-    feature was never configured.
-    """
+    """Guards the state dropdown's derived catalog and the per-referrer overrides that select from it."""
+
+    def test_catalog_covers_every_state_white_label(self):
+        """Every state white label appears once, with a name to show in the dropdown."""
+        catalog = state_options()
+        codes = [state["code"] for state in catalog]
+
+        self.assertEqual(len(codes), len(set(codes)))
+        for code, white_label_data in white_label_config.items():
+            expected = white_label_data.is_state and not white_label_data.is_default
+            with self.subTest(white_label=code):
+                self.assertEqual(code in codes, expected)
+
+        for state in catalog:
+            with self.subTest(white_label=state["code"]):
+                self.assertTrue(state["name"], f'White label "{state["code"]}" has no state name to display.')
 
     def test_default_white_label_declares_state_options(self):
-        """The "_default" config is the only one the state dropdown reads."""
+        """The "_default" config is the only one the dropdown reads its referrer overrides from."""
         self.assertIn("stateOptions", white_label_config["_default"].referrer_data)
 
-    def test_state_options_codes_are_real_white_labels(self):
-        """Every state code offered to a referrer must be a white label the screener can route to."""
-        for code, white_label_data in white_label_config.items():
-            state_options = white_label_data.referrer_data.get("stateOptions", {})
+    def test_referrer_overrides_name_states_in_the_catalog(self):
+        """An override naming a state outside the catalog would render an empty dropdown."""
+        codes = {state["code"] for state in state_options()}
 
-            for referrer, states in state_options.items():
+        for code, white_label_data in white_label_config.items():
+            for referrer, states in white_label_data.referrer_data.get("stateOptions", {}).items():
                 for state in states:
                     with self.subTest(white_label=code, referrer=referrer, state=state):
                         self.assertIn(
                             state,
-                            white_label_config,
-                            f'"{referrer}" in white label "{code}" offers state "{state}", which '
-                            "is not a white label. The dropdown would drop it.",
-                        )
-                        self.assertNotEqual(
-                            state,
-                            "_default",
-                            f'"{referrer}" in white label "{code}" offers "_default", which is not a state.',
+                            codes,
+                            f'"{referrer}" in white label "{code}" offers "{state}", which the state '
+                            "dropdown has no entry for.",
                         )
--- a/configuration/tests.py
+++ b/configuration/tests.py
@@ -194,20 +194,53 @@ class TestLegalLinkConfiguration(SimpleTestCase):
 class TestStateOptionsConfiguration(SimpleTestCase):
     """Guards the state dropdown's derived catalog and the per-referrer overrides that select from it."""
 
-    def test_catalog_covers_every_state_white_label(self):
-        """Every state white label appears once, with a name to show in the dropdown."""
+    # Pinned rather than re-derived from is_state/publicly_launched: computing the expectation with
+    # the same predicate state_options() filters on would only restate the implementation. A new
+    # state has to be added here deliberately, which is the point — that edit is where someone
+    # decides whether it belongs in the public dropdown yet.
+    EXPECTED_CATALOG = {
+        "co": True,
+        "il": True,
+        "ks": False,
+        "ma": True,
+        "mo": False,
+        "nc": True,
+        "tx": True,
+        "wa": True,
+    }
+
+    def test_catalog_matches_the_expected_states(self):
+        """The dropdown offers exactly these states, with exactly these launched flags."""
         catalog = state_options()
         codes = [state["code"] for state in catalog]
 
-        self.assertEqual(len(codes), len(set(codes)))
-        for code, white_label_data in white_label_config.items():
-            expected = white_label_data.is_state and not white_label_data.is_default
-            with self.subTest(white_label=code):
-                self.assertEqual(code in codes, expected)
+        self.assertEqual(len(codes), len(set(codes)), f"The catalog repeats a state: {codes}.")
+        self.assertEqual(
+            {state["code"]: state["public"] for state in catalog},
+            self.EXPECTED_CATALOG,
+            "The state dropdown's contents changed. If that was intended, update EXPECTED_CATALOG; "
+            'a state added with "publicly_launched" unset is offered only by link and to referrers '
+            "that name it.",
+        )
 
         for state in catalog:
             with self.subTest(white_label=state["code"]):
                 self.assertTrue(state["name"], f'White label "{state["code"]}" has no state name to display.')
+
+    def test_catalog_excludes_non_state_white_labels(self):
+        """A sub-brand like "cesn" routes to a state screener but is not itself a state choice."""
+        codes = {state["code"] for state in state_options()}
+
+        for code, white_label_data in white_label_config.items():
+            if white_label_data.is_state and not white_label_data.is_default:
+                continue
+
+            with self.subTest(white_label=code):
+                self.assertNotIn(
+                    code,
+                    codes,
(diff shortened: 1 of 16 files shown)


configuration/white_labels/_template.py
```diff
@@ -266,7 +269,8 @@ def get_white_label(self) -> WhiteLabel:
     #
     # theme: CSS theme name (usually "default")
     #   - Used to apply custom styling/themes
-    #   - Available themes: "default", "twoOneOne", "twoOneOneNC", "nc_lanc"
+    #   - Available themes: "default", "twoOneOne", "twoOneOneNC", "co_energy", "nc_lanc",
```
catonph: shoudl we reference where these live instead of having to keep the had coded list up to date

configuration/white_labels/base.py
```diff
@@ -15,6 +15,12 @@
 class ConfigurationData:
     is_default = False
 
+    # Whether this white label is a state screener the state dropdown can offer.
+    is_state = True
+
+    # Whether this state appears in the public state dropdown, as opposed to being reachable only by link.
+    publicly_launched = True
```
catonph: We should probably default to False?

configuration/white_labels/README.md
```diff
@@ -235,6 +235,25 @@ referrer_data = {
...
+
+Keep partner entries out of `_default.py`. It stays the generic every-public-state list, since it
+backs the state-agnostic `/select-state` route rather than any one partner. The tradeoff is that a
+bare `/?referrer=code` link cannot narrow the dropdown — a referrer that needs one should be given
+state-scoped links instead.
+
+The dropdown's catalog of states is not hand-maintained — `add_config` derives the `state_options`
+config from the white label registry, using each state's `state["name"]`, `is_state` and
+`publicly_launched`. Because the catalog is written to the `_default` row, a new state appears
+only once its config exists and `add_config --all` (or `add_config _default`) runs.
```
catonph: This says the catalog "is written to the `_default` row" and offers `add_config _default` as a valid shortcut, but neither is accurate. The `state_options` write at `add_config.py:95` sits **above** the `if WhiteLabelData.is_default: continue` guard at line 107, so it writes a `state_options` row for *every* white label — and the frontend reads the loaded label's own row (`GET /api/configuration/ks/`), not `_default`'s.

Anyone following the documented `add_config _default` shortcut would leave every state row empty and render an empty dropdown — precisely the outage the PR description warns about. Suggest rewording to say the catalog is written to every white label's row, and that `--all` is required.

configuration/tests.py
```diff
@@ -191,6 +191,103 @@ def test_legal_links_are_populated_urls(self):
...
+    def test_catalog_covers_every_state_white_label(self):
+        """Every state white label appears once, with a name to show in the dropdown."""
+        catalog = state_options()
+        codes = [state["code"] for state in catalog]
+
+        self.assertEqual(len(codes), len(set(codes)))
+        for code, white_label_data in white_label_config.items():
+            expected = white_label_data.is_state and not white_label_data.is_default
+            with self.subTest(white_label=code):
+                self.assertEqual(code in codes, expected)
```
catonph: This assertion is a tautology: `expected` on line 204 is computed from `white_label_data.is_state and not white_label_data.is_default`, which is the same predicate `state_options()` filters on. The test therefore can't fail for any implementation that uses that predicate — inverting or deleting the filter inside `state_options()` keeps this green.

Pinning a literal expected set (e.g. `{"co", "nc", "ks", "mo", ...}`) would make it actually load-bearing, and would fail loudly when a state is added without a deliberate test update.

configuration/tests.py
```diff
@@ -191,6 +191,103 @@ def test_legal_links_are_populated_urls(self):
...
+        """
+        branding_keys = ("theme", "logoSource", "logoClass")
+
+        for code, white_label_data in white_label_config.items():
+            for referrer, states in white_label_data.referrer_data.get("stateOptions", {}).items():
+                if referrer == "default" or len(states) < 2:
+                    continue
+
+                for state in states:
+                    state_referrer_data = white_label_config[state].referrer_data
```
catonph, line 287: `white_label_config[state]` indexes with an unvalidated code that came from a referrer's `stateOptions` config, and this line sits *outside* the `subTest` block below it. A typo in a state code therefore surfaces as a bare `KeyError` with no indication of which referrer or white label produced it.

Worse, alphabetical test ordering runs `test_multi_state_referrers_...` before `test_referrer_overrides_name_states_in_the_catalog` — the test written to diagnose exactly that typo — so the useful diagnostic gets buried behind this crash. Moving the lookup inside the `subTest`, or asserting `state in white_label_config` first with a message, would fix both.

configuration/white_labels/base.py
catonph: `publicly_launched = True` as the default (mirrored in `_template.py`) means any new state white label becomes publicly selectable in the dropdown on the next `add_config` run unless the author explicitly opts out — and no test enforces that they do.

The PR description already flags this as an open question for reviewers, so: I'd default it to `False`. The two failure modes aren't symmetric — a launched state missing from the dropdown is a visible, quickly-reported bug, whereas a pre-launch state exposed to the public dropdown sends real users into an unfinished screener. Opt-in-to-launch also matches how `ks`/`mo` are being handled here.

catonph (commented):
Code review for MFB-1672 (API side). Design is sound — the list-in-JSONField round-trips correctly via `OrderedJSONField`, and the `_default.py` addition is genuinely required since it replaces `referrer_data` rather than spreading the base. Four findings below, one of which pairs with a blocker on benefits-calculator#2187.

catonph (changes requested):
See comments - some are from me some are from Claude

PR #1733, opened 2026-08-31
Materialize the FPL thresholds into a table analytics can read [MFB-1182]

david-j-stoll:
## Context & Motivation

The Federal Poverty Limit dollar amounts are not in the database. `FederalPoveryLimit` stores only a `year` and a `period`; the thresholds themselves are `_FPL_DEFAULTS`, a module constant in `programs/models.py`, and `as_dict()` just looks the period up in it. The table named for the concept holds no values, which means anything that can only read Postgres cannot compute a percent-of-FPL figure.

MFB-1182 needs exactly that. The CESN heat pump dashboard has to segment by income band, and the analytics pipeline (dbt over Postgres, feeding Metabase) reads the database directly. This unblocks that without ever exposing a household's actual income — only which band it falls in.

- Related ticket: [MFB-1182](https://linear.app/myfriendben/issue/MFB-1182/cesn-heat-pump-user-engagement-analytics)
- Related PR: MyFriendBen/data-queries#136 (the consumer)

## Changes Made

- **New `FederalPovertyLimitValue` model** — one row per (period, household size) holding the output of `FederalPoveryLimit.get_limit()`. Sizes beyond `MAX_DEFINED_SIZE` are materialized with the per-additional-person amount already applied, up to 20, so a consumer joining on household size never reimplements that arithmetic. Unique constraint on (period, household_size).
- **`programs/fpl_values.py`** — `sync_fpl_values()` rewrites the table from the constant. Idempotent, deletes rows for periods the constant no longer defines, and takes a `dry_run` flag so the preview and the write share one diff implementation. Model class and FPL table are injectable so the data migration can pass its historical model from `apps.get_model()`.
- **Migration `0173`** — creates the table and performs the initial fill. Reversible.
- **`manage.py sync_fpl_values`** — with `--dry-run`, which reports the diff and **exits non-zero when the table is out of date**, so it can be wired up as a drift check.
- **Deploy hook** — `sync_fpl_values` now runs on every release, in `.github/actions/heroku-migrations` alongside the existing `sync_feature_flags` step. A migration runs once; this is what keeps the table current when a year is later added to the constant.
- **Tests** — 24 pass, covering the sync's correctness against `get_limit()`, idempotency, drift repair, stale-period deletion, extrapolated sizes, the uniqueness constraint, the migration's `populate()` called directly, and that `--dry-run` writes nothing while returning counts identical to a real sync.

**Nothing on the eligibility path changed.** `as_dict()` and `get_limit()` still read the constant, and every production calculator keeps resolving limits through them. This table is a read-only mirror for analytics.

## Testing

- Migrations to run: `python manage.py migrate programs` (applies `0173`, inserts 80 rows — 4 periods x 20 sizes)
- Manual testing steps:
  1. `python -m pytest programs/tests.py` — 24 pass. **Use pytest, not `manage.py test`**: `pytest.ini` sets `--nomigrations`, so the two runners build the schema differently and only pytest matches CI.
  2. `python manage.py migrate programs`
  3. `python manage.py sync_fpl_values --dry-run` — should print "FPL values are up to date." and exit 0
  4. Spot-check against the constant — a 4-person household in 2026 should be `33000`:
     `python manage.py shell -c "from programs.models import FederalPovertyLimitValue as V; print(V.objects.get(period='2026', household_size=4).annual_limit)"`
  5. Confirm eligibility is untouched: run an FPL-dependent program (`co_legal_services`, `ks_liheap`, `wa_head_start`) and check results are unchanged.
  6. Prove the drift alarm works: change one row's `annual_limit`, re-run `--dry-run`, confirm it reports OUT OF DATE and exits 1, then `sync_fpl_values` repairs it.

## Deployment

- Post-deployment scripts needed: none. The migration fills the table and the deploy hook keeps it current.
- Admin updates needed: none. The model is deliberately not registered in Django admin — it is a derived mirror, and an editable copy of the constant is exactly the drift this PR prevents.
- Notify team/users of: nothing user-facing. Worth flagging to anyone editing `_FPL_DEFAULTS` that the table syncs automatically on deploy, and `sync_fpl_values --dry-run` will tell them whether an environment is behind.
- **Ordering:** data-queries#136 depends on this. Its bridge model reads `programs_federalpovertylimitvalue`, so this must reach **production** (Thursday's release, not the staging auto-deploy) before that PR merges. Merging them the other way round fails the nightly Postgres build, which aborts the job before the BigQuery build and Metabase sync — breaking the pipeline for every tenant, not just CESN.

## Notes for Reviewers

Addressed from @catonph's review: migration renumbered to `0173` off `0172`; `--dry-run` no longer reimplements the diff; `bulk_create(ignore_conflicts=True)` dropped; `limits_for_period` guarded against a table with no integer sizes; the consumer contract documented on the model; duplicated rationale collapsed into the model docstring.

- **The parity claim in the original description was wrong and has been removed.** It said adding a year to the constant without syncing would fail CI. It would not: `setUp` syncs the table from the constant, so the assertion is a tautology over `sync_fpl_values`, and CI's freshly-built database can never observe production's state. That is an environment invariant no unit test can close. The fix is the deploy hook, not a cleverer test — the tests now say explicitly what they do and do not cover.
- **Two review points were not adopted, with reasons.** The migration's `print` stays: migrations `0129`, `0166`, `0168` and `0169` all print progress, so it is house style. And `period` did not become a FK: `FederalPoveryLimit.period` is not unique (only `year` is), so there is no single row to point at; the model now documents why the loose key is intentional.
- **One review premise did not hold.** The concern that a household above the size cap "joins to nothing and gets a NULL band" does not apply to the actual consumer — the dbt bridge clamps with `least(greatest(household_size, 1), max(household_size))`, so it gets the largest available band. The cap is nonetheless now documented as a contract, since it is a cross-repo coupling.
- Known limitations: the constant remains the source of truth, so there are two representations of the same numbers. Deliberate here, to keep the eligibility path untouched.
- Future considerations: `as_dict()` could be collapsed onto this table so there is a single source of truth, which would also fix the oddity that a `period` pointing at a non-existent key fails at eligibility time rather than at save time. The read path is unusually well positioned for it — 61 call sites, all through `get_limit()`/`as_dict()`, none touching `_FPL_DEFAULTS` directly — so it is two method bodies rather than 61 call sites. The obstacle is performance: `as_dict()` runs ~56x per screen and is currently a dict lookup on a module constant, so a DB read needs prefetching or a request-scoped cache, and this codebase has been burned there before (see the `_get_fpl_data` docstring). Worth its own ticket.

### DIFF
--- a/.github/actions/heroku-migrations/action.yml
+++ b/.github/actions/heroku-migrations/action.yml
@@ -45,6 +45,30 @@ runs:
           exit $STATUS
         fi
 
+    - name: Sync FPL values
+      id: fpl-values
+      shell: bash
+      env:
+        HEROKU_API_KEY: ${{ inputs.heroku-api-key }}
+      run: |
+        # FederalPovertyLimitValue mirrors the _FPL_DEFAULTS constant into a table
+        # so SQL-only consumers (the analytics pipeline) can compute a
+        # percent-of-FPL band. Migration 0173 does the initial fill, but a
+        # migration only runs once -- adding a year to the constant later would
+        # leave the table behind, and the dashboard would band households against
+        # stale thresholds. Syncing on every deploy means that cannot be
+        # forgotten. Idempotent: a no-op when nothing changed.
+        echo "Syncing FPL values..."
+        set +e
+        OUTPUT=$(heroku run --exit-code -a ${{ inputs.heroku-app-name }} "python manage.py sync_fpl_values" 2>&1)
+        STATUS=$?
+        set -e
+        echo "$OUTPUT"
+
+        if [ $STATUS -ne 0 ]; then
+          exit $STATUS
+        fi
+
     - name: Sync feature flags
       id: feature-flags
       shell: bash
--- a/programs/fpl_values.py
+++ b/programs/fpl_values.py
@@ -1,26 +1,13 @@
-"""Materialize the FPL table into the database so SQL-only consumers can read it.
+"""Sync the FederalPovertyLimitValue mirror from the _FPL_DEFAULTS constant.
 
-The thresholds themselves live in `_FPL_DEFAULTS`, a module constant in
-programs.models, and `FederalPoveryLimit` stores only a year and a period that
-point into it. That is fine for the calculators, which go through
-`FederalPoveryLimit.get_limit()` in Python, but it leaves anything that can only
-read the database with no way to compute a percent-of-FPL figure -- the analytics
-pipeline (dbt over Postgres, feeding Metabase) among them.
-
-This module writes `get_limit()`'s output to `FederalPovertyLimitValue`, one row
-per (period, household size). It is deliberately a mirror, not a new source of
-truth: the constant stays authoritative, the calculators keep reading it, and a
-test asserts the two agree so they cannot drift silently.
-
-Sizes above `MAX_DEFINED_SIZE` are extrapolated exactly as `get_limit()` does,
-up to `MAX_MATERIALIZED_SIZE`, so a consumer joining on household size never has
-to reimplement the per-additional-person arithmetic.
+See the FederalPovertyLimitValue docstring in programs/models.py for why the
+mirror exists and what it guarantees to consumers.
 """
 
 from typing import Optional, Type
 
-# Imported lazily inside the functions so a data migration can hand in its own
-# historical model class via apps.get_model().
+# Largest household size given a row. Consumers must clamp to this; see the
+# CONTRACT WITH CONSUMERS note on the model.
 MAX_MATERIALIZED_SIZE = 20
 
 
@@ -31,6 +18,8 @@ def limits_for_period(table: dict, max_size: int = MAX_MATERIALIZED_SIZE) -> dic
     the per-additional-person amount once per extra member.
     """
     defined = sorted(size for size in table if isinstance(size, int))
+    if not defined:
+        raise ValueError("FPL table defines no household sizes; cannot expand it.")
     largest_defined = defined[-1]
 
     limits = {size: table[size] for size in defined}
@@ -44,11 +33,16 @@ def sync_fpl_values(
     value_model: Optional[Type] = None,
     fpl_table: Optional[dict] = None,
     max_size: int = MAX_MATERIALIZED_SIZE,
+    dry_run: bool = False,
 ) -> dict[str, int]:
     """Upsert one row per (period, household size). Idempotent.
 
     Returns a {created, updated, deleted} count so the management command and the
     migration can report what they did.
+
+    `dry_run` returns the same counts without writing. The diff is computed once
+    either way -- a separate preview implementation would drift from the real one
+    and start lying about what a sync would do.
 
     `value_model` and `fpl_table` are injectable so a data migration can pass the
     historical model from apps.get_model() rather than importing the live one.
@@ -76,22 +70,24 @@ def sync_fpl_values(
     for (period, household_size), annual_limit in wanted.items():
         row = existing.get((period, household_size))
         if row is None:
-            value_model.objects.create(
-                period=period,
-                household_size=household_size,
-                annual_limit=annual_limit,
-            )
             created += 1
+            if not dry_run:
+                value_model.objects.create(
+                    period=period,
+                    household_size=household_size,
+                    annual_limit=annual_limit,
+                )
         elif row.annual_limit != annual_limit:
-            row.annual_limit = annual_limit
-            row.save(update_fields=["annual_limit"])
             updated += 1
+            if not dry_run:
+                row.annual_limit = annual_limit
+                row.save(update_fields=["annual_limit"])
 
     # Drop rows for periods or sizes the constant no longer defines, so a removed
     # year cannot linger and be joined against.
     stale = [row.pk for key, row in existing.items() if key not in wanted]
     deleted = len(stale)
-    if stale:
+    if stale and not dry_run:
         value_model.objects.filter(pk__in=stale).delete()
 
     return {"created": created, "updated": updated, "deleted": deleted}
--- a/programs/fpl_values.py
+++ b/programs/fpl_values.py
@@ -1,0 +1,97 @@
+"""Materialize the FPL table into the database so SQL-only consumers can read it.
+
+The thresholds themselves live in `_FPL_DEFAULTS`, a module constant in
+programs.models, and `FederalPoveryLimit` stores only a year and a period that
+point into it. That is fine for the calculators, which go through
+`FederalPoveryLimit.get_limit()` in Python, but it leaves anything that can only
+read the database with no way to compute a percent-of-FPL figure -- the analytics
+pipeline (dbt over Postgres, feeding Metabase) among them.
+
+This module writes `get_limit()`'s output to `FederalPovertyLimitValue`, one row
+per (period, household size). It is deliberately a mirror, not a new source of
+truth: the constant stays authoritative, the calculators keep reading it, and a
+test asserts the two agree so they cannot drift silently.
+
+Sizes above `MAX_DEFINED_SIZE` are extrapolated exactly as `get_limit()` does,
+up to `MAX_MATERIALIZED_SIZE`, so a consumer joining on household size never has
+to reimplement the per-additional-person arithmetic.
+"""
+
+from typing import Optional, Type
+
+# Imported lazily inside the functions so a data migration can hand in its own
+# historical model class via apps.get_model().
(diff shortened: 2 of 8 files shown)


catonph (changes requested):
CR: PR #1733 — Materialize FPL thresholds into a table
Verified locally: 20/20 tests pass, eligibility path genuinely untouched (as_dict()/get_limit() unmodified).

P0 — Blocks merge
1. Migration number collision will break migrate

The PR adds 0171_federal_poverty_limit_value.py depending on 0170_deactivate_il_rent_asst. But main already has 0171_alter_programcategory_white_label.py and 0172_consolidate_shared_program_categories.py, both also descending from 0170. Merging gives you two leaf migrations in programs and migrate refuses with "Conflicting migrations detected; multiple leaf nodes."

This branch predates the MFB-1601 merge (PR #1741). The file is named 0170 in the PR body but 0171 on the branch — already bumped once for this reason, needs one more.

Fix: rebase onto main, renumber to 0173_..., dependencies = [("programs", "0172_consolidate_shared_program_categories")].

2. The parity test does not catch the drift it's advertised to catch
This is the thing the PR body explicitly asks reviewers to scrutinize:

The thing worth scrutinising is the parity test … adding a year to the constant without syncing fails CI rather than quietly serving last year's thresholds.

It doesn't. [programs/tests.py:125-126](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/tests.py#L125-L126):

```
def setUp(self):
    sync_fpl_values()
```

Every test syncs the table from the constant first, so the assertion is a tautology over sync_fpl_values. Add "2027" to _FPL_DEFAULTS, never run the command: CI builds a fresh DB, syncs it in setUp, passes green — while prod's table has no 2027 rows and the dashboard silently bands against 2026.

The setUp is defensible on its own terms (the docstring correctly notes --nomigrations means RunPython never runs). But the consequence is that CI structurally cannot detect prod drift — CI's DB is always freshly synced, prod's isn't. That's an environment invariant; no unit test can close it.

Fix, in preference order: add the sync to the release phase so it can't be forgotten (per the Heroku setup, release runs migrate but not config commands, so this is an explicit addition); or run sync_fpl_values --dry-run as a post-deploy check that alarms non-zero. At absolute minimum, strike the parity claim from the PR body and the [class docstring](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/tests.py#L113-L118) so the next person doesn't trust a net that isn't there.

3. --dry-run reimplements the diff logic
[sync_fpl_values.py](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/management/commands/sync_fpl_values.py) computes created/updated/deleted counts again, separately from sync_fpl_values(). Classic dry-run-drifts-from-real-run bug: change the sync logic and the preview silently lies. Refactor to sync_fpl_values(dry_run=False) — compute the diff once, write only when not dry_run. Saves ~15 lines and removes the hazard.

P1 — How hard to kill the codebase version?

The read path is already funneled through exactly two methods: [get_limit()](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/models.py#L83) and [as_dict()](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/models.py#L92). I counted the call sites: ~56 across the calculators, every one going through those two methods. Not a single calculator touches _FPL_DEFAULTS directly — they all go self.program.year.as_dict() or .get_limit(...). Unusually good position. Collapsing to one source of truth is two method bodies, not 56 call sites:

```
def as_dict(self):
    rows = self.values.all()          # related_name on a new FK
    if not rows:
        raise ...                     # preserves today's KeyError-on-bad-period behavior
    d = {r.household_size: r.annual_limit for r in rows if r.household_size <= self.MAX_DEFINED_SIZE}
    d["additional"] = self.additional_amount
    return d
```

Cost: one model field (additional), two method bodies, plus a caching decision. Benefit: no mirror, no drift class (P0 #2 disappears entirely), no sync_fpl_values runbook step, and the "period pointing at a nonexistent key fails at eligibility time" oddity the PR files as future work gets fixed free.

The one real obstacle — and the legitimate reason to split: as_dict() runs ~56× per screen, currently a dict lookup on a module constant. Making it a DB read puts 56 queries on the eligibility hot path. Solvable via prefetch_related or a request-scoped cache, but this codebase already got burned here — [the _get_fpl_data docstring](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/models.py#L60) documents ripping out a Redis cache precisely because it added a network hop per read. So it's not purely mechanical.

Recommendation: ship this PR for the MFB-1182 unblock, but file the consolidation ticket before merge, with the plan above as acceptance criteria. A mirror with a named successor is a bridge; a mirror without one is permanent — and given the runbook burden in P0 #2, we want that ticket to exist, not be promised.

P2 — Correctness worth fixing
MAX_MATERIALIZED_SIZE = 20 is an unenforced ceiling. get_limit() extrapolates to any size; the table stops at 20. A 21-person household joins to nothing and gets a NULL band instead of an error — a silent wrong-answer path in the consumer. Document the cap as a contract in the model docstring, or make the consumer's join inner so it fails loudly.

period vs year is a modeling trap. FederalPoveryLimit has both; [import_program_config.py:924](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/management/commands/import_program_config.py#L924) does get(year=year_value, period=year_value), so they're equal today but the model doesn't require it. The new table keys on period with no FK. Defensible for a denormalized analytics mirror, but nothing prevents a FederalPoveryLimit row whose period has no matching values — exactly the [non-numeric-period case SNAP already warns about](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/programs/cross_white_label/snap/base.py#L92). Either add the FK or state why the loose key is intentional.

bulk_create(..., ignore_conflicts=True) masks a real signal. A fresh table can't conflict, so the flag only hides the case where the table isn't fresh (re-run after partial clear()). Dropping it surfaces that instead. The flag is buying nothing and costing you an error.

limits_for_period assumes non-empty. [fpl_values.py](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/fpl_values.py) does defined[-1] unguarded — empty or int-key-less table is an IndexError. [test_every_year_defines_all_household_sizes](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/tests.py#L103) makes it unreachable today. One-line guard.

P3 — Brevity and polish
Functional code is ~145 lines; comments/docstrings are comparable in volume and repeat the same rationale four times:

- [fpl_values.py](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/fpl_values.py) module docstring — 18 lines
- [models.py:99](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/models.py#L99) model docstring — 17 lines
- [migration 0171](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/migrations/0171_federal_poverty_limit_value.py) header — 18 lines
- [tests.py:113](vscode-webview://1jrihtje43b7cumvhk5dugrllr34t814skp8ljhjmlal1lcpccmr/benefits-api/programs/tests.py#L113) class docstring — 13 lines

All four say: constant is source of truth, this is a mirror, analytics needs SQL, a test guards drift. Keep one canonical version (the model docstring — where a reader lands first), cut the rest to a one-line pointer. ~60 lines.

Per your comments-describe-current-behavior convention, several also drift into changelog territory: the _get_fpl_data "it used to sit behind a cache whose update() opened with return self.default" narrative and the migration's "a data migration reading app code is normally a smell, but…" justification both explain decisions, not current behavior. Those belong in the PR description — which already has them.

Smaller items:

- print(f"✅ materialized {len(rows)} FPL threshold rows") — bare print in a migration is deploy-log noise.
- from typing import Optional, Type — the "imported lazily inside the functions" comment is explaining the function-body imports but sits above MAX_MATERIALIZED_SIZE.
- PR body says 19 tests; actual is 20.

PR #1741, opened 2026-09-01
MFB-1601: Consolidate ProgramCategory onto shared rows

catonph:
Closes [MFB-1601](https://linear.app/myfriendben/issue/MFB-1601/consolidate-programcategory-onto-shared-rows-across-white-labels). Also closes [MFB-1706](https://linear.app/myfriendben/issue/MFB-1706/trump-accounts-dont-stack-under-cash-assistance-should-move-to-savings) (see below). Supersedes #1739.

## Problem

Every white label had its own row for the same semantic category — **64 rows across 10 white labels**, 8 of them "Cash Assistance". Nothing kept them consistent, and they had already drifted:

- `health_care` was spelled **"Health Care"**, **"Healthcare"** and **"Health Coverage"**
- `child_care` was **"Child Care"** in TX and KS, **"Child Care, Youth, and Education"** elsewhere
- `nc taxCredits` — the only `external_name` with a space and camelCase
- KS had both `ks_health_care` and `ks_healthcare`, each with programs

## Approach

`ProgramCategory.white_label` becomes **nullable**, where null means shared. Programs in any white label point at the same row, so a category exists **once** and cannot drift — rather than being policed into agreement by a constraint.

**64 rows → 15**: 10 shared + 5 genuinely white-label-specific (CESN's four own categories and the CO tax calculator, all untouched).

A CO program and an MO program now point at the same `child_care` row.

`external_name` **keeps** `unique=True`. Once each name exists once, the global constraint becomes correct rather than an obstacle — which is also why the unscoped lookup in `models.py` is now right instead of a latent bug.

## Shared categories (10)

| `external_name` | Icon | Name | `tax_category` |
|---|---|---|---|
| `cash` | cash | Cash Assistance | false |
| `food` | food | Food and Nutrition | false |
| `health_care` | health_care | Health Care | false |
| `housing` | housing | Housing and Utilities | false |
| `child_care` | child_care | Child Care and Youth | false |
| `tax_credit` | tax_credit | Tax Credits | **true** |
| `transportation` | transportation | Transportation | false |
| `employment` | job_resources | Employment | false |
| `savings` | savings | Savings | false |
| `education` | education | Education and Training | false |

CO already held the unprefixed names, so its 7 rows are **promoted** by nulling `white_label` — preserving their ids and Translation labels instead of re-pointing every CO program.

**Copy changes** (a shared row can only carry one name): TX "Healthcare" → "Health Care", TX "Child Care" → "Child Care, Youth, and Education" → now "Child Care and Youth", IL "Health Coverage" → "Health Care". Names are machine-translated into every supported language.

## New Education category + audit

Education and training had nowhere to go, so scholarships sat under child care. An audit of all 243 programs and every config found exactly four:

| WL | Program | What it is |
|---|---|---|
| wa | `wa_wsos_bas` | STEM/health care undergraduate scholarship |
| wa | `wa_wsos_cts` | Associate degree, certificate, apprenticeship |
| wa | `wa_wsos_grd` | Nurse practitioner graduate scholarship |
| ks | `ks_promise_act` | Community/technical college tuition, fees, books |

`child_care` is renamed to **"Child Care and Youth"** since education is no longer part of it. Early-childhood and K-12 programs stay put: Head Start, Early Head Start, preschool, afterschool, NSLP, CCAP/CCS child-care subsidies.

Deliberately **not** moved: `co_collegeinvest_first_step` and `ma_bsp` are 529 *savings* products (→ `savings`); adult-education and literacy items are **urgent needs**, which use a separate `UrgentNeedCategory` model.

## MFB-1706 folded in

Rather than create 7 prefixed `*_savings` rows for this PR to immediately consolidate, MFB-1706 ships here: `trump_account` moves from cash → `savings` in all 8 white labels, plus `co_collegeinvest_first_step` from child care.

Trump Accounts is a custodial investment account locked under IRA rules until the child turns 18. Because `CategoryHeading` divides a category total by 12, the one-time $1,000 was rendering as **~$83/month of recurring cash** inside Cash Assistance — the "doesn't stack" symptom the partner reported.

## Three bugs this surfaced

1. **Cross-white-label data leak.** `ProgramCategorySerializer.get_programs()` returned *every* program in a category — safe only while a category belonged to one white label. On a shared `cash` row that's 21 programs across 6 states, so a CO user would have seen MO's programs. Now filtered by the requested white label, with a test.
2. **Shared categories would have been unselectable.** `SecureAdmin._set_select_queryset` filters a program's category dropdown to `white_label=obj.white_label`, which excludes null — nobody could have assigned a shared category to a program.
3. **`import_program_config` would fail on every shared category.** It looked up by `(external_name, white_label)`, missing the shared row and then violating the unique constraint. Verified by actually importing all four education programs.

## Admin

Shared rows stay visible and editable to any white-label admin (`SecureAdmin` would otherwise hide them from every non-superuser, since it filters `white_label__in`). Added a "Shared" column and a **warning on the change form** naming the white labels an edit affects — with one row, a KS admin renaming "Cash Assistance" changes it everywhere, and nothing else on the page said so.

## Testing

- **3783 passed, 0 failed** across `programs/ screener/ translations/ validations/ configuration/` on a fresh test DB.
- 9 new tests for shared-row behavior: null white label, cross-white-label sharing, serializer isolation, and the endpoint returning a shared category per white label.
- Migration verified on a local DB: 64 rows → 15, **zero programs lost a category**, verified against a before/after snapshot of all 243 programs.
- **Re-runnable after a partial failure** — verified by re-running against a half-migrated state; it adopts an existing same-named row rather than creating a duplicate.
- Imported all four education programs to confirm the shared-category import path.

## Review fixes

**1. `calc_max_cap` crash (blocker) — fixed.** `max(*values)` raised `TypeError` on a single value, taking down `all_results()` for the whole screen. One value is the *common* case: exactly one of the capped programs present for a member. Now `max(values, default=0)`.

Pre-existing, but this PR widened its reach — `co_health_care` previously only ran on CO screens and now sits on a shared row. My original test passed `{}`, the one case that *doesn't* crash, so the green suite was not evidence. Replaced with single- and multi-value cases; verified the new test fails without the fix.

On also nulling `calculator` on shared rows: that doesn't work, and it's worth saying why. CO's programs now live *on* the shared rows, and the cap is read from `program.category.calculator` — so a CO-scoped row would hold zero CO programs and the $8,640 preschool cap would silently never fire. The decoupling is the right instinct, but it needs the cap keyed off something other than the category row. Filed as follow-up rather than half-done here.

**2. `set_name` discarding human translations — fixed.** Confirmed the mechanism: `add_translation` always marks English `edited=True`, and `edit_translation_by_id`'s `manual=False` guard only spares a row that is *both* `no_auto` and `edited` — categories are neither. `set_name` now skips a category whose English already matches, so the five unchanged categories keep their curated copy.

Verified by seeding `'CURATED Asistencia'` on `cash`, running the migration, and confirming it survived — while `child_care` (English genuinely changed) was correctly renamed and re-translated.

**3. `priority` never normalized — fixed.** Added to the `SHARED_CATEGORIES` tuple and set explicitly, with a comment on why: it's serialized to the FE, which sorts categories by it *ahead of* value ([Programs.tsx](https://github.com/MyFriendBen/benefits-calculator/blob/main/src/Components/Results/Programs/Programs.tsx)), so a value carried over from CO would silently reorder results for all 8 white labels. All ten ship `null` = sort by value, which is today's behaviour everywhere.

## Deploy verification

Took the reviewer's point that row counts wouldn't catch any of the above. Added **`audit_program_categories`**, a read-only command that dumps the rendered payload per white label — each category's icon, `priority`, `calculator`, per-language names, and each white label's own programs grouped by category:

```
python manage.py audit_program_categories > before.json
python manage.py migrate
python manage.py audit_program_categories > after.json
diff <(jq -S . before.json) <(jq -S . after.json)
```

It surfaced the seeded translation change and flagged `co_care` as uncategorized, so it does catch what counts alone would miss. Run this on the staging restore rather than trusting the row totals.

## Deploy notes

⚠️ **Take a database snapshot first.** The data migration is intentionally **irreversible** (reverse is a no-op): consolidation is lossy — rows collapse into one and their orphaned Translation rows are deleted, so per-white-label names and ids can't be reconstructed.

⚠️ **My inventory is from a local DB.** The ticket's row list was written 2026-08-06 and my local DB already differs. The migration is defensive — it skips rows that don't exist and never deletes a row that still has programs — but **dry-run against a staging restore of prod before shipping.**

Needs the FE companion for the icon (falls back to a generic dot without it): MyFriendBen/benefits-calculator#2190


programs/migrations/0172_consolidate_shared_program_categories.py
```diff
@@ -0,0 +1,336 @@
+# Consolidates ProgramCategory onto shared rows for MFB-1601.
```
david-j-stoll, line 1: Not sure if we want to keep our standard of keeping out ticket names here, given it's a migration. Let me know your thoughts.

catonph: I thought since migration it was fine to leave and to let Claude do its verbose thing but open to push back! I know I'm usually the strong anti-ticket one...

david-j-stoll (dismissed):

programs/admin.py
```diff
@@ -545,12 +545,61 @@ def action_buttons(self, obj):
...
 class ProgramCategoryAdmin(SecureAdmin):
     search_fields = ("external_name",)
-    list_display = ["get_str", "external_name", "action_buttons"]
+    list_display = ["get_str", "external_name", "shared", "action_buttons"]
     exclude = ["name", "description"]
 
     def has_add_permission(self, request):
         return False
 
+    def get_queryset(self, request):
```
david-j-stoll: Add type hints here, and to the rest of the file.

david-j-stoll (approved):

PR #1759, opened 2026-09-04
MFB-1760: add migrations to backfill missing seed data

elanatee:
## Context & Motivation

- Fixes [MFB-1760](https://linear.app/myfriendben/issue/MFB-1760/create-new-migration-to-fix-seed-data-gaps-when-setting-up-benefits).
- While setting up `benefits-api` locally, several bugs surfaced that traced back to missing/incorrect data rather than code bugs: the referral-source dropdown had no options, eligibility calculation crashed with `KeyError: None` (blank `state_code`), and crashed again with a `KeyError` on specific programs (`has_calculator` incorrectly `True` on tracking-only programs).
- Root cause: `python manage.py bulk_import < translations.json` (a required step in the local setup docs) does more than import translation text — it fully creates or updates any `Program` it's given data for, and as a side effect creates missing `WhiteLabel` rows too (see `programs/models.py`, `ProgramDataController.from_model_data()` and `ProgramManager.new_program()`). Neither of those code paths ever sets `state_code` on the white label or `has_calculator` on the program, so anything created this way silently ends up wrong. Referrer rows are missing separately, because the migration that originally seeded them (`0145_seed_referrer_rows_from_referral_options.py`) only covered white labels that existed at the time it ran — `il`, `tx`, and `wa` were created afterward and never got seeded.
- Per team discussion, these are one-time migrations rather than logic in `add_config`, since `add_config` is meant for keeping config files in sync with the database on an ongoing basis, not for one-off data repairs.

## Changes Made

- `programs/migrations/0176_create_state_white_labels.py` — for each real state white label (`co`, `il`, `wa`, `tx`, `ma`, `nc`, `ks`, `mo`, `cesn`):
  - Creates the `WhiteLabel` row with the correct display name and `state_code` if it doesn't exist yet, or backfills `state_code` on it if it already exists but is blank.
  - Seeds the 7 standard generic `Referrer` codes (`flyers`, `friend`, `merit`, `other`, `searchEngine`, `socialMedia`, `testOrProspect`) as `is_partner=False`, correcting any pre-existing `merit` row that `0145` mis-classified as a partner.
- `programs/migrations/0177_create_gap_tracking_programs_with_calculator_flag.py` — for the 4 known tracking-only programs (`co_andso`, `co_section_8`, `ma_section_8`, `co_care`): creates them with `has_calculator=False` from the start if they don't exist yet (reusing `0141_create_gap_tracking_programs.py`'s creation logic, adjusted for schema drift — `value_type_id` was dropped from `programs_program` after `0141` was written and had to be removed from the INSERT), or corrects `has_calculator` back to `False` if the row already exists with the wrong value. Depends on `0176` so the white labels it needs already exist.
- Both migrations are idempotent (`get_or_create`, `ON CONFLICT DO NOTHING`, targeted `.update()` calls) and safe to re-run.
- No changes to `add_config` or the white label config files in this PR — an earlier version of this work put the fixes there, but per team discussion they've moved to migrations instead.

## Testing

- Migrations to run: `python manage.py migrate` (applies `0176` then `0177`).
- Configuration updates needed: none.
- Manual testing steps, both tested directly by Claude:
  1. **Partially-seeded database** (mirrors staging/production/an existing local setup): blanked `state_code` on an existing white label, deleted its referrer rows, and reset the 4 tracking programs' `has_calculator` to `True`. Ran `migrate` and confirmed all three were corrected without touching any other white label's data.
  2. **Genuinely empty database** (mirrors a new engineer's actual setup flow — `migrate` before `bulk_import` ever runs): ran `migrate` against a database with zero prior data and confirmed all 9 state white labels were created with correct `state_code` and referrers, and all 4 tracking programs were created with `has_calculator=False` from birth — meaning `bulk_import`, which never touches `has_calculator`, can no longer leave these wrong, since it only updates an already-existing row's other fields rather than creating a fresh one.
- I also manually ran through the set up steps in the [Getting Started wiki](https://github.com/MyFriendBen/benefits-api/wiki/Get-Started) to verify that this works. I was able to complete the rest of the steps, starting from a fresh database, and run the screener locally with the errors I was previously seeing resolved. 

## Deployment

- Post-deployment scripts needed: none — `python manage.py migrate` already runs automatically on every staging/production deploy via `.github/actions/heroku-migrations`.
- Production config updates: none.
- Admin updates needed: none.
- Notify team/users of: nothing required, but worth a quick spot-check of `il`/`tx`/`wa`'s referral options and `state_code` on staging/production after this deploys, since we haven't directly verified those environments' current data — only that this migration will correct them automatically if they're wrong.

## Notes for Reviewers

- Known limitations:
  - `0177`'s program-creation logic only covers the 4 tracking-only programs known today (matching `0141`'s original list). It doesn't fix the underlying reason `has_calculator` can end up wrong for *any* future tracking-only program — that's `bulk_import`'s `ProgramDataController` never syncing that field at all (see `programs/models.py`, `ProgramManager.new_program()` line ~548 and `from_model_data()`). A new tracking-only program added later would need the same treatment unless that's fixed at the root.
  - `bulk_import` can silently create a `WhiteLabel` row for a code that has no corresponding entry in `configuration/white_labels/__init__.py`'s registry (e.g. `co_tax_calculator` exists in our DB this way). Neither of these migrations covers a white label in that state, since they only act on the known state-code list. [Documentation for this gap still to be added.]
  - Both migrations use raw SQL instead of the ORM for `Referrer`/`Program` inserts — required, not a style choice: `apps.get_model()` returns a historical model that django-parler never registers `_parler_meta` on, so `.objects.create()` raises `AttributeError: 'NoneType'.get_all_fields()` (same issue `0141`/`0145`/`0159`/`0164` already worked around).
- Future considerations:
  - Consider fixing `bulk_import`'s `ProgramDataController` to actually sync `has_calculator` — this would close the gap at the root for any program, not just today's known 4, but it's a bigger change since that same import command also runs against staging and production.
  - Consider a model-level constraint requiring `state_code` on any non-`_default` white label, so a missing value fails loudly at creation time.

## Summary by CodeRabbit

- **New Features**
  - Added state-specific white-label configurations, including CESN, with associated state codes.
  - Added generic referral options for supported programs.
  - Added four gap-tracking programs with localized display names and descriptions.

- **Updates**
  - Gap-tracking programs are now configured without calculator functionality.
  - Existing program and translation details are updated to reflect the new configurations.


screener/migrations/0165_backfill_whitelabel_state_code.py
```diff
@@ -0,0 +1,46 @@
...
+    for code, state_code in STATE_CODES.items():
+        # SQL's IN never matches NULL, so this can't be state_code__in=["", None] —
+        # that silently matches zero rows against a NULL column (learned the hard
+        # way testing this migration locally).
+        WhiteLabel.objects.filter(Q(code=code) & (Q(state_code="") | Q(state_code__isnull=True))).update(
+            state_code=state_code
+        )
+
+
+def reverse_backfill(apps, schema_editor):
```
david-j-stoll: Do these backfills support when DBs are empty or do they only support DBs that are partially seeded? Just wondering as if both workflows aren't supported the reverse backfill may be worth digging into.

It may make more sense to include this in `add_config`, so the process is idempotent? Worth running that through Claude to see if that makes sense given the two workflows. To be clear, the migrations are already idempotent. The gap is that they run before any white label rows exist on a fresh DB, and only once. `add_config` runs after white label creation and re-runs on demand, which is what would cover both workflows.

david-j-stoll (changes requested):
One comment to address. I think it could be worth testing (either manually or through an automated test) to determine whether or not both workflows are supported. This includes both seeding an empty db and populating a partially-seeded db.

I'd also recommend moving the PR out of draft form, then letting Code Rabbit run. Sometimes there are some good recommendations there as well!

screener/migrations/0165_backfill_whitelabel_state_code.py
elanatee: Ah, you're right. 2 out of these 3 migrations only support DBs that have been partially seeded with white labels. The `has_calculator` migration sufficiently corrects the one-time mistake of those four gap tracking programs being marked as having a calculator. 

It looks like the step for referrer seeding could be added to `add_config`, but a separate approach is needed to seed white labels + their state codes. Looking into it now!

configuration/management/commands/add_config.py
```diff
@@ -5,9 +5,29 @@
...
+    "other": "Other",
+    "searchEngine": "Google or other search engine",
+    "socialMedia": "Social Media",
+    "testOrProspect": "Test / Prospective Partner",
+}
+
+# Tracking-only programs (no eligibility calculator) that both
+# 0141_create_gap_tracking_programs.py's "existing row" branch and
+# bulk_import's ProgramDataController (which never syncs has_calculator)
+# can leave at the model default of True. MFB-1760.
```
david-j-stoll, line 28: I would remove this ticket ref, this will also become stale pretty quick

configuration/management/commands/add_config.py
```diff
@@ -46,6 +66,19 @@ def handle(self, *args, **options):
                 self.stdout.write(self.style.WARNING(f'White label for "{white_label_code}" is not in the database'))
                 continue
 
+            # Set state_code on the WhiteLabel row itself (not a Configuration
+            # entry). Nothing else populates this field automatically — required
+            # by SMI/income-limit lookups. MFB-1760.
```
david-j-stoll, line 71: Ticket ref, same as above

configuration/management/commands/add_config.py
```diff
@@ -46,6 +66,19 @@ def handle(self, *args, **options):
...
+            white_label.state_code = WhiteLabelData.state_code
+            white_label.save()
+
+            # Correct has_calculator for known tracking-only programs that
+            # bulk_import or legacy migrations may have left at the model
+            # default (True). MFB-1760.
```
david-j-stoll, line 77: Ticket ref, same as above

configuration/management/commands/add_config.py
```diff
@@ -104,6 +137,26 @@ def handle(self, *args, **options):
                 defaults={"data": WhiteLabelData.override_text, "active": True},
             )
 
+            # Ensure the standard generic referral-source dropdown options
+            # exist. Without these, a newly-launched white label's referral
+            # source step has nothing selectable until someone adds them by
+            # hand (MFB-1760).
```
david-j-stoll, line 143: Ticket ref, same as above

configuration/management/commands/add_config.py
```diff
@@ -5,9 +5,29 @@
...
+    "searchEngine": "Google or other search engine",
+    "socialMedia": "Social Media",
+    "testOrProspect": "Test / Prospective Partner",
+}
+
+# Tracking-only programs (no eligibility calculator) that both
+# 0141_create_gap_tracking_programs.py's "existing row" branch and
+# bulk_import's ProgramDataController (which never syncs has_calculator)
+# can leave at the model default of True. MFB-1760.
+GAP_TRACKING_PROGRAMS = ["co_andso", "co_section_8", "ma_section_8", "co_care"]
```
catonph, line 29: why do we need to list these explicitly instead of relying on checking has_calculator in our filter below?

configuration/management/commands/add_config.py
catonph, line 71: Let's keep ticket numbers out of comments

programs/migrations/0177_create_gap_tracking_programs_with_calculator_flag.py
```diff
@@ -0,0 +1,177 @@
...
+        # (see 0141_create_gap_tracking_programs.py, which hit the same issue).
+        with db.cursor() as cursor:
+            cursor.execute(
+                """
+                INSERT INTO programs_program (
+                    name_abbreviated, external_name, year_id,
+                    active, low_confidence, has_calculator,
+                    show_in_has_benefits_step, show_on_current_benefits,
+                    base_program, white_label_id,
+                    name_id, description_short_id, description_id,
```
catonph, line 122: Heads up — this line was copied from `0141`, where `external_name` was still `null=True`. Since `0148_alter_program_external_name` it's `CharField(max_length=120, unique=True)`, so `None` is no longer a legal value and this INSERT would hit a not-null violation and abort `migrate` mid-deploy.

It only trips if some row already holds one of these four values under a different `(white_label__code, name_abbreviated)` — so it may never fire in practice, but it's a hard failure in a migration if it does.

`ProgramManager.new_program` already defines the convention for exactly this collision (programs/models.py:529-543): fall back to `{white_label}_{name_abbreviated}`, and raise if that's taken too. Worth mirroring here so both paths agree:

```python
if not Program.objects.filter(external_name=name_abbreviated).exists():
    external_name = name_abbreviated
else:
    external_name = f"{p['white_label_code']}_{name_abbreviated}"
    if Program.objects.filter(external_name=external_name).exists():
        raise ValueError(f"Cannot generate unique external_name. Conflict on '{external_name}'")
```

programs/migrations/0176_create_state_white_labels.py
```diff
@@ -0,0 +1,93 @@
...
+}
+
+
+def create_state_white_labels(apps, schema_editor):
+    WhiteLabel = apps.get_model("screener", "WhiteLabel")
+    db = schema_editor.connection
+
+    for code, (name, state_code) in STATE_WHITE_LABELS.items():
+        white_label, _ = WhiteLabel.objects.get_or_create(
+            code=code,
```
catonph: Minor robustness thought: `WhiteLabel.code` has no unique constraint (screener/models.py:53), and `bulk_import` creates white labels via bare `WhiteLabel.objects.create()` in several places — so a database that already has two `co` rows would make this `get_or_create` raise `MultipleObjectsReturned` and fail the deploy.

`configuration/views.py:46` already guards against this same case with `.filter(...).order_by("id").first()`. Since this migration exists specifically to repair databases that drifted, it seems worth being defensive here too rather than assuming the column is clean.

programs/migrations/0177_create_gap_tracking_programs_with_calculator_flag.py
```diff
@@ -0,0 +1,177 @@
+from django.db import migrations
+
+# Same list as 0141_create_gap_tracking_programs.py's GAP_TRACKING_PROGRAMS.
+# That migration skips creating a program when its white label doesn't exist
+# yet (WhiteLabel.DoesNotExist: continue) — true on a fresh database, since
+# these white labels didn't exist when 0141 first ran. This migration runs
+# after 0175_create_state_white_labels, which guarantees they now do, so it
```
catonph: Tiny nit: this comment refers to `0175_create_state_white_labels`, but the migration is actually `0176_create_state_white_labels` (which matches the `dependencies` at the bottom). Probably a leftover from a rebase.

catonph (approved):
LGTM!

PR #1776, opened 2026-09-17
MFB-1230: Implement MO Refugee Cash Assistance (mo_rca)

elanatee:
## Context & Motivation

- Implements [MFB-1230](https://linear.app/myfriendben/issue/MFB-1230/program-mo-refugee-cash-assistance): the Missouri Refugee Cash Assistance (RCA) custom calculator, built against the `spec.md` and `mo_rca_initial_config.json`

## Changes Made

- Added `MoRca` (`programs/programs/white_labels/mo/rca/calculator.py`), `program_code = "mo_rca"`.
  - Resolves the household into RCA cases first: an adult child (18+) with a `child`-type relationship forms their own single-member case (criterion 6); everyone else shares one case.
  - Removes, from each case, any member with a reported `sSI` or `cashAssistance` (TANF) payment (criterion 2) — reported receipt only, never eligibility, and never a household-level TANF tile read.
  - Computes each case's net countable income (`$90` work exemption + three-fourths earned disregard, unearned counted in full except the Data Gap 5 `cashAssistanceOther`/`gifts` exclusion) and compares it against `rca_max_payment(case_size)`, strictly.
  - Value is `(standard - net_income) * 8` per payable case, summed across the household, matching `value_format: lump_sum`. `member_value()` returns 0 so the award isn't double-counted.
- Added `member_reports_tanf_amount()` beside `member_reports_ssi_amount()` in `programs/framework/pe_dependencies/receipt.py` — the per-member TANF-receipt read the spec calls for (no twin existed yet).
- Added `programs/programs/white_labels/mo/rca/tests/test_mo_rca.py` covering all 20 `spec.md` Test Scenarios plus the Acceptance Criteria not already pinned by a scenario (the payment-table function, the `DependencyError` path, and the Data Gap 5 exclusion).

## Testing

- Migrations to run: none.
- Configuration updates needed: none — `mo_rca_initial_config.json` already exists on this branch; import it with the existing `import_program_config_data` command when ready to load the `Program` row (currently `active: false`).
- Manual testing steps:
  - `python manage.py test programs.programs.white_labels.mo.rca`
  - Full suite: `python manage.py test programs.programs.white_labels.mo` (38 pre-existing failures in `mo_pts`/`mo_wftc` are unrelated live-network PolicyEngine lookups, reproducible on `main`)

## Deployment

- Post-deployment scripts needed: none.
- Production config updates: flip `active: true` in the program config when MO-ORA review is complete (currently `false`, matching KS RCA's launch pattern).
- Admin updates needed: none.
- Notify team/users of: n/a.

## Notes for Reviewers

- Known limitations: the data gaps in `spec.md` (no ORR eligibility date/8-month clock, no higher-ed student distinction, no TANF assistance-unit membership, no per-member immigration status for the § 400.208 mixed-status payment unit) are all handled inclusively, as documented there — consistent with how `ks_rca` handles its own gaps.
- Future considerations: none beyond what `spec.md`'s Related Programs section already flags (RMA, MO-RTAC) as explicitly out of scope.

## Summary by CodeRabbit

* **New Features**
  * Added Missouri Refugee Cash Assistance (RCA) eligibility and benefit calculations.
  * Added support for income-based eligibility, household case splitting, and benefit calculations over an eight-month period.
  * Added Missouri RCA program information, application links, required documentation, and assistance contact details.
  * Added per-member detection of reported TANF income.

* **Tests**
  * Added comprehensive coverage for eligibility rules, income handling, household scenarios, benefit amounts, and program registration.


programs/management/commands/import_program_config_data/data/mo_rca_initial_config.json
```diff
@@ -0,0 +1,53 @@
...
+    "year": "2026",
+    "legal_status_required": [
+      "refugee",
+      "otherWithWorkPermission",
+      "gc_5less"
+    ],
+    "name": "Refugee Cash Assistance (RCA)",
+    "description": "Refugee Cash Assistance (RCA) gives monthly cash to refugees and other newcomers. It can help pay for food, rent, travel, and other basic needs while you look for work.\n\nThe Missouri Office of Refugee Administration (MO-ORA) runs RCA. You apply through a local agency, and that agency pays you each month. This amount covers the whole 8-month period, not one month. The agency reviews your income each month. It leaves out part of what you earn from work, so a job does not cost you the full benefit.\n\nThe 8 months start from the day you arrived or got your status. The agency will tell you how many months you have left. It also checks your exact status. If you can get Temporary Assistance (TANF), you get that instead. The agency's money that you receive to get settled will not lower your payment.\n\nThe agency may not count everyone in your home, so your total may be lower. Select \"Contact Local Agency\" to find a Missouri agency you can ask about applying.",
```
catonph, line 17: Let's clarify what "This amount" is - I think it should be that the benefit value shown is the full 8 months? but they might not have 8 months left

programs/programs/white_labels/mo/rca/calculator.py
```diff
@@ -0,0 +1,182 @@
...
+    override. The award itself lives entirely in `household_value` — `member_value`
+    returns 0 — since a per-member award would multiply it by case size.
+
+    Criterion 1 (ORR-eligible immigration status) is carried by the program row's
+    `legal_status_required` and is not evaluated here. Criterion 3 (the SSI-pending
+    pathway) needs no implementation: it is satisfied by construction, since
+    criterion 2 fires only on a *reported* SSI amount. The ORR eligibility window,
+    the higher-education student exclusion, and TANF assistance-unit membership are
+    data gaps handled inclusively — see spec.md.
+    """
```
catonph, line 115: Let's not reference criterion by number which may drift.

programs/programs/white_labels/mo/rca/calculator.py
```diff
@@ -0,0 +1,182 @@
+from typing import ClassVar, Optional
+
+from programs.framework.base import MemberEligibility, ProgramCalculator
+from programs.framework.pe_dependencies.receipt import member_reports_ssi_amount
```
catonph: Can you do me a favor and clean this one up to? I think I introduced it 😵‍💫

programs/programs/white_labels/mo/rca/calculator.py
```diff
@@ -0,0 +1,182 @@
...
+EXCLUDED_UNEARNED_TYPES = ["cashAssistanceOther", "gifts"]
+
+# The per-member twin of `member_reports_ssi_amount`: whether this member reports a
+# TANF (`cashAssistance`) dollar amount of their own. Local to this calculator since
+# it is the only caller — see spec.md criterion 2.
+TANF_INCOME_TYPE = "cashAssistance"
+
+
+def _member_reports_tanf_amount(member: HouseholdMember) -> bool:
+    return member.calc_gross_income("yearly", [TANF_INCOME_TYPE]) > 0
```
catonph: idk if this really warrants being a function - could probably just check this and SSI in-line (and update MO to check SSI in-line too)

programs/programs/white_labels/mo/rca/calculator.py
```diff
@@ -0,0 +1,182 @@
...
+
+def _splits_into_own_case(member: HouseholdMember) -> bool:
+    """Criterion 6: an adult child (18+) forms their own single-member case, separate
+    from the rest of the household. `calc_age()` returning `None` fails open — an
+    unknown-age member stays in the primary case."""
+    age = member.calc_age()
+    return age is not None and age >= 18 and member.relationship in CHILD_RELATIONSHIPS
+
+
+class _Case:
```
catonph: _Case feels heavier than what we probably need here

Sub-household units are real, but I think this models them backwards. The framework hands member_eligible() one member at a time, so building case objects creates a reverse-lookup problem — which is what `contains()`, `_case_for_member()`, and the `hasattr` cache all exist to solve. `contains()` also has to match on `.id` rather than identity because the framework re-queries `household_members.all()`; correctness resting on a comment about an identity mismatch is the bit I'd most like gone.

`MoTwha` (cross_white_label/medicaid/disability/mo.py) has the same shape — assistance unit ≠ household — and goes member → unit instead, so no lookup and no cache:

```
def _assistance_unit(self, worker: HouseholdMember) -> list[HouseholdMember]:
    """The worker, plus their spouse if there is one — never the whole household."""
```

RCA's split rule is member-local too, so the same inversion works:

```
def _case_members(self, member: HouseholdMember) -> list[HouseholdMember]:
    if _splits_into_own_case(member):
        return [member]
    return [m for m in self.screen.household_members.all() if not _splits_into_own_case(m)]


def _case_award(self, members: list[HouseholdMember]) -> float:
    """Monthly award for a case, or 0 if net income doesn't clear the standard."""
    remaining = [m for m in members if not _removed_from_case(m)]
    if not remaining:
        return 0
    net_income = sum(_net_countable_income(m) for m in remaining)
    standard = rca_max_payment(len(remaining))
    return standard - net_income if net_income < standard else 0
```

`contains()`,` _case_for_member()`, the cache, and the `Optional[_Case]` all drop out.


Catch: `MoTwha` doesn't face this — its value is per-member (member_amount, no `household_value()` override), so nothing needs deduping. RCA's award is per case, so `household_value()` must count each case once. I tried it: the dedupe is four lines with no bookkeeping, and all 26 tests pass unchanged (182 → 165 lines). The one real cost is that _splits_into_own_case gets called in two places instead of the partition being stated once. I think that's worth it to delete the class, the cache, and the .id matching — but it's your call.

catonph (changes requested):
A few requested changes!

programs/management/commands/import_program_config_data/data/mo_rca_initial_config.json
elanatee: Do you mean it'd be more clear to change this sentence: `This amount covers the whole 8-month period, not one month.` to something more like: `This benefit amount represents what you're eligible to receive over an 8-month period.`?

programs/management/commands/import_program_config_data/data/mo_rca_initial_config.json
```diff
@@ -0,0 +1,53 @@
...
+      "refugee",
+      "otherWithWorkPermission",
+      "gc_5less"
+    ],
+    "name": "Refugee Cash Assistance (RCA)",
+    "description": "Refugee Cash Assistance (RCA) gives monthly cash to refugees and other newcomers. It can help pay for food, rent, travel, and other basic needs while you look for work.\n\nThe Missouri Office of Refugee Administration (MO-ORA) runs RCA. You apply through a local agency, and that agency pays you each month. The Estimated Annual Value represents what you may be eligible to receive over an 8-month period. The 8-month period starts from the day you arrived or got your status. The agency will tell you how many months you have left. It also checks your exact status. The agency reviews your income each month. It leaves out part of what you earn from work, so a job does not cost you the full benefit.\n\nIf you are eligible for Temporary Assistance (TANF), you will receive TANF instead. The agency's money that you receive to get settled will not lower your payment.\n\nThe agency may not count everyone in your home, so your total may be lower. Select \"Contact Local Agency\" to find a Missouri agency you can ask about applying.",
```
catonph: the value format is "lump_sum" which means this is a one time payment not an "Estimated Annual Value"

programs/programs/white_labels/mo/rca/tests/test_mo_rca.py
```diff
@@ -0,0 +1,293 @@
...
+    screen = Mock()
+    screen.household_members.all.return_value = members
+    # Acceptance Criterion 6: the calculator must make no CurrentBenefit/has_benefit
+    # read for TANF at all — only a member's own reported `cashAssistance` amount.
+    screen.has_base_benefit.side_effect = AssertionError("mo_rca must not read has_base_benefit for TANF")
+
+    missing_dependencies = Mock()
+    missing_dependencies.has.return_value = False
+
+    return MoRca(screen, Mock(), {}, missing_dependencies)
```
catonph: can we rebase your PR onto the test scaffolding one and then use the shared fixtures instead of creating new ones here

https://linear.app/myfriendben/review/mfb-1564-test-scaffolding-for-custom-calculators-and-unit-tests-for-a84a28e859b9

programs/programs/white_labels/mo/rca/calculator.py
```diff
@@ -0,0 +1,173 @@
...
+        return RCA_MAX_PAYMENT[case_size]
+
+    return RCA_MAX_PAYMENT[LARGEST_TABULATED_SIZE] + ADDITIONAL_MEMBER_AMOUNT * (case_size - LARGEST_TABULATED_SIZE)
+
+
+def _removed_from_case(member: HouseholdMember) -> bool:
+    """A member with a reported SSI or TANF (`cashAssistance`) payment is removed
+    from their own RCA case, taking their income with them. Reported receipt only
+    — never eligibility for either program, and never a household-level TANF
+    report (spec.md's SSI-pending pathway; see Scenario 19)."""
```
catonph: i would just delete this piece in parentheses

catonph (approved):
Approved because I dont need to take another look but check out my 2 comments that should be addressed first!

